PR #581, opened 2015-12-02
Adding blog posts

mjbrender:
And shifting Active Development down so we don't overload the beginning


mjbrender:
Confirmed it looks good and links work. :ship:ing at @lynxbat request :goat:

PR #595, opened 2015-12-04
fix #530: exit if given an empty keyring path

### LINKED ISSUE #530: Daemon will continue to start with trust set to 1 or 2 if an empty directory is passed for keyring path
With trust set to 1 or 2, if an empty directory is passed as the path to the keyrings, pulse daemon will successfully start up. With trust set to 1 or 2, at least one keyring should have to set, or the daemon should throw an error and exit.

danielscottt:


tiffanyfj:
LGTM

PR #599, opened 2015-12-07
Address golint warning for package plugin and schedule

candysmurf:


danielscottt:
This looks pretty good, but I'm not sure how helpful the `// const defines constants` and the `// scheduler vars`-style comments are.  I think the vet/lint errors you see here are with respect to commenting on the exported items, be they functions or variables.

scheduler/scheduler.go
```diff
@@ -36,6 +36,7 @@ import (
 	"github.com/intelsdi-x/snap/scheduler/wmap"
 )
 
+// Schedule var variables
```
jcooklin: You could describe how HandlerRegistrationName is used to register a gomit event handler.

scheduler/task.go
```diff
@@ -35,11 +35,13 @@ import (
 	"github.com/intelsdi-x/snap/scheduler/wmap"
 )
 
+// const defines constants
```
jcooklin: A description of what the DefaultDeadlineDuration and DefaultStopOnFailure do should be provided.

danielscottt:
LGTM

danielscottt:
@candysmurf Are you ready to merge this one?

PR #600, opened 2015-12-07
fix #479 - Plugins can remain running after a task fails to start

### LINKED ISSUE #479: Plugins can remain running after a task fails to start
If a task fails to start, any plugins started during the process of starting the task remain running and are not cleaned up.

andrzej-k:
This commit intends to fix issue described in #479


danielscottt:
This is right on, but I'm going to nitpick on the style.  Both `sub_errs` and `unsub_errs` should be `subErrs` and `unsubErrs` respectively.  I also like `serrs` and `uerrs`.

andrzej-k:
@danielscottt Thanks for review. Refactor done.

jcooklin:
:+1:

PR #602, opened 2015-12-07
Removed github token from docker run tests script

tiffanyfj:
Since this repo is public, a github token is no longer needed.


danielscottt:
:+1:

PR #607, opened 2015-12-08
Fixed Example usage in SNAPCTL.md

tiffanyfj:


danielscottt:
LGTM

As a side note, I wonder if we should change the UX for unload to be something like `snapctl unload type:name:version`  The existing one seems a bit kludgy.

tiffanyfj:
I think that would be much nicer. The current one is a bit annoying. Also I think it would be nice if we could unload all plugins with the same name or type or version if someone wanted.

PR #611, opened 2015-12-09
Fixes #610 for control/plugin/client/client.go nil comparison

### LINKED ISSUE #610: Error in control/plugin/client/client.go nil comparison
./scripts/test.sh                                ✹ ✭
gofmt
goimports
go vet
control/plugin/client/client.go:148: comparison of function Labels == nil is always false
exit status 1

Labels should be Labels()

tiffanyfj:
#610


jcooklin:
:+1:

PR #623, opened 2015-12-10
Fixes #525 swap plugins of diff types

### LINKED ISSUE #525: Swap plugins should not be able to swap plugins of different types
Currently, there are no checks in swap plugins that the plugin being loaded and unloaded are the same minus a different in version number. Any plugin can be swapped with any loaded plugin in the current implementation.

tiffanyfj:
#525 
Based on feedback from @danielscottt, mock 1 and mock 2 plugin names should just be mock since they're different versions of mock plugins. Made this change.

Also swap now verifies that the plugins are the same type and name.

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -381,7 +381,6 @@ func (p *pluginControl) Unload(pl core.Plugin) (core.CatalogedPlugin, serror.Sna
 }
 
 func (p *pluginControl) SwapPlugins(in *core.RequestedPlugin, out core.CatalogedPlugin) serror.SnapError {
-
 	details, serr := p.returnPluginDetails(in)
 	if serr != nil {
 		return serr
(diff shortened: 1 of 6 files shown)


control/control.go
```diff
@@ -395,16 +395,41 @@ func (p *pluginControl) SwapPlugins(in *core.RequestedPlugin, out core.Cataloged
 		return err
 	}
 
+	// Make sure plugin types and names are the same
+	if lp.TypeName() != out.TypeName() || lp.Name() != out.Name() {
+		fmt.Println(lp.TypeName())
```
danielscottt: Want to pull out these `Println`s?

control/control.go
```diff
@@ -395,16 +395,41 @@ func (p *pluginControl) SwapPlugins(in *core.RequestedPlugin, out core.Cataloged
...
+		fmt.Println(out.TypeName())
+		fmt.Println(lp.Name())
+		fmt.Println(out.Name())
+		serr := serror.New(errors.New("Plugin types and names must match."))
+		serr.SetFields(map[string]interface{}{
+			"in Type":  lp.TypeName(),
+			"out Type": out.TypeName(),
```
danielscottt: I think by convention these keys shouldn't contain spaces.

Technically it's valid, but I'd prefer something like `in_type`.

tiffanyfj: Did in-type to match the format of the ones already there.

danielscottt:
:ship: it.

Nicely done.

tiffanyfj:
Thanks.

PR #624, opened 2015-12-10
address golint warnings for cmd package

candysmurf:


cmd/snapctl/task.go
```diff
@@ -49,10 +49,15 @@ var (
 )
 
 const (
+	// K constant is the default count for truncating task hit and miss
```
danielscottt: Not sure these are all necessary.  Maybe something like:

```go
// used to truncate task hit and miss counts
// e.g. 1.4K, 3M, etc
```

cmd/snapctl/task.go
```diff
@@ -48,6 +48,9 @@ var (
 	unionParseFormat = timeParseFormat + " " + dateParseFormat
 )
 
+// Constants used to truncate task hit and miss counts
+// e.g. 1k, 1M, etc. As a side note, we probably don't need
+// anything after the M.
```
tiffanyfj: K should be capitalized :). I'm not sure if @danielscottt really meant to have the side note as part of the comment.

danielscottt: +1 on both statements above.

danielscottt:
LGTM

PR #642, opened 2015-12-20
Use the global TTL if it is greater than the (minimum) value provided by the plugin

### LINKED ISSUE #641: If the global TTL is greater than the minimum advertised by the plugin us it instead
If a plugin advertises a cache TTL value it is the minimum and should not override a global system default that is greater.

jcooklin:
Fixes #641


tiffanyfj:
LGTM

PR #643, opened 2015-12-21
Factor out promise abstraction

ConnorDoyle:
This patch set makes the notion of a `Promise` more abstract and re-usable, and demonstrates usage in `queuedJob`.  Promise is pretty bare-bones right now.  Later, we could add more features.  One that is noticeably missing is the ability to supply a timeout for deferred executions.

### DIFF
--- a/pkg/promise/promise.go
+++ b/pkg/promise/promise.go
@@ -1,0 +1,63 @@
+package promise
+
+import (
+	"sync"
+)
+
+// A disposable write-once latch, to act as a synchronization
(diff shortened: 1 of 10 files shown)


danielscottt:
I dig it.
 :+1:

lynxbat:
This is sweet. Maybe when this flushes out a good thing to break out of snap for reuse.

ConnorDoyle:
Thanks guys, hope it proves useful.  Landing this now.

Issue #493, opened 2015-11-12
TestWorkerManager in `work_manager_test` is flaky.


ConnorDoyle:
Observed spurious assertion failures on OS X.

```go
      ...
      go manager.Work(j1)                                                                                                                                                                                                                                                             
      go manager.Work(j2)                                                                                                                                                                                                                                                             
      manager.Work(j3)                                                                                                                                                                                                                                                                
      time.Sleep(time.Millisecond * 100)                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                      
      worked := 0                                                                                                                                                                                                                                                                     
      for _, j := range []*mockJob{j1, j2, j3} {                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                      
        if j.worked == true {                                                                                                                                                                                                                                                         
          worked++                                                                                                                                                                                                                                                                    
        }                                                                                                                                                                                                                                                                             
      }                                                                                                                                                                                                                                                                               
      So(worked, ShouldEqual, 2)  // <= is sometimes 1
      ...
```

ConnorDoyle:
Have seen this one crop up in CI a number of times lately.  Submitted #645 as a new attempt to fix this :beetle:

ConnorDoyle:
Fixed by #645.

PR #645, opened 2015-12-23
Guarantee queueing error in work manager test.

ConnorDoyle:
- Adds a new `RendezVous` abstraction in `pkg/promise` to make it easier for coordinating goroutines to synchronize.


ConnorDoyle:
Not to jinx anything, but I believe this finally squashes #493 :santa:

danielscottt:
LGTM

PR #646, opened 2015-12-28
Updates in PLUGIN_CATALOG

andrzej-k:
1. Moved Ethtool and IOstat plugins from "Committed plugins" to "Maintained plugins" list
2. Added several plugins (Disk, Swap, Meminfo, Load, Interface, Df and HAProxy) to "Committed plugins" list
3. Removed Nova plugin from "Wish List" as it's already on "Committed plugins" list


tiffanyfj:
LGTM

PR #647, opened 2015-12-29
Added timeout variants for blocking promise ops.

ConnorDoyle:
Supplying a timeout when awaiting completion of an asynchronous event can help avoid "leaking" goroutines that can never make progress.

### DIFF
--- a/pkg/promise/promise.go
+++ b/pkg/promise/promise.go
@@ -1,7 +1,9 @@
 package promise
 
 import (
+	"fmt"
 	"sync"
+	"time"
 )
 
 // A disposable write-once latch, to act as a synchronization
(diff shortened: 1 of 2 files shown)


ConnorDoyle:
Rebased.

pkg/promise/promise.go
```diff
@@ -49,16 +65,47 @@ func (p *promise) Complete(errors []error) {
...
+	if duration.Nanoseconds() > 0 {
+		timeoutChan = time.After(duration)
+	}
+
+	select {
+	case <-p.completeChan:
+		return p.errors
+	case <-timeoutChan:
+		msg := fmt.Sprintf("Await timed out for promise after [%s]", duration)
+		return []error{NewTimeoutError(msg)}
```
danielscottt: I think you may want to do:
```go
return []error{fmt.Errorf("Await timed out for promise after [%s]", duration)}
```
here.

danielscottt:
LGTM mod commend above.

ConnorDoyle:
Restarted build and filed #648 to track that spurious test failure.

pkg/promise/promise.go
ConnorDoyle: I can make that change.  Admittedly, the custom error type doesn't add much in this case.

ConnorDoyle: I will also drop the `TimeoutError` type, since this is the only use site.

ConnorDoyle:
Addressed review comments.

Issue #655, opened 2016-01-07
Publisher job not running


thomastaylor312:
Is there any reason why a publisher job would just stop running without any error? All the data is showing up when I run `snapctl task watch`, but nothing is publishing to the influx data source. I am testing out a plugin and it was working fine until I added new metric types. But instead of failing, it is like the publish job doesn't even exist:

**Normal (before I made the change)**

```
time="2016-01-07T12:40:36-08:00" level=debug msg="collector run completed" _module=scheduler-job block=run job-type=collector metric-count=10 
time="2016-01-07T12:40:36-08:00" level=debug msg="starting publisher job" _module=scheduler-job block=run content-type=snap.gob job-type=publisher plugin-config=map[port:{8086} user:{foo} database:{t
esting} host:{example.com} password:{bar}] plugin-name=influx plugin-version=-1
time="2016-01-07T12:40:36-08:00" level=debug msg="plugin selected" _module=control-routing block=select hitcount=0 index="publisher:influx:v8:id7" pool size=1 strategy=least-recently-used 
time="2016-01-07T12:40:51-08:00" level=debug msg="task spin loop" _module=scheduler-task 
```

**Broken**

```
time="2016-01-07T12:44:53-08:00" level=debug msg="collector run completed" _module=scheduler-job block=run job-type=collector metric-count=18 
time="2016-01-07T12:44:53-08:00" level=debug msg="task spin loop" _module=scheduler-task
```

geauxvirtual:
My initial thought is the publisher plugin is panic'ing on an error (which
is the default behavior for the InfluxDB plugin that has not been fixed
yet). Can you look at the logs for the plugin and see if anything is in
those?

On Thu, Jan 7, 2016 at 12:56 PM, Taylor Thomas <notifications@github.com>
wrote:

> Is there any reason why a publisher job would just stop running without
> any error? All the data is showing up when I run snapctl task watch, but
> nothing is publishing to the influx data source. I am testing out a plugin
> and it was working fine until I added new metric types. But instead of
> failing, it is like the publish job doesn't even exist:
>
> *Normal (before I made the change)*
>
> time="2016-01-07T12:40:36-08:00" level=debug msg="collector run completed" _module=scheduler-job block=run job-type=collector metric-count=10
[...]

PR #658, opened 2016-01-08
Fix #657: Pass deadline for a task from task manifest or from command line for a workflow manifest

### LINKED ISSUE #657: Task deadline is not propagated from task manifest or from snapctl cli for a workflow manifest
Referenced from #655 

Found the issue why adding deadline to the task manifest is not working. We have deadline support on the server side of the REST api, but the client and the command line tool currently do not support the deadline option from the task manifest as seen here https://github.com/intelsdi-x/snap/blob/master/cmd/snapctl/task.go#L86-L91 and here https://github.com/intelsdi-x/snap/blob/master/mgmt/rest/client/task.go#L50-L58.

geauxvirtual:
This issue was found while troubleshooting #655. Task deadline was implemented in the REST api and in the TaskCreationRequest, but was not implemented in the REST client code or snapctl.

This PR adds support for specifying the deadline in a task manifest or on the command line for a workflow manifest via the --deadline flag.

### DIFF
--- a/core/task.go
+++ b/core/task.go
@@ -98,6 +98,14 @@ func TaskDeadlineDuration(v time.Duration) TaskOption {
 	return func(t Task) TaskOption {
 		previous := t.DeadlineDuration()
 		t.SetDeadlineDuration(v)
+		log.WithFields(log.Fields{
+			"_module":                "core",
+			"_block":                 "TaskDeadlineDuration",
+			"task-id":                t.ID(),
+			"task-name":              t.GetName(),
+			"task deadline duration": t.DeadlineDuration(),
+		}).Debug("Setting deadlineDuration on task")
+
 		return TaskDeadlineDuration(previous)
 	}
 }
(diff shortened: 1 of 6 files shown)


core/task.go
```diff
@@ -98,6 +98,14 @@ func TaskDeadlineDuration(v time.Duration) TaskOption {
 	return func(t Task) TaskOption {
 		previous := t.DeadlineDuration()
 		t.SetDeadlineDuration(v)
+		log.WithFields(log.Fields{
+			"_module":                   "core",
+			"_block":                    "TaskDeadlineDuration",
+			"task-id":                   t.ID(),
+			"task-name":                 t.GetName(),
+			"consecutive failure limit": t.DeadlineDuration(),
```
thomastaylor312: Should the field name for this be called `consecutive failure limit`?

geauxvirtual: Good catch. Bad copy and paste job last night.

danielscottt:
LGTM

PR #664, opened 2016-01-09
Fix #662: Task hangs on a large number of metrics

### LINKED ISSUE #662: Requesting large number of metrics leads to task hang
I found out requesting large number of metrics (around 500 is enough) from any collector plugin leads to task hang. 
Issue was check with lower number of metrics (50, 100, 200, 300) and there was no error.

Reproduction:
1. Load file publisher
```bash
$ $SNAP_PATH/bin/snapctl plugin load  build/plugin/snap-publisher-file
```
2. Load plugin 
```bash
$ $SNAP_PATH/bin/snapctl plugin load <plugin_name>
```
3. Prepare task manifest with more then 500 metrics (example attached)
4. Create task
```bash
$SNAP_PATH/bin/snapctl task create -t examples/task/task.json
```
5. Try to watch task
```bash
$SNAP_PATH/bin/snapctl task watch <id>
```

Fails:
Task watcher starts but no output shows up.
File publisher produces only 1 iteration of results, then no new content published
It is not possible to stop started task with
```bash
$SNAP_PATH/bin/snapctl task stop <id>
```
Task stays on the list of tasks with state "Running"

Publisher plugin, collector plugin heartbeat leaves. It is possible to unload publisher and collector.

[files_and_logs.zip](https://github.com/intelsdi-x/snap/files/82810/files_and_logs.zip)

jcooklin:
This issue was caused by a log statement in the file publisher plugin which printed the metrics to stdout.  If the log statement exceeded 4096 bytes the buffio errors and does not advance.  

* The routine that captures standard out and error has been fixed to handle the case where a plugin author prints a statement that exceeds 4096 bytes
* The file publisher plugin now prints the number of metrics being published to stdout

### DIFF
--- a/control/plugin/execution.go
+++ b/control/plugin/execution.go
@@ -276,11 +276,13 @@ func waitForPluginTimeout(timeout time.Duration, p pluginExecutor, waitChannel c
 func waitForResponseFromPlugin(r io.Reader, waitChannel chan waitSignalValue, logpath string) {
 	lp := strings.TrimSuffix(logpath, filepath.Ext(logpath))
 	lf, _ := os.OpenFile(lp+".stdout", os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0666)
+	defer lf.Close()
 	logger := log.New(lf, "", log.Ldate|log.Ltime)
 	processedResponse := false
 	scanner := bufio.NewScanner(r)
 	resp := new(Response)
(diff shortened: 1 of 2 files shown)


marcin-krolik:
Works now with processes as well as with dummy plugin.

IzabellaRaulin:
LGTM

PR #665, opened 2016-01-12
Add errors to dropped overdue jobs.

### LINKED ISSUE #656: Snap should log when jobs are skipped due to deadline expiration.
Quoting @geauxvirtual in a comment from related issue #655:

> The problem revolves around what is mentioned earlier in that the exported task having a new deadline field. By default, the deadline for a job is 5 seconds. The publish job takes the parents deadline which is the time the job was created + 5 seconds. You can see here in this part of the code https://github.com/intelsdi-x/snap/blob/master/scheduler/worker.go#L49-L56, we do not generate an error log if the deadline passes, and we just kill the job. If you look at the gist you posted,

ConnorDoyle:
- Added AddErrors() to job interface.
- Fixes #656.

### DIFF
--- a/scheduler/job.go
+++ b/scheduler/job.go
@@ -23,6 +23,7 @@ import (
 	"bytes"
 	"encoding/gob"
 	"fmt"
+	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


ConnorDoyle:
Rebased.

danielscottt:
LGTM

ConnorDoyle:
Thanks Dan, shipping.

PR #667, opened 2016-01-12
Fix typo in name of snap published mock file

IzabellaRaulin:
In exemplary task (/examples/tasks/mock-file.yaml) publisher is publishing to /tmp/snap_published_mock_file.log


jcooklin:
:+1:

PR #672, opened 2016-01-15
Add new collector plugins to PLUGIN_CATALOG

andrzej-k:
Added Processes, Users, MySQL, CPU and DBI collector plugins to the "Committed plugins" section.


IzabellaRaulin:
LGTM

PR #677, opened 2016-01-26
Change log message from error to debug

geauxvirtual:
Change log message from error to debug and log the correct namespace being added to cache. Added a log message for when an item is added to the cache that did not have any labels attached to the metric.

Current:
```
ERRO[0012] putting [foo bar baz qux]:0 in the cache  _module=control-routing
```

Change:
```
DEBU[0015] putting [foo bar * qux]:0 in the cache  _module=control-routing
```


tiffanyfj:
:+1:

PR #685, opened 2016-01-29
Changes to README.md

pjholmes:
Changed README.md sections on running snap, loading plugins and starting tasks. 

- Removed /bin following <snapBinPath> as this would make it .../bin/bin

- Added instruction to open a second shell to load plugins. 

- Ensured the user is in the proper directory before referencing plugin and task files.

- Instructions to list tasks in case the task ID already scrolled by (for example, if the user tailed the log file after starting the task.


mjbrender:
This is a great correction @pjholmes - thank you for the PR. LGTM.

PR #688, opened 2016-02-01
SDI-818: Restart failed plugins

### LINKED ISSUE #497: A failed or panicked available plugin can leave an apPool with not enough resources.
When a plugin panics while it is in use by a task, it emits a `DeadAvailablePlugin` event.  This event should be handled by runner, and it should apply the same eligibility rules to the affected `apPool`. I.e. start a new available plugin, and insert it in the dead plugin's place.

candysmurf:
resolves #497.

### DIFF
--- a/control/runner.go
+++ b/control/runner.go
@@ -268,6 +268,14 @@ func (r *runner) HandleGomitEvent(e gomit.Event) {
 					return
 				}
 				pool.IncRestartCount()
+
+				runnerLog.WithFields(log.Fields{
+					"_block":        "handle-events",
+					"event":         v.Name,
+					"aplugin":       v.Version,
+					"restart_count": pool.RestartCount(),
+				}).Warning("plugin restarted")
+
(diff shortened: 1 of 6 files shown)


tiffanyfj:
You need to rebase.

control/runner.go
```diff
@@ -247,9 +252,39 @@ func (r *runner) HandleGomitEvent(e gomit.Event) {
...
 			return
 		}
+
 		if pool != nil {
 			pool.Kill(v.Id, "plugin dead")
 		}
+
+		if pool.Eligible() {
+			if pool.RestartCount() < MaxPluginRestartCount {
+				e := r.restartPlugin(v.Key)
```
jcooklin: Let's add a log statement (warn) that we restarted plugin (name:version) x times.

jcooklin:
LGTM, after adding warn statement when restarting and rebasing.

control/runner.go
candysmurf: Add the warn log. please check. thanks

PR #697, opened 2016-02-03
Removing the old way of releasing snap from the project

geauxvirtual:
Removing the old script for releasing snap that was originally used when the project was first created. This script has been superseded by a better tool for releasing snap.


jcooklin:
:+1:

Issue #703, opened 2016-02-05
Task watch fails if https is enabled for REST API


geauxvirtual:
When HTTPs is enabled for the REST API, watching a task fails with the error: 

```
Get https://localhost:8181/v1/tasks/cd075085-463a-4ace-bed5-e4ad2f4cfa8b/watch: x509: certificate signed by unknown authority
```

PR #704, opened 2016-02-08
Fixes #701 and #703

### LINKED ISSUE #701: Incorrect global flags for snapctl
The help for snapctl lists global flags for url and apiversion as having two options:
`--url or -u` and `--api-version or -a`. However, neither -u or -a actually work because the code in cmd/snapctl/main.go is checking for `--u` and `--a`. Single letter options should only need a single dash in front of them.

https://github.com/intelsdi-x/snap/blob/master/cmd/snapctl/main.go#L67-L78

It could be argued that passing `--a` or `--u` should throw an error for an invalid flag instead of just ignoring the flag.

geauxvirtual:
This PR fixes issues #701 and #703.

#701 is an issue around global option flags for snapctl. Typically, options are defined as --foo or -f. snapctl had options defined as --foo and --f. Commit `2e4d245` changes the short form options for url and api-version to -u and -a respectively.

#703 is an issue where watch task would not work when HTTPs was enabled for the rest api. In the client code for WatchTask, WatchTask was calling http.Get directly instead of using the configured http client in client. Commit `dd1ed36` changes http.Get to c.http.Get to use the http client configured for client which allows support for watching a task when HTTPs is enabled.

### DIFF
--- a/cmd/snapctl/main.go
+++ b/cmd/snapctl/main.go
@@ -64,7 +64,7 @@ func init() {
 			if err := f1.Parse(os.Args[idx : idx+2]); err == nil {
 				url = *prtURL
 			}
-		case "--u":
+		case "-u":
 			if err := f1.Parse(os.Args[idx : idx+2]); err == nil {
 				url = *prtU
 			}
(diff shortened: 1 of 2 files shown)


jcooklin:
:+1:

PR #707, opened 2016-02-09
Added ability to start snapd with global config

tiffanyfj:


geauxvirtual:
Global config should not be appended to what already exists in control. It should be a separate package that handles parsing a config file on start of snapd, and then config options for Scheduler, Rest API, and Control are passed their config options when configured and started by the daemon.

control/config.go
```diff
@@ -27,12 +27,34 @@ import (
...
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/codegangsta/cli"
 
 	"github.com/intelsdi-x/snap/core"
 	"github.com/intelsdi-x/snap/core/cdata"
 	"github.com/intelsdi-x/snap/core/ctypes"
 )
 
+type flagConfig struct {
```
jcooklin: Looks good.  If you move this struct and its three methods to to pkg (e.g. pkg/cli/flags.go) I think we can get it merged.

geauxvirtual:
I don't see how this addresses global config by calling control.NewConfig. Global config is for all of snap, and should be a config package outside of control.

tiffanyfj:
@geauxvirtual Thoughts now? Thanks.

geauxvirtual:
Why was this merged without a LGTM?

jcooklin:
Note:  It had a verbal LGTM during the planning meeting by myself and Dan (before @geauxvirtual joined). 

After debating the PR further #710 was opened to address the remaining suggestions.

PR #716, opened 2016-02-17
added Elasticsearch collector plugin into the committed plugin catalog

candysmurf:


jcooklin:
@candysmurf :ship it after updating based on @mjbrender's comment above and rebasing.

PR #719, opened 2016-02-18
Remove dependency on glibc on linux platform.

jellonek:
With this change, we obtain a possibility to get truly static binaries, independent from glibc, which can be used under any linux distribution supporting elf binaries.
Binaries build using this updated appc/spec library will be usable under systems with libc different than glibc such as alpine linux (based on musl), openwrt (which could be build using klibc, musl, or uclibc), but also without any system, as a plain alone binary (probably under docker or other containerizing execution environment).

Warning: `Godeps.json` was modified manually. Could you describe in docs how to properly use `godep` in snap in time of updating library which is a dependency? `godep save -r ./...` gives me nice mess in resulting diff...

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -28,18 +28,14 @@ BINDIR=$BUILDDIR/bin
 COLLECTORDIR=$BUILDDIR/$PLUGINDIR/collector
 PUBLISHERDIR=$BUILDDIR/$PLUGINDIR/publisher
 PROCESSORDIR=$BUILDDIR/$PLUGINDIR/processor
-BUILDCMD='go build -a -ldflags "-w"'
+BUILDCMD='go build -a -ldflags "-w" -tags netgo'
 
 echo
 echo "****  snap build ($GITVERSION)  ****"
 echo
 
-# Disable CGO for builds...unless Linux...because appc
+# Disable CGO for builds.
 export CGO_ENABLED=0
-platform=$(uname)
-if [[ $platform == "Linux" ]]; then
-	export CGO_ENABLED=1
-fi
 
 # Clean build bin dir
 rm -rf $BINDIR/*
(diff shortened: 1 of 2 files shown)


scripts/build.sh
```diff
@@ -36,10 +36,6 @@ echo
 
 # Disable CGO for builds...unless Linux...because appc
```
andrzej-k: This comment is no longer valid - remove or update

jellonek: Thx. Done.

danielscottt:
Ping @geauxvirtual 

What are your thoughts here?

geauxvirtual:
I need to do pull the PR down and do some testing with the changes.

jellonek:
At least `appc/spec` was updated, so there could be main differences - but i didn't find in theirs diffs some greater behavior changes.

geauxvirtual:
I'm good with this PR. The PR just needs to be rebased against the latest Master branch so it can be merged in.

I do not have a good answer around godep right now. We are investigating other tools for managing dependencies of snap. `godep update` theoretically should work to update packages godep added to the Godeps.json file during a `godep save`, but this doesn't work as one would think. After doing a `godep restore` and then only changing the `github.com/appc/spec` dependency to 0.7.4, the diff between the existing Godeps.json file and what was in the GOPATH was definitely more than just the changes to `github.com/appc/spec`. I'd say manual updates for this dependency is fine for now.

Testing the snapd binaries built with these changes, it did allow the use of the binary inside a scratch container; however, due to the nature of scratch being completely empty, mounting volumes to autoload plugins, or even loading plugins through snapctl fail in scratch container since the files cannot be created. We'll need to investigate another minimal image to use that will allow full use of snapd with plugins.

danielscottt:
Merging this in.

PR #729, opened 2016-02-24
Fixes #260 Add url checking and comment about invalid target

### LINKED ISSUE #260: snapctl should do url checking and throw a better error
snapctl should do url checking and throw a formatting error if a user does not input a url in the format snapctl is expecting.

screen shot 2015-07-07 at 12 09 45 pm

tiffanyfj:
#260


geauxvirtual:
Format your commit message `Fix #260: ...`

geauxvirtual:
A commit to fix an issue should only include changes for fixing the issue. Any other changes should be in a separate commit.

mgmt/rest/client/client.go
```diff
@@ -118,7 +124,7 @@ func (c *Client) do(method, path string, ct contentType, body ...[]byte) (*rbody
 			if strings.Contains(err.Error(), "tls: oversized record") || strings.Contains(err.Error(), "malformed HTTP response") {
 				return nil, fmt.Errorf("error connecting to API URI: %s. Do you have an http/https mismatch?", c.URL)
 			}
-			return nil, err
+			return nil, fmt.Errorf("URL target is not available. %v", err)
```
geauxvirtual: What is the reasoning for changing the error returned?

tiffanyfj: It was to address the "throw a better error" and previous comment about valid target. It didn't make sense to do another http request to check if the target was valid when it will error if it isn't anyway. What extra changes were there?

geauxvirtual: There are two changes in this PR. One that addresses issue #260. A second change that changes error messages returned after a request is made to the API.

tiffanyfj: I mixed up commit to equal PR. Got it. Will fix.

danielscottt:
LGTM

geauxvirtual:
I feel the URL checking should be done on the creation of the client vs on every call to `client.do`. The validation is not needed on the server url + path, but just on the server url passed into the --url flag on snapctl. Since snapctl is stateless, a client should not be created if URL is not in a valid scheme and errors before proceeding down any code path.

tiffanyfj:
Ok. The reason I had put it where I did was when I was looking to actually see if the target was valid and you can't do that in the creation since you get like a 404 from that url. I'll move it since I didn't go add additional target checking.

danielscottt:
I think this ready to land, what are your thoughts @geauxvirtual?

geauxvirtual:
Tests should be added for passing in invalid urls, unless I missed those being added in this PR.

tiffanyfj:
https://github.com/tiffanyfj/snap/blob/tb/edits/mgmt/rest/client/client_func_test.go#L512

mgmt/rest/client/client_func_test.go
```diff
@@ -160,332 +164,341 @@ func TestSnapClient(t *testing.T) {
...
+
+				p3 := c.GetPlugins(false)
+				So(p3.Err, ShouldBeNil)
+				So(len(p3.LoadedPlugins), ShouldEqual, 0)
+			})
 		})
 	})
-
-	c = New("http://localhost:127.0.0.1:-1", "v1", true)
+	c, err := New("http://localhost:127.0.0.1:-1", "v1", true)
```
geauxvirtual: Why is this line `http://localhost:127.0.0.1:`? It should be either `localhost` or `127.0.0.1`.

mgmt/rest/client/client_func_test.go
```diff
@@ -495,4 +508,10 @@ func TestSnapClient(t *testing.T) {
 		So(p2.Err, ShouldNotBeNil)
 		So(p3.Err, ShouldNotBeNil)
 	})
+
+	c, err = New("test", "", true)
```
geauxvirtual: `_, err`, don't need `c` if nothing is being tested.

mgmt/rest/client/client_func_test.go
tiffanyfj: No clue to be honest. Would have to ask @lynxbat.

mgmt/rest/client/client_func_test.go
geauxvirtual: Tests should be expanded to more than just testing passing a string. Will the client error if someone passes in `htp://` or `ftp://`, etc?

mgmt/rest/client/client_func_test.go
geauxvirtual: I think this brings up an issue with the change in that `http://localhost:127.0.0.1:-1` is an invalid url and should not return a client, but url.Parse is setting `localhost:127.0.0.1:-1` as the host and a client is being returned.

tiffanyfj: Yea, fixing. Broke something else with having to rearrange some convey statements...will submit again after I fix it.

tiffanyfj:
OK, updated. 
I did things like the following
```
	if cerr == nil {
		p2 = c.LoadPlugin(MOCK_PLUGIN_PATH2)
	}
```
to avoid putting it in a convey and ending up with multiple loads.

tiffanyfj:
I accidentally edited the url vs the convey message. Oops.

geauxvirtual:
LGTM

PR #743, opened 2016-03-03
Ensures scheduler jobs submissions are sent concurrently for a workflow.

lynxbat:
The existing schedulerWorkflow.workJob() function would submit all processor and publisher jobs for a workflow in a synchronous manner. This would mean each job would block the next going into the queue until it was completed in the queue.

This change ensures the jobs are submitted as a batch for a workflow.

### DIFF
--- a/scheduler/task.go
+++ b/scheduler/task.go
@@ -348,6 +348,16 @@ func (t *task) waitForSchedule() {
 	}
 }
 
+// RecordFailure updates the failed runs and last failure properties
+func (t *task) RecordFailure(e []error) {
+	// We synchronize this update to ensure it is atomic
+	t.Lock()
+	defer t.Unlock()
+	t.failedRuns++
+	t.lastFailureTime = t.lastFireTime
+	t.lastFailureMessage = e[len(e)-1].Error()
+}
+
 type taskCollection struct {
 	*sync.Mutex
 
--- a/scheduler/task.go
+++ b/scheduler/task.go
@@ -43,7 +43,7 @@ const (
 )
 
 var (
-	schedulerLogger = log.WithField("_module", "scheduler-task")
+	taskLogger = schedulerLogger.WithField("_module", "scheduler-task")
 
 	// ErrTaskNotFound - The error message for task not found
 	ErrTaskNotFound = errors.New("Task not found")
@@ -256,7 +256,7 @@ func (t *task) Schedule() schedule.Schedule {
 func (t *task) spin() {
 	var consecutiveFailures uint
 	for {
-		schedulerLogger.Debug("task spin loop")
+		taskLogger.Debug("task spin loop")
 		// Start go routine to wait on schedule
 		go t.waitForSchedule()
 		// wait here on
@@ -273,7 +273,7 @@ func (t *task) spin() {
 				t.fire()
 				if t.lastFailureTime == t.lastFireTime {
 					consecutiveFailures++
-					schedulerLogger.WithFields(log.Fields{
+					taskLogger.WithFields(log.Fields{
 						"_block":                    "spin",
 						"task-id":                   t.id,
 						"task-name":                 t.name,
@@ -285,7 +285,7 @@ func (t *task) spin() {
 					consecutiveFailures = 0
 				}
 				if consecutiveFailures >= t.stopOnFailure {
-					schedulerLogger.WithFields(log.Fields{
+					taskLogger.WithFields(log.Fields{
 						"_block":               "spin",
 						"task-id":              t.id,
 						"task-name":            t.name,
@@ -393,7 +393,7 @@ func (t *taskCollection) add(task *task) error {
 		//If we don't already have this task in the collection save it
 		t.table[task.id] = task
 	} else {
-		schedulerLogger.WithFields(log.Fields{
+		taskLogger.WithFields(log.Fields{
 			"_module": "scheduler-taskCollection",
 			"_block":  "add",
 			"task id": task.id,
@@ -411,7 +411,7 @@ func (t *taskCollection) remove(task *task) error {
 	defer t.Unlock()
 	if _, ok := t.table[task.id]; ok {
 		if task.state != core.TaskStopped {
-			schedulerLogger.WithFields(log.Fields{
+			taskLogger.WithFields(log.Fields{
 				"_block":  "remove",
 				"task id": task.id,
 			}).Error(ErrTaskNotStopped)
@@ -419,7 +419,7 @@ func (t *taskCollection) remove(task *task) error {
 		}
 		delete(t.table, task.id)
 	} else {
-		schedulerLogger.WithFields(log.Fields{
+		taskLogger.WithFields(log.Fields{
 			"_block":  "remove",
 			"task id": task.id,
 		}).Error(ErrTaskNotFound)
--- a/scheduler/workflow.go
+++ b/scheduler/workflow.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"sync"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/intelsdi-x/gomit"
 
 	"github.com/intelsdi-x/snap/control/plugin"
@@ -44,6 +45,8 @@ const (
 
 // WorkflowStateLookup map and error vars
 var (
+	workflowLogger = schedulerLogger.WithField("_module", "scheduler-workflow")
+
 	WorkflowStateLookup = map[WorkflowState]string{
 		WorkflowStopped: "Stopped",
 		WorkflowStarted: "Started",
@@ -299,6 +302,11 @@ func bindPluginContentTypes(pus []*publishNode, prs []*processNode, mm managesPl
 
 // Start starts a workflow
 func (s *schedulerWorkflow) Start(t *task) {
+	workflowLogger.WithFields(log.Fields{
+		"_block":    "workflow-start",
+		"task-id":   t.id,
+		"task-name": t.name,
+	}).Info(fmt.Sprintf("Starting workflow for task (%s\\%s)", t.id, t.name))
 	s.state = WorkflowStarted
 	j := newCollectorJob(s.metrics, t.deadlineDuration, t.metricsManager, t.workflow.configTree, t.id)
 
@@ -336,8 +344,20 @@ func (s *schedulerWorkflow) StateString() string {
 // workJobs takes a slice of proccess and publish nodes and submits jobs for each for a task.
 // It then iterates down any process nodes to submit their child node jobs for the task
 func workJobs(prs []*processNode, pus []*publishNode, t *task, pj job) {
+	// optimize for no jobs
+	if len(prs) == 0 && len(pus) == 0 {
+		return
+	}
 	// Create waitgroup to block until all jobs are submitted
 	wg := &sync.WaitGroup{}
+	workflowLogger.WithFields(log.Fields{
+		"_block":              "work-jobs",
+		"task-id":             t.id,
+		"task-name":           t.name,
+		"count-process-nodes": len(prs),
+		"count-publish-nodes": len(pus),
+		"parent-node-type":    pj.TypeString(),
+	}).Debug("Batch submission of process and publish nodes")
 	// range over the process jobs and call submitProcessJob
 	for _, pr := range prs {
 		// increment the wait group (before starting goroutine to prevent a race condition)
@@ -354,6 +374,14 @@ func workJobs(prs []*processNode, pus []*publishNode, t *task, pj job) {
 	}
 	// Wait until all job submisson goroutines are done
 	wg.Wait()
+	workflowLogger.WithFields(log.Fields{
+		"_block":              "work-jobs",
+		"task-id":             t.id,
+		"task-name":           t.name,
+		"count-process-nodes": len(prs),
+		"count-publish-nodes": len(pus),
+		"parent-node-type":    pj.TypeString(),
+	}).Debug("Batch submission complete")
 }
 
 func submitProcessJob(pj job, t *task, wg *sync.WaitGroup, pr *processNode) {
@@ -361,6 +389,14 @@ func submitProcessJob(pj job, t *task, wg *sync.WaitGroup, pr *processNode) {
 	defer wg.Done()
 	// Create a new process job
 	j := newProcessJob(pj, pr.Name(), pr.Version(), pr.InboundContentType, pr.config.Table(), t.metricsManager, t.id)
+	workflowLogger.WithFields(log.Fields{
+		"_block":           "submit-process-job",
+		"task-id":          t.id,
+		"task-name":        t.name,
+		"process-name":     pr.Name(),
+		"process-version":  pr.Version(),
+		"parent-node-type": pj.TypeString(),
+	}).Debug("Submitting process job")
 	// Submit the job against the task.managesWork
 	errors := t.manager.Work(j).Promise().Await()
 	// Check for errors and update the task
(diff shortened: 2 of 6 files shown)


scheduler/workflow.go
```diff
@@ -335,26 +336,38 @@ func (s *schedulerWorkflow) StateString() string {
...
-			t.failedRuns++
-			t.lastFailureTime = t.lastFireTime
-			t.lastFailureMessage = errors[len(errors)-1].Error()
-			return
-		}
+		go func() {
+			wg.Add(1)
+			defer wg.Done()
+			j := newProcessJob(pj, pr.Name(), pr.Version(), pr.InboundContentType, pr.config.Table(), t.metricsManager, t.id)
+			errors := t.manager.Work(j).Promise().Await()
```
ConnorDoyle: Just a suggestion: instead of creating a new goroutine here and then blocking with `Promise::Await()`, we could chain the `wg.Done()` and error handling after job completion using `Promise::AndThen()`.

scheduler/workflow.go
```diff
@@ -335,26 +336,38 @@ func (s *schedulerWorkflow) StateString() string {
 }
 
 func (s *schedulerWorkflow) workJobs(prs []*processNode, pus []*publishNode, t *task, pj job) {
+	// create a waitgroup track pending goroutines submitting jobs
```
ConnorDoyle: missing to, "to track"?

scheduler/workflow.go
ConnorDoyle: See usage of `AndThen` here: https://github.com/intelsdi-x/snap/blob/master/pkg/promise/promise_test.go#L56

lynxbat: My only thing about that is the WaitGroup and go outine use here is pretty idiomatic. The promise and queueing is not go channels so it is unique. I feel like this goroutine use is safer to understand and support longer term since it is a pretty normal use.

I did look at AndThen and I feel its use here would be harder to understand while note necessarily better.

scheduler/workflow.go
lynxbat: Oops

scheduler/workflow.go
ConnorDoyle: Fair enough

scheduler/workflow.go
```diff
@@ -335,26 +336,38 @@ func (s *schedulerWorkflow) StateString() string {
...
-			t.lastFailureMessage = errors[len(errors)-1].Error()
-			return
-		}
+		go func() {
+			wg.Add(1)
+			defer wg.Done()
+			j := newProcessJob(pj, pr.Name(), pr.Version(), pr.InboundContentType, pr.config.Table(), t.metricsManager, t.id)
+			errors := t.manager.Work(j).Promise().Await()
+			if len(errors) != 0 {
+				t.failedRuns++
```
ConnorDoyle: Do we need to make these updates atomic for thread-safety? (all of the dispatched goroutines close over `t`)

scheduler/workflow.go
```diff
@@ -335,26 +336,38 @@ func (s *schedulerWorkflow) StateString() string {
...
 	for _, pr := range prs {
-		j := newProcessJob(pj, pr.Name(), pr.Version(), pr.InboundContentType, pr.config.Table(), t.metricsManager, t.id)
-		errors := t.manager.Work(j).Promise().Await()
-		if len(errors) != 0 {
-			t.failedRuns++
-			t.lastFailureTime = t.lastFireTime
-			t.lastFailureMessage = errors[len(errors)-1].Error()
-			return
-		}
+		go func() {
```
danielscottt: The reason this is bombing in travis, and I think rightfully so, is because you need to pass `pr` into the go routine.  The scope here is a lie -- I think every time you run the loop you're getting the item at index `len(prs)-1`.

scheduler/workflow.go
```diff
@@ -335,26 +336,38 @@ func (s *schedulerWorkflow) StateString() string {
...
 	for _, pu := range pus {
-		j := newPublishJob(pj, pu.Name(), pu.Version(), pu.InboundContentType, pu.config.Table(), t.metricsManager, t.id)
-		errors := t.manager.Work(j).Promise().Await()
-		if len(errors) != 0 {
-			t.failedRuns++
-			t.lastFailureTime = t.lastFireTime
-			t.lastFailureMessage = errors[len(errors)-1].Error()
-			return
-		}
+		go func() {
```
danielscottt: Same here.

scheduler/workflow.go
lynxbat: The mutex on `t (scheduler.Task)` is used for state change safety:
https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L207
https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L218
https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L228
https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L240
https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L335
https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L296
https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L309
https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L325

In fact, the above makes me think the internals of task should have a private function for changing state and unexport the sync.Mutex.

Since we are are just sending jobs into the task queue effectively, this does not conflict with state change since it never changes the state directly. Effectively this use of `t` is only for calling Work(j) and any possible `t.state` change is not a thread safety problem.

However, one could reason that job queues should be cleared if `t.state` were to switch to an ended or disabled state. This is right now not being done safely. I think we may want a job submission mutex to synchronize the clearing of the queue of jobs for a task with new submissions that may (or may not) be allowed.

ConnorDoyle: OK, I was not even thinking that far into it.  I'm concerned about these multiple goroutines all having a reference to the same struct that they will possibly mutate.  You don't know when you will get rescheduled following the Await() call, so conceivably another job could be worked and also fail... now you have two goroutines concurrently modifying `t`.  Or am I still missing something that prevents that case?

scheduler/workflow.go
lynxbat: I am going to refactor this.

scheduler/workflow.go
ConnorDoyle: I think a minimal fix, besides the refactoring you described above, would be to acquire `t`'s Mutex here before updating those members.

lynxbat:
Note: in the latest push I have converted `scheduler.workJobs()` to a package unexported function `workJobs()`. In the original function the only purpose the `*scheduler` pointer was providing was to reference the `workJobs()` function for child nodes of process nodes. By switching to a package function I was able simplify and remove the passing of `*scheduler` everywhere.

lynxbat:
@danielscottt I switched from goroutines with anonymous functions to functions for each type of submission. I feel this is way more readable and cleaner, especially given what needs to be passed into the function.

lynxbat:
This is NOT ready for pull. I am going to add logging into this since it has never been added. Debug logging for the timing, identify, and order of the submissions is something I have in a test harness. Moving the debug logging into the workflow.go natively.

scheduler/workflow.go
```diff
@@ -334,27 +335,65 @@ func (s *schedulerWorkflow) StateString() string {
...
+	// Decrement the waitgroup
+	defer wg.Done()
+	// Create a new process job
+	j := newPublishJob(pj, pu.Name(), pu.Version(), pu.InboundContentType, pu.config.Table(), t.metricsManager, t.id)
+	// Submit the job against the task.managesWork
+	errors := t.manager.Work(j).Promise().Await()
+	// Check for errors and update the task
+	if len(errors) != 0 {
+		// Note: we just update the error tracking, the change of state for a task is handled before this is reached.
+		t.failedRuns++
```
ConnorDoyle: Last comments about this got hidden by refactor...
> ...you don't know when you will get rescheduled following the Await() call, so conceivably another job could be worked and also fail... now you have two goroutines concurrently modifying `t`.  I think we should acquire `t`'s mutex here to prevent that, unless that can't happen for some reason I didn't see.

scheduler/workflow.go
lynxbat: Here is more detail to think about:

Only the task itself reaches workJob and by design the function at the top only allows one root goroutine to ever be running workJob:
1. spin() blocks while executing - https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L256
2. spin() calls fire() - https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L273
4. fire() call starts the workflow - https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L256
5. the workflow calls workJobs - https://github.com/intelsdi-x/snap/blob/master/scheduler/workflow.go#L326

The risk of mutation on task is on the properties that this flow accesses which are:
1. deadlineDuration (mutable and thread safe)
2. metricsManager (immutable)
3. *schedulerWorkflow (immutable - workflows cannot be altered)
4. task id (immutable)
5. failedRuns, lastFailureTime, lastFailureMessage (mostly thread safe, probably need a mutex on these to ensure updates to all three are atomic). Two ops writing at the same time does not cause problems. But error messages would be mixed. Acquiring `t's` lock would be an easy way to synchronize this.

The only issue here is that by acquiring t's lock we are preventing tasks from being stopped until the entire workflow is worked (not submitted) since the Work(j).Promise().Await() is blocking. This is what bothers me since calls to change state on task are not async. A long running workflow would block something that should work. This is why I would prefer not to mix the state change mutex `t.Lock()` with a possible mutex around internal properties of `t` that we need to synchronize on.

To add flavor to the above, workJobs does not submit all of the work at once. Each iteration on a process node has to complete before calling the child node. So we would not be locking `t` and submitting all the jobs. We would be locking `t` for the duration of all jobs where the parent node has a child.

We would basically move the blocking operation we had on submission up to state changes on task.

lynxbat: To add to the above: it is fine for the state to change on a task while the workflow is running. In fact, we want this in some cases. The internals use of `t` here should kill the running workflow and flush the queue.

scheduler/workflow.go
lynxbat: I updated the original comment.

lynxbat:
@ConnorDoyle moving this to main thread from the hidden comments above:

Here is more detail to think about:

Only the task itself reaches workJob and by design the function at the top only allows one root goroutine to ever be running workJob:
1. spin() blocks while executing - https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L256
2. spin() calls fire() - https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L273
4. fire() call starts the workflow - https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L256
5. the workflow calls workJobs - https://github.com/intelsdi-x/snap/blob/master/scheduler/workflow.go#L326

The risk of mutation on task is on the properties that this flow accesses which are:
1. deadlineDuration (mutable and thread safe)
2. metricsManager (immutable)
3. *schedulerWorkflow (immutable - workflows cannot be altered)
4. task id (immutable)
5. failedRuns, lastFailureTime, lastFailureMessage (mostly thread safe, probably need a mutex on these to ensure updates to all three are atomic). Two ops writing at the same time does not cause problems. But error messages would be mixed. Acquiring t's lock would be an easy way to synchronize this.

The only issue here is that by acquiring t's lock we are preventing tasks from being stopped until the entire workflow is worked (not submitted) since the Work(j).Promise().Await() is blocking. This is what bothers me since calls to change state on task are not async. A long running workflow would block something that should work. This is why I would prefer not to mix the state change mutex t.Lock() with a possible mutex around internal properties of t that we need to synchronize on.

To add flavor to the above, workJobs does not submit all of the work at once. Each iteration on a process node has to complete before calling the child node. So we would not be locking t and submitting all the jobs. We would be locking t for the duration of all jobs where the parent node has a child.

We would basically move the blocking operation we had on submission up to state changes on task.

To add to the above: it is fine for the state to change on a task while the workflow is running. In fact, we want this in some cases. The internals use of t here should kill the running workflow and flush the queue.

lynxbat:
Thins that have to sync in task:
1. State (the property name, not the generic term) (externally or internally requested, e.g. Stop the task, Start the task)
2. jobs being submitted (task submits to the workManager)
3. properties of task (error, counters, etc)

Things that don't have to sync in task:
1. jobs being worked (workManager owns the queue and the work, not task. The workManager queue is thread safe)

ConnorDoyle:
OK, it all makes sense given the granularity at which that lock is already being used.  I'm just going to mention that the postfix increment to count failed runs is not atomic (no CAS), so concurrent execution means we could miscount.  Sorry for being pedantic, there are some nontrivial interactions here, as you described above.  If you did want to provide another more fine-grained lock, we could hide it behind a function like `Task::RecordFailure(...)`, which would only require the error as input.

lynxbat:
@ConnorDoyle yeah, I didn't look at the implementation there. You're right - it is unsafe.

I have a path I think solves for your needs and applies minimum lock stall. Let me write it up today. I won't update until you are sound asleep :)

ConnorDoyle:
The `atomic` package builds up useful stuff on top of CAS but you'd definitely want to abstract that usage instead of burdening the caller with handling that correctly.

lynxbat:
This is ready for review @intelsdi-x/snap-maintainers

jcooklin:
LGTM :+1:  

The comments on this PR are great.

scheduler/workflow_test.go
```diff
@@ -121,3 +125,172 @@ func TestCollectPublishWorkflow(t *testing.T) {
...
+func (m *Mock1) Work(j job) queuedJob {
+	m.Lock()
+	defer m.Unlock()
+	time.Sleep(m.delay)
+	m.queue[j.TypeString()]++
+	return m
+}
+
+func (m *Mock1) Promise() promise.Promise {
+	return m
```
ConnorDoyle: Would it make sense to embed a `Promise` in the `Mock1` struct instead?

scheduler/workflow_test.go
```diff
@@ -121,3 +125,172 @@ func TestCollectPublishWorkflow(t *testing.T) {
...
+}
+
+func (m *Mock1) CollectMetrics([]core.Metric, time.Time, string) ([]core.Metric, []error) {
+	return nil, nil
+}
+
+func (m *Mock1) Work(j job) queuedJob {
+	m.Lock()
+	defer m.Unlock()
+	time.Sleep(m.delay)
```
ConnorDoyle: What is the purpose of this sleep?  (Can we document that somewhere?)

scheduler/task.go
```diff
@@ -348,6 +348,16 @@ func (t *task) waitForSchedule() {
 	}
 }
 
+// RecordFailure updates the failed runs and last failure properties
+func (t *task) RecordFailure(e []error) {
```
ConnorDoyle: :+1:

scheduler/workflow_test.go
lynxbat: I didn't want to test the `Promise` behavior in this test so I mocked completely. I figured the exported `Promise` functions are pretty fixed.

scheduler/workflow_test.go
lynxbat: I was using this test as a way of being sure the behavior I wanted was happening when queue time (before work) was greater than 0. This is left over when I converted the test harness into a test. I can remove this since I am not sending a delay into the Mock1 struct to remove confusion.

lynxbat:
I am merging this now with the LGTM.

PR #745, opened 2016-03-03
Adding more psutil for :cloud: :computer:-ing

mjbrender:
Making a video that includes more metrics, particularly for vm data, and want to have them in our examples.

### DIFF
--- a/examples/tasks/psutil-file.yaml
+++ b/examples/tasks/psutil-file.yaml
@@ -1,16 +1,24 @@
---- # The Header starts after this line
-  version: 1        # All tasks, like plugins, have versioning
-  schedule:         # How frequently should we run this task?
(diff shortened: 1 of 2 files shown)


tjmcs:
it's actually a bit more than those three lines that are missing; this file also doesn't include the pass thru processor plugin...

mjbrender:
@tjmcs noted - fixing that now.

mjbrender:
@tjmcs I made the exact file name match the exact output, which will also be in the post. I want to keep the other example to show the difference between using a processor and not in YAML. Whatcha think?

tjmcs:
:+1: I like the idea of having multiple examples (with and without processor) and the code shown in the video linked to from your blog posting now matches one of the files available here :)

mjbrender:
thank you sir - merging.

PR #748, opened 2016-03-07
[ready-for-review] Adds grpc to the scheduler and control

jcooklin:
Supports distributed workflows 

Summary of changes:
- Adds grpc support to the scheduler

Testing done:
- unit
- manual 
secure-grpc

@intelsdi-x/snap-maintainers


scheduler/rpc/scheduler.proto
```diff
@@ -0,0 +1,147 @@
...
+}
+
+message Task {
+    string id = 1;    
+    uint64 task_state = 2;
+    string name = 3;
+    uint64 hit = 4;
+    uint64 misses = 5;
+    uint64 failed = 6; 
+    string last_failure = 7;
```
ConnorDoyle: `optional`?

ConnorDoyle: Please disregard -- apparently optional is the default in _proto3_ syntax.

scheduler/scheduler.go
```diff
@@ -105,25 +123,115 @@ type managesWork interface {
...
+// PublishWkrSizeOption sets the publisher worker pool size
+// and returns the previous previous publisher worker pool state.
+func PublishWkrSizeOption(v uint) SchedulerOption {
+	return func(s *scheduler) {
+		s.workerManagerOpts = append(s.workerManagerOpts, publishWkrSizeOption(v))
+	}
+}
+
+// rootCerts required
+func NewClient(addr, port, certPath, keyPath string) (rpc.TaskManagerClient, error) {
```
IRCody: What is going on here?

IRCody:
Summary of changes:
* add gRPC support to control

testing done:
* tests in control/control_proxy_test.go
* manual testing by running through common tasks (starting/stopping plugins/tasks/etc).

snapd.go
```diff
@@ -422,7 +457,6 @@ func action(ctx *cli.Context) {
 	if autodiscoverPath != "" {
 		log.Info("auto discover path is enabled")
 		paths := filepath.SplitList(autodiscoverPath)
-		c.SetAutodiscoverPaths(paths)
```
geauxvirtual: Is there a reason AutoDiscoverPaths is removed out of Control struct? Granted it does nothing right now, but auto discovering plugins (and eventually monitoring for plugin in said directories) needs to move out of snapd and into Control. Unless this is directly related to RPC changes, I would not remove this and let any changes regarding this be handled by a PR relating to plugin discovery changes.

control/control_proxy.go
```diff
@@ -0,0 +1,237 @@
...
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"runtime"
+	"time"
+
+	"github.com/intelsdi-x/snap/control/rpc"
+	"github.com/intelsdi-x/snap/core"
+	"github.com/prometheus/common/log"
```
geauxvirtual: We should stick to our common logging package import of logrus which is used throughout snap.

snapd.go
IRCody: I removed it because I noticed GetAutoDiscoverPaths() was part of the managesMetrics interface mgmt/rest/server.go. I didn't think it made sense to keep it in the interface when implementing RPC since it wasn't used. I decided to remove it from pluginControl because it didn't contain any special logic for handling autodiscover paths. It was only a getter/setter like below:
```golang
func (p *pluginControl) SetAutodiscoverPaths(paths []string) {
	p.autodiscoverPaths = paths
}

func (p *pluginControl) GetAutodiscoverPaths() []string {
	return p.autodiscoverPaths
}
```

geauxvirtual: Correct, but that's going to change. Removing it will just see it get added back in a later date. This should be left in for now.

geauxvirtual:
This PR currently does not build on my machine. I take it all the files required to run this code are in the PR and nothing needs to be generated, correct?

control/control_proxy.go
```diff
@@ -0,0 +1,237 @@
...
+	"golang.org/x/net/context"
+)
+
+type controlProxy struct {
+	control *pluginControl
+}
+
+func (pc *controlProxy) Load(ctx context.Context, arg *rpc.PluginRequest) (*rpc.PluginReply, error) {
```
geauxvirtual: This load method is incompatible with auto discovery. Auto discovery passes a path as we do not write the plugin to a temporary directory. This implementation is fine for plugins loaded from a REST API, but just gives more weight behind auto discovery needing to be moved inside of control.

geauxvirtual:
Fixing a very odd download of the github.com/golang/protobuf dependency resolved the issue of snapd starting.

geauxvirtual:
When starting snapd, this is seen which I conclude is from both Scheduler and Control trying to register the same rpc.Time,

```
--- intelsdi-x/snap » go run snapd.go -l 1 -t 0
2016/03/23 10:19:00 proto: duplicate proto type registered: rpc.Time
```

geauxvirtual:
I feel this PR and subsequent work on distributed workflow needs more discussion, or at least point me to any design documentation to implement distributed workflow for snap.

After looking through this PR that contains changes for scheduler and control, this PR could have been a lot smaller with a lot less changes to the actual implementation of how snapd currently runs. In this PR, only the REST API uses GRPC to communicate to the scheduler and control modules on the same node. The GRPC servers for scheduler and control are not started when scheduler and control are started, but only when a client is created to pass to the REST API server using the same listen addresses that the control and scheduler uses to start the server. While this demonstrates GRPC works on the local node, this does not support talking across nodes. On the actual snapd instance itself, scheduler still gets an instance of control when it starts, not a GRPC client to use to communicate with control.

Outstanding items for distributed workflow I see so far are:
- scheduler needs to communicate with control via GRPC
- scheduler at the moment needs a control instance to start, which will affect running scheduler without control also loaded on the same machine.
- scheduler needs to be able to talk to multiple controls running on different snapd instances
- REST API needs a scheduler and control module to start, which may not be running on the same node, and possibly needs to support talking to multiple scheduler and control instances

IRCody:
>After looking through this PR that contains changes for scheduler and control, this PR could have been a lot smaller with a lot less changes to the actual implementation of how snapd currently runs.

What changes have been made that you see as unnecessary, ignoring for now the autodiscoverPaths mentioned above.

>In this PR, only the REST API uses GRPC to communicate to the scheduler and control modules on the same node. 

I believe that is a reasonable scope for this PR given how large it is already. 

>The GRPC servers for scheduler and control are not started when scheduler and control are started, but only when a client is created to pass to the REST API server using the same listen addresses that the control and scheduler uses to start the server. While this demonstrates GRPC works on the local node, this does not support talking across nodes.

The gRPC servers are started when there respective Start() functions are called. This happens [here](https://github.com/intelsdi-x/snap/blob/distributed-workflow/snapd.go#L324). Talking across nodes is future work that needs to happen to enable distributed workflow.

> On the actual snapd instance itself, scheduler still gets an instance of control when it starts, not a GRPC client to use to communicate with control.

Passing in a client to the scheduler would involve implementing the scheduler.managesMetrics interface [here](https://github.com/intelsdi-x/snap/blob/distributed-workflow/scheduler/scheduler.go#L64-L72). I agree this needs to be done before distributed workflow can be fully implemented. The changes that will need to be done to the scheduler for this have overlap with what would need to be done to handle distributed workflow from workflow maps. I think more discussion needs to happen about how distributed workflow is going to be enabled in the manifest before this is done.

>Outstanding items for distributed workflow I see so far are:
> * scheduler needs to communicate with control via GRPC
> * scheduler at the moment needs a control instance to start, which will affect running scheduler without control also loaded on the same machine.
> * scheduler needs to be able to talk to multiple controls running on different snapd instances

These are addressed in the comment above.

> * REST API needs a scheduler and control module to start, which may not be running on the same node, and possibly needs to support talking to multiple scheduler and control instances

The REST API currently needs scheduler and control gRPC clients and an instance of control.config. Needing the control.config is a little odd. I think the way to fix this is to switch [this](https://github.com/intelsdi-x/snap/blob/distributed-workflow/mgmt/rest/server.go#L91-L98) interface to use gRPC and implement it in the control gRPC client. Then you could pass in only the gRPC clients.

snapd.go
IRCody: If it was a recent addition that had a clear plan to be used in the near future I'd be inclined to agree but this has been in and unused for 9 months. I don't think the fact that it would need to be added back in at a later date is a compelling argument for leaving it in. The removed code contains no logic that would make implementing those future changes any easier.

control/control_proxy.go
IRCody: I'm not sure I see the connection with auto discovery. What's the use case where auto discovery is happening over rpc?

geauxvirtual:
> What changes have been made that you see as unnecessary, ignoring for now the autodiscoverPaths mentioned above.

Not actually changing REST API to use GRPC. This PR could have been just adding GRPC and starting the servers, with others PRs after this adjusting the implementation of how components talk to each other.

> The gRPC servers are started when there respective Start() functions are called. This happens here. 

Oversight on my part when reading through the PR. I missed the light grey line showing what function it was added to, automatically rolling that change into the NewClient() function added which was shown right above it.

> Passing in a client to the scheduler would involve implementing the scheduler.managesMetrics interface here. I agree this needs to be done before distributed workflow can be fully implemented. The changes that will need to be done to the scheduler for this have overlap with what would need to be done to handle distributed workflow from workflow maps. I think more discussion needs to happen about how distributed workflow is going to be enabled in the manifest before this is done.

:+1:

>The REST API currently needs scheduler and control gRPC clients and an instance of control.config. Needing the control.config is a little odd. I think the way to fix this is to switch this interface to use gRPC and implement it in the control gRPC client. Then you could pass in only the gRPC clients.

:+1:

IRCody:
>>What changes have been made that you see as unnecessary, ignoring for now the autodiscoverPaths mentioned above.

>Not actually changing REST API to use GRPC. This PR could have been just adding GRPC and starting the servers, with others PRs after this adjusting the implementation of how components talk to each other.

That could have been done but I feel like changing the REST API at the same time felt very natural since either way you'd probably be referencing how the API did things when implementing the rpc server to ensure you are going to have the right behavior.

geauxvirtual:
Going to close this PR, and open a new PR from the rebase work done to have PR be able to eventually merge into Master.

PR #750, opened 2016-03-08
Documentation for labeling strategy - SDI-808

mjbrender:
* Consolidated repetitive types, especially enhancement and feature since we continue to fall into the trap of arguing which is which when we treat both similarly during sprints. TL;DR - we're good with just feature requests
* Aligned categories by some thematic use of color and embedded an image of it
* Adding this information to CONTRIBUTING.md
* Corrected some capitalization errors

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,11 +1,11 @@
-#Contributing to snap
+#Contributing to Snap
 
-snap is Apache 2.0 licensed and accepts contributions via GitHub pull requests. This document
+Snap is Apache 2.0 licensed and accepts contributions via GitHub pull requests. This document
 will cover how to contribute code, report issues, build the project and run the tests.
 
-## Contributing code
+## Contributing Code
 
-Contributing code to snap is a snap (pun intended).
+Contribution of code to snap is a snap (pun intended):
 - Fork the project to your own repository
 - Create a topic branch from where you want to base your work (usually master)
 - Make commit(s) (following commit guidelines below)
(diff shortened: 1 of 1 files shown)


mjbrender:
@intelsdi-x/snap-maintainers  - welcome feedback.

danielscottt:
How would you want to tag an RFC that would differentiate it from a spec?

mjbrender:
@danielscottt the thought is that any open issue is a request for feedback in same way. As a user, I'd write up a spec, tag it as specification. If I'd like to queue people to give feedback, I'd use "Help Wanted" as well. 

I have the idea of having Help Wanted be a flag to remind users (and me) to pay closer attention. It's the opposite of On Hold in many ways. 

Do you think RFC differentiates from Specification in a bigger way? I could see noting RFC in the spec description here.

danielscottt:
> Do you think RFC differentiates from Specification in a bigger way? I could see noting RFC in the spec description here.

Actually, no; you're right.  I think I'd rather change spec to RFC.  To me, a spec is _accepted_.  

Maybe there shouldn't be an issue created for either of these things; but rather, a folder called `spec` in the source, and PRs that get opened and tagged RFCs which get iterated on, then merged into `snap/specs`.

A workflow similar to: https://github.com/rust-lang/rfcs + https://github.com/rust-lang/rust, but all contained in this repository.

mjbrender:
@danielscottt awesome point. I think you're about to unveil a ton of missing documentation. 

We'd be in good company when reviewing [InfluxDB](https://github.com/influxdata/influxdb/labels/RFC). I don't see nearly enough others doing something with RFC or Spec. Hm. I'm still a fan and happy to add another example :smile:

CONTRIBUTING.md
```diff
@@ -13,15 +13,15 @@ Contributing code to snap is a snap (pun intended).
...
 
 ### Commit Guidelines
 
 Commits should have logical groupings. A bug fix should be a single commit. A new feature
-should be a single commit. 
+should be a single commit.
 
 Commit messages should be clear on what is being fixed or added to the code base. If a
-commit is addressing an open issue, please start the commit message with "Fix #XXX" or 
+commit is addressing an open issue, please start the commit message with "Fix #XXX" or
```
geauxvirtual: I'd like see `Fix #XXX:` and `Feature #XXX:`

CONTRIBUTING.md
```diff
@@ -13,15 +13,15 @@ Contributing code to snap is a snap (pun intended).
...
 - Open a pull request against the original repo and follow the pull request guidelines below
 
-The maintainers of the repo utilize a "Looks Good To Me" (LGTM) message in the pull request.
+The maintainers of the repo utilize a "Looks Good To Me" (LGTM) message in the pull request. After one or more maintainer states LGTM, we will merge. If you have questions or comments on your code, feel free to correct these in your branch through new commits.
 
 ### Commit Guidelines
 
 Commits should have logical groupings. A bug fix should be a single commit. A new feature
-should be a single commit. 
+should be a single commit.
```
ConnorDoyle: Are we asking contributors to "squash" commits following rounds of review before merge to satisfy this goal?

ConnorDoyle: Related to the request to include the "Fix #issue" in the commit message - without squashing, the "fix" ends up spread across several commits but only the first one gets linked back to the issue.

ConnorDoyle: An alternative is to ask people to tag issues in the PR instead (granted, issue numbers wouldn't get into the branch history, but those issue numbers are meaningless outside the context of GitHub anyway).

geauxvirtual: Yes, if a PR is opened with 5 commits that was the work involved to fix a single issue, it should be rebased to a single commit. If a PR is opened with 5 commits, with the first three to fix one issue and the second two to fix a separate issue, then it should be rebased to two commits, one for each issue.

geauxvirtual: We pull commit messages as part of the changelog currently when we issue a release, so having Fix #xxx makes it easy to see in the changelog what issue the commit addressed, and link is auto inserted by github to the issue.

CONTRIBUTING.md
mjbrender: Adding in the `:`

CONTRIBUTING.md
mjbrender: > Yes, if a PR is opened with 5 commits that was the work involved to fix a single issue, it should be rebased to a single commit. If a PR is opened with 5 commits, with the first three to fix one issue and the second two to fix a separate issue, then it should be rebased to two commits, one for each issue.

I'd like to elaborate on this section given that example. Thanks for that @geauxvirtual & @ConnorDoyle

ConnorDoyle: Cool, that just means we would also have to be more consistent as maintainers by insisting on single-commit PRs. I hadn't seen many requests to squash before merge, which is why I asked for the clarification. Good thing to explain here.

mjbrender: @ConnorDoyle agreed. Diligence will be important.

mjbrender:
@danielscottt I'd like to separate out any feature request, like the separate `/spec` folder, into another PR given this maps to work due this sprint regarding labels. How does the text / process look for the labeling?

danielscottt:
Lgtm

PR #752, opened 2016-03-09
Fixes #751

### LINKED ISSUE #751: Hangs when using example tasks/plugins
When running the example task under examples/tasks/mock-file.json It has an error:
```
ERRO[0080] error with processor job                      _module=scheduler-job block=run content-type=snap.gob error=plugin pool not found job-type=processor plugin-config=map[] plugin-name=passthru plugin-version=-1
```
This causes snapd to hang.  It doesn't respond to ctrl+c. The task is shown as still running in using `snapctl task list` but has only a single hit. It seems like it might be related to the changes from #743  since I do not get the error if I reset before that.

Steps to reproduce:
``` 
snapd -t 0 -l 1
```

Then load the plugins:
```
snapctl plugin load ../plugin/snap-collector-mock1
snapctl plugin load ../plugin/snap-collector-mock2
snapctl plugin load ../plugin/snap-publisher-file
snapctl plugin load ../plugin/snap-processor-passthru
```

Create the task:
```
snapctl task create -t examples/tasks/mock-file.json
```

jcooklin:
Summary of changes:
- removes pluginName and pluginVersion fields from processJob.   They are not needed since name and version were added to coreJob.

@intelsdi-x/snap-maintainers

### DIFF
--- a/scheduler/job.go
+++ b/scheduler/job.go
@@ -253,14 +253,12 @@ func (c *collectorJob) Run() {
 
 type processJob struct {
 	*coreJob
-	processor     processesMetrics
-	parentJob     job
-	metrics       []core.Metric
-	pluginName    string
-	pluginVersion int
-	config        map[string]ctypes.ConfigValue
-	contentType   string
-	content       []byte
+	processor   processesMetrics
+	parentJob   job
+	metrics     []core.Metric
+	config      map[string]ctypes.ConfigValue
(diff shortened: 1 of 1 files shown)


danielscottt:
LGTM pending good old t-rav.

geauxvirtual:
Verified on my system as well. LGTM

PR #769, opened 2016-03-15
Fix #768: Return and check error from *schedulerWorkflow.BindPluginContentType() method

### LINKED ISSUE #768: *schedulerWorkflow.BindPluginContentType() returns an error that's never checked
In scheduler/scheduler.go line 189, *schedulerWorkflow.BindPluginContentType() is called, but the err variable set in this call is never checked.

https://github.com/intelsdi-x/snap/blob/master/scheduler/scheduler.go#L189

In scheduler/workflow.go, *schedulerWorkflow.BindPluginContentType() calls an internal method bindPluginContentType(), but then directly returns nil instead of any error from bindPluginContentType().

https://github.com/intelsdi-x/snap/blob/master/scheduler/workflow.go#L233-L236

In bindPluginContentType(), when a publisher is included with a process node, bindPluginContentType() is called for the publish nodes (and any other process nodes), but any error from this call is never caught and returned.

https://github.com/intelsdi-x/snap/blob/master/scheduler/workflow.go#L274

geauxvirtual:
Fixes #768 

Summary of changes:
- Check for error returned by *schedulerWorkflow.BindPluginContentType during task creation
- Return output from bindPluginContentType() instead of nil in *schedulerWorkflow.BindPluginContentType()
- In bindPluginContentType(), check recursive call to bindPluginContentType() for an error and return the error if one is received.
- Moved the call to BindPluginContentType after validating metrics and plugins in the workflow. 

Testing done:
- Ran tests to verify tests pass.

@intelsdi-x/snap-maintainers

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -185,14 +185,20 @@ func (s *scheduler) createTask(sch schedule.Schedule, wfMap *wmap.WorkflowMap, s
 		return nil, te
 	}
 
-	// Bind plugin content type selections in workflow
-	err = wf.BindPluginContentTypes(s.metricManager)
-
 	// validate plugins and metrics
 	mts, plugins := s.gatherMetricsAndPlugins(wf)
 	errs := s.metricManager.ValidateDeps(mts, plugins)
 	if len(errs) > 0 {
 		te.errs = append(te.errs, errs...)
+		return nil, te
+	}
+
+	// Bind plugin content type selections in workflow
+	err = wf.BindPluginContentTypes(s.metricManager)
+	if err != nil {
+		te.errs = append(te.errs, serror.New(err))
+		f := buildErrorsLog(te.Errors(), logger)
+		f.Error("unable to bind plugin content types")
 		return nil, te
 	}
 
--- a/scheduler/workflow.go
+++ b/scheduler/workflow.go
@@ -231,8 +231,7 @@ type wfContentTypes map[string]map[string][]string
 
 // BindPluginContentTypes
 func (s *schedulerWorkflow) BindPluginContentTypes(mm managesPluginContentTypes) error {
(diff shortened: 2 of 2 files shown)


scheduler/scheduler.go
```diff
@@ -187,6 +187,12 @@ func (s *scheduler) createTask(sch schedule.Schedule, wfMap *wmap.WorkflowMap, s
 
 	// Bind plugin content type selections in workflow
 	err = wf.BindPluginContentTypes(s.metricManager)
+	if err != nil {
```
lynxbat: I had looked at this last week and wondered if it wouldn't be better to go ahead to the next block and capture if plugins are missing also. Return on either and/or.

The idea would be to return more of the failure reasons inclusive of binding problems and missing plugins.

geauxvirtual: Looking at the errors that were generated in the tests, and doing some testing, we should be binding the content types after validating the metrics and plugins. We already have a method to check if the metrics and plugins exist for the workflow, and we shouldn't try to bind plugin content types until after verifying all the plugins exist.

We could investigate making them more inclusive, but the content types really only focus on workflows that include a processor with additional processor and/or publisher nodes, correct? If the the processor or publisher plugin doesn't exist, there is no way to bind the plugin content types and check for errors. If a metric doesn't exist, but any processor or publisher plugins in the workflow do exist, then you could try to bind the content types and include any errors that may also exist. 

If the workflow does not include any processing, is there any reason to even call BindPluginContentTypes? Right now, if the workflow is collect -> publish, the metrics returned by collect get encoded in the publishJob.Run before PublishMetrics is called.

lynxbat: This solves my concern (reversing the total order).

IRCody:
LGTM

PR #770, opened 2016-03-15
Correcting example task

mjbrender:
* The current psutil w/o a processor is an exact copy of the psutil-file.yaml task and I'd like it to show an example without a processor
* I've added complex-example.yaml, which feeds from multiple collectors and branches into multiple publishers that's used in a presentation I give

Note, some of the Collectors used in the complex example are in development. They'll be out soon.

### DIFF
--- a/examples/tasks/psutil-file_no-processor.yaml
+++ b/examples/tasks/psutil-file_no-processor.yaml
@@ -1,24 +1,16 @@
---- # The Header starts after this line
-  version: 1        # All tasks, like plugins, have versioning
-  schedule:         # How frequently should we run this task?
-    type: "simple"  # Simple means run forever at the interval below
-    interval: "1s"  # This accepts seconds ('s') and milliseconds  ('ms')
-  workflow:           # For a small number of statistics, we have tests as small
-    collect:          # as 10ms. For a large number, more than 1s may be needed.
-      metrics:        # Collector caching defaults to 500ms.
-        /intel/psutil/load/load1: {}   # Here are the specific metrics we collect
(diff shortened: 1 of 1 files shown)


examples/tasks/complex-example.yaml
```diff
@@ -0,0 +1,55 @@
...
+       /intel/haproxy/info/MaxSessRate: {}
+       /intel/haproxy/info/MaxSslConns: {}
+       /intel/linux/docker/cpu_stats/cpu_usage/total_usage: {}
+       /intel/linux/docker/cpu_stats/cpu_usage/usage_in_kernelmode: {}
+       /intel/linux/docker/cpu_stats/cpu_usage/usage_in_usermode: {}
+       /intel/linux/docker/cpu_stats/throttling_data/periods: {}
+       /intel/linux/docker/cpu_stats/throttling_data/throttled_periods: {}
+       /intel/linux/docker/cpu_stats/throttling_data/throttled_time: {}
+     config:
+        /intel:
```
geauxvirtual: If this config section stays, there should be some comments about it. I find it odd to show a config that would apply for every `/intel` metric across different plugins that is actually never used by any plugin. What I see that happens is people will just copy the examples verbatim without knowing what everything exactly does, thinking everything is required when it's actually not.

mjbrender: That's a great point. I'm guilty of continuing that practice as I've seen it in example tasks since day one. Let's be more accountable for that across all examples @intelsdi-x/snap-maintainers

mjbrender:
I removed the second example I proposed and reduced this down to a correction of the ps-util example.

PR #778, opened 2016-03-18
snapctl: use default in `metric get`

danielscottt:
##### Summary of changes:
The flag for metric version has a default, but `snapctl metric get`
still required both namespace and version to be passed in. Now, the default
(0 / latest) will be used when no `-v` is passed.
##### Testing done:
Locally:
screen shot 2016-03-18 at 9 04 47 am

@intelsdi-x/snap-maintainers


geauxvirtual:
Rebase against master for the Go 1.4.3 tests to run.

geauxvirtual:
LGTM once tests are green.

PR #782, opened 2016-03-18
Fix #776

### LINKED ISSUE #776: Task can enter zombie state when plugin specifies sticky routing
When sticky routing is enabled by a plugin, if max-running-plugins is set to `n`, and `n+1` tasks are created that utilize the plugin with sticky routing, the `n+1` task enters a zombie state. `snapctl task list` will show the task as running with 1 hit, snapd will no longer exit cleanly when `ctrl + c` is issued, and trying to stop the task via snapctl results in snapctl just hanging until `ctrl + c` is issued. This results in having to issue `kill -9` to the snapd PID to shut it down.

When the `n+1` task is created, this is seen in the logs
```
ERRO[0011] could not select a plugin                     _block=findAvailablePlugin _module=control-routing error=0 of 1 plugins are available strategy=sticky
DEBU[0011] collector run completed                       _module=scheduler-job block=run job-type=collector metric-count=0
ERRO[0011] collector run error                           _module=scheduler-job block=run error=could not select a plugin job-type=collector
```

At this point, this task is a zombie, unable to be stopped, and not letting the snap daemon shut down properly.

geauxvirtual:
Fixes #776

Summary of changes:
- Remove `t.Lock()` and `t.Unlock()` calls from `*task.RecordFailures()` as these calls were trying to lock an already locked task causes tasks that fail to hang.

Testing done:
- Verified tasks that fail do not hang and continue to run per schedule defined in task until consecutive failure limit is reached.

@intelsdi-x/snap-maintainers

*NOTE* This needs PR #780 to be merged first.

### DIFF
--- a/scheduler/task.go
+++ b/scheduler/task.go
@@ -74,6 +74,7 @@ type task struct {
 	deadlineDuration   time.Duration
 	hitCount           uint
 	missedIntervals    uint
+	failureMutex       sync.Mutex
 	failedRuns         uint
 	lastFailureMessage string
 	lastFailureTime    time.Time
@@ -351,8 +352,8 @@ func (t *task) waitForSchedule() {
 // RecordFailure updates the failed runs and last failure properties
 func (t *task) RecordFailure(e []error) {
 	// We synchronize this update to ensure it is atomic
-	t.Lock()
-	defer t.Unlock()
+	t.failureMutex.Lock()
+	defer t.failureMutex.Unlock()
 	t.failedRuns++
 	t.lastFailureTime = t.lastFireTime
 	t.lastFailureMessage = e[len(e)-1].Error()


lynxbat:
Hold on removing this. We should handle the lock conflict path and not remove the safety here.

lynxbat:
A better fix would be to add another mutex for error updates and switch this to that lock. We still need to make error updating atomic and this current one does cause a deadlock.

geauxvirtual:
Added a separate mutex for recording failures.

danielscottt:
Since the fields `lastFailureTime` and `lastFailureMessage` are, by name, last write wins, would it be worth just using an atomic counter for `failedRuns` and forgoing the mutex altogether here?

lynxbat:
I would say no since we actually want consistency with the update per the convo @ConnorDoyle and I had on the previous thread.

lynxbat:
This is pretty cheap since it is almost always on an ending task.

ConnorDoyle:
+1 for updating those three members atomically.

Order of lock acquisition is still not guaranteed of course, so to @danielscottt's point we could move the error members to a new struct and then just do one atomic pointer update to avoid the extra lock.  @lynxbat already hit the point that there shouldn't be much contention on that mutex however so maybe it's not a big deal either way.

lynxbat:
Ok, up to you @danielscottt.

danielscottt:
@ConnorDoyle - I think @lynxbat is officially off into the wilderness.  Could you elaborate on this?
> consistency with the update

ConnorDoyle:
Sure.  If you have multiple threads executing `RecordFailures` in parallel then you have no guarantee of atomicity without the mutex.  e.g. the time, message, and count  fields might not correspond to a consistent cut.  The other suggestion was to say that if we introduced a new (immutable) type to hold the latest error data then `RecordFailures` could avoid locking by only updating a pointer in the task struct.  That seems to be no worse than the mutex as written since even though the mutex provides an atomic update, there's no guarantee about which failure "writes last" like you mentioned above.

danielscottt:
Thanks, @ConnorDoyle 

ship it, @geauxvirtual

PR #784, opened 2016-03-21
Cron scheduler

andrzej-k:
Implements #818. 

Summary of changes:
- Ability to use cron-like entries in task manifest file, for example:
```
{
    "version": 1,
    "schedule": {
        "type": "cron",
        "interval" : "@every 3s"
    },
    "workflow": {
``` 
or
```
{
    "version": 1,
    "schedule": {
        "type": "cron",
        "interval" : "0 * * * * *"
    },
    "workflow": {

```
More on supported cron expressions here: https://godoc.org/github.com/robfig/cron

Testing done:
- unit tests
- developer tests: snapctl workflow and task manifest, regression tests, different cron entries in task manifest

@intelsdi-x/snap-maintainers

### DIFF
--- a/cmd/snapctl/flags.go
+++ b/cmd/snapctl/flags.go
@@ -2,7 +2,7 @@
 http://www.apache.org/licenses/LICENSE-2.0.txt
 
 
-Copyright 2015 Intel Corporation
+Copyright 2015,2016 Intel Corporation
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -90,7 +90,7 @@ var (
 
 	flTaskSchedInterval = cli.StringFlag{
 		Name:  "interval, i",
-		Usage: "Interval for the task schedule [ex: 250ms, 1s, 30m]",
+		Usage: "Interval for the task schedule [ex (simple schedule): 250ms, 1s, 30m (cron schedule): \"0 * * * * *\"]",
 	}
 
 	flTaskSchedStartTime = cli.StringFlag{
--- a/pkg/schedule/cron_schedule.go
+++ b/pkg/schedule/cron_schedule.go
@@ -1,0 +1,137 @@
+/*
+http://www.apache.org/licenses/LICENSE-2.0.txt
+
+Copyright 2016 Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package schedule
+
+import (
+	"errors"
+	"time"
+
+	"github.com/robfig/cron"
+)
+
+// ErrMissingCronEntry indicates missing cron entry
+var ErrMissingCronEntry = errors.New("Cron entry is missing")
+
+// CronSchedule is a schedule that waits as long as specified in cron entry
+type CronSchedule struct {
+	entry    string
+	enabled  bool
+	state    ScheduleState
+	schedule *cron.Cron
+}
+
+// NewCronSchedule creates and starts new cron schedule and returns an instance of CronSchedule
+func NewCronSchedule(entry string) *CronSchedule {
+	schedule := cron.New()
+	return &CronSchedule{
+		entry:    entry,
+		schedule: schedule,
+		enabled:  false,
+	}
+}
+
+// GetState returns state of CronSchedule
+func (c *CronSchedule) GetState() ScheduleState {
+	return c.state
+}
+
+// Validate returns error if cron entry dosn't match crontab format
+func (c *CronSchedule) Validate() error {
+	if c.entry == "" {
+		return ErrMissingCronEntry
+	}
+	_, err := cron.Parse(c.entry)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Wait waits as long as specified in cron entry
+func (c *CronSchedule) Wait(last time.Time) Response {
+	var err error
+	now := time.Now()
+
+	// first run
+	if (last == time.Time{}) {
+		last = now
+	}
+	// schedule not enabled, either due to first run or invalid cron entry
+	if !c.enabled {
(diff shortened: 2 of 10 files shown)


cmd/snapctl/flags.go
```diff
@@ -90,7 +90,7 @@ var (
 
 	flTaskSchedInterval = cli.StringFlag{
 		Name:  "interval, i",
-		Usage: "Interval for the task schedule [ex: 250ms, 1s, 30m]",
+		Usage: "Interval for the task schedule [ex (simple schedule): 250ms, 1s, 30m (cron schedule): \"0 * * * * *\"]",
```
geauxvirtual: We should not overload the interval flag. A new flag should be created for handling cron pattern strings.

geauxvirtual: Disregard the above comment for now. After thinking about this more, interval may be ok.

geauxvirtual:
I've pulled this branch down to do some testing with the addition of the cron scheduler. One issue I've run into is that if a task is stopped and started again, the task doesn't actually start running again.

pkg/schedule/cron_schedule.go
```diff
@@ -0,0 +1,121 @@
...
+type CronSchedule struct {
+	entry    string
+	state    ScheduleState
+	schedule *cron.Cron
+}
+
+// NewCronSchedule creates and starts new cron schedule and returns an instance of CronSchedule
+func NewCronSchedule(entry string) *CronSchedule {
+	schedule := cron.New()
+	schedule.Start()
```
geauxvirtual: This cron scheduler introduces a difference in implementation from the simple and windowed schedule types since simple is just an interval from the time the task starts to the time the task stops, and windowed is running at a certain interval for a period of time. 

The implementation of the cron package being used for the cron schedule has actual methods for starting and stopping that are not timed based. The schedule should not start when the schedule is created, but when the task is started, and the schedule should stop when the task is stopped as the schedule is only valid when the task is running. In this instance, issuing `schedule.Start()` when the schedule is created is firing off a go routine with no ability to stop it without killing snapd. This may require adding additional methods for simple and windowed schedules `Start()` and `Stop()` respectively that for those schedule do nothing, but for cron schedule would actually start and stop the schedule. This would allow `task.Start()` and `task.Stop()` to call start and stop on any schedule without having to check if every task is a cron schedule or not.

pkg/schedule/cron_schedule.go
```diff
@@ -0,0 +1,121 @@
...
+			if next.After(now) {
+				break
+			}
+			misses++
+		}
+	}
+
+	// wait as long as specified in cron entry
+	done := make(chan struct{})
+	err := c.schedule.AddFunc(c.entry, func() { done <- struct{}{} })
```
geauxvirtual: Am I reading this wrong or does this add the same entry to the running scheduler for every call to Wait()?

andrzej-k:
Hey @geauxvirtual: thanks for your comments, I've made some updates to address them.

geauxvirtual:
This is looking better. Tasks can be restarted after being stopped, and a memory leak I was seeing when running more tasks than available plugins is no longer seen.

The only oddity I see now is when a task becomes disabled. When this task is re-enabled and started again, the misses counter is incremented the number of times the schedule wasn't run from when the task is disabled to when it was re-started. We normally just increment misses for a task when a workflow should run but missed the schedule window. When a task is stopped and started, misses is not incremented when the task is re-started, no matter how long it stayed in the stopped state.

andrzej-k:
Nice catch @geauxvirtual  - however simple schedule exhibited the same behavior (increment misses count after task became disabled, then enabled and started). The fix for this (misses counter reset) is in scheduler/task.go.

andrzej-k:
@intelsdi-x/snap-maintainers: I'd appreciate a follow-up review...

geauxvirtual:
This needs to be rebased down to a single commit. We should open an issue detailing the additions the cron scheduler brings to snap, and tag the commit as `Feature #XXX` to correspond to the issue that was opened.

andrzej-k:
Done @geauxvirtual

pkg/schedule/cron_schedule.go
```diff
@@ -0,0 +1,138 @@
...
+		err = c.schedule.AddFunc(c.entry, func() {})
+		if err == nil {
+			c.enabled = true
+		}
+	}
+
+	var misses uint
+	if c.enabled {
+		e := c.schedule.Entries()
+		if len(e) == 0 {
```
candysmurf: shouldn't you check the length of cron entries first before enabling to save some effort?

pkg/schedule/cron_schedule.go
```diff
@@ -0,0 +1,138 @@
...
+			err = ErrMissingCronEntry
```
candysmurf: even when an error occurred, just recording is needed, nothing else?

pkg/schedule/cron_schedule.go
```diff
@@ -0,0 +1,138 @@
...
+				next = s.Next(next)
+				if next.After(now) {
+					break
+				}
+				misses++
+			}
+
+			// wait
+			waitTime := s.Next(now)
+			time.Sleep(waitTime.Sub(now))
```
candysmurf: As this program use https://github.com/robfig/cron, does it has a delay method for leverage? Also, what do misses represent here?

andrzej-k: Regarding delay method - it was used in previous version of this code - but required starting and stopping cron service, which was more problematic (no way currently to stop gracefully) - for details see previous comments from @geauxvirtual - also feels like this version is more readable.
misses - as defined for Response interface (schedule.go) - Returns any intervals that were missed since the call to Wait()

pkg/schedule/cron_schedule.go
andrzej-k: setting schedule state to Error in updated version

pkg/schedule/cron_schedule.go
andrzej-k: this got refactored in newer version

andrzej-k:
Any more comments @intelsdi-x/snap-maintainers ?

candysmurf:
LGTM

scheduler/task.go
```diff
@@ -239,6 +239,7 @@ func (t *task) Enable() error {
 		return ErrTaskNotDisabled
 	}
 	t.state = core.TaskStopped
+	t.lastFireTime = time.Time{}
```
geauxvirtual: This line needs to be removed before merge. PR #808 resolves the issue with the missed column being incremented.

andrzej-k: @geauxvirtual Done.

PR #796, opened 2016-03-22
Fix #795: Return both cached and collected metrics

### LINKED ISSUE #795: collectMetrics only returns metrics collected from the plugin
When collectMetrics is called on an available plugin, the cache is checked and only metrics that do not hit the cache are collected from the plugin. However, the method only returns the metrics collected from the plugin even though a combined slice of cached and collected metrics are created.

https://github.com/intelsdi-x/snap/blob/master/control/available_plugin.go#L399-L408
https://github.com/intelsdi-x/snap/blob/master/control/available_plugin.go#L414

geauxvirtual:
Fixes #795

Summary of changes:
- Return results slice that includes cache and collected metrics instead of metrics slice that only included metrics returned from the plugin.

Testing done:
- Verified both cached and collected metrics are returned.

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/available_plugin.go
+++ b/control/available_plugin.go
@@ -411,7 +411,7 @@ func (ap *availablePlugins) collectMetrics(pluginKey string, metricTypes []core.
 	p.(*availablePlugin).hitCount++
 	p.(*availablePlugin).lastHitTime = time.Now()
 
-	return metrics, nil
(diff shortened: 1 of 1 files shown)


jcooklin:
LGTM

PR #801, opened 2016-03-23
give fixed names to gob registered types

### LINKED ISSUE #804: Vendoring in Go 1.6 breaks gob registration
When using vendoring in Go 1.6, reflection is in gob.Register to create
a key for type lookups.  This `reflect.PkgPath()` call returns a "globally
unique" key for looking up types, and in the case of a vendored type, is
not valid for matching gob registration names.  This is because on the
server side, the type is registered as `import/path.(typename)`, but on
the client side (vendored) it looks like
`client/vendor/import/path.(typename)`.  Using fixed keys bypasses the
need for reflection and is safe on either side, even when using
vendoring.

danielscottt:
Fixes #804 

When using vendoring in Go 1.6, reflection is used in gob.Register to create
a key for type lookups.  This `reflect.PkgPath()` call returns a "globally
unique" key for looking up types, and in the case of a vendored type, is
not valid for matching gob registration names.  This is because on the
server side, the type is registered as `import/path.(typename)`, but on
the client side (vendored) it looks like
`client/vendor/import/path.(typename)`.  Using fixed keys bypasses the
need for reflection and is safe on either side, even when using
vendoring.

### DIFF
--- a/control/plugin/client/native.go
+++ b/control/plugin/client/native.go
@@ -260,16 +260,16 @@ func newNativeClient(address string, timeout time.Duration, t plugin.PluginType,
 }
 
 func init() {
-	gob.Register(*(&ctypes.ConfigValueStr{}))
-	gob.Register(*(&ctypes.ConfigValueInt{}))
-	gob.Register(*(&ctypes.ConfigValueFloat{}))
-	gob.Register(*(&ctypes.ConfigValueBool{}))
+	gob.RegisterName("conf_value_string", *(&ctypes.ConfigValueStr{}))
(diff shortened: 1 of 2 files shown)


geauxvirtual:
Need an issue and commit message linked to that issue.

tjmcs:
Would be nice if this could be defined as a constant that is a map from the string to the type; then we could just define it in one place and use it wherever we need to use it...just a thought

jcooklin:
@danielscottt, once you update the commit as @geauxvirtual suggested, LGTM.

Issue #679, opened 2016-01-26
Add framework support for evaluating a query


jcooklin:
As a user when I ask for ```/acme/reactor/foo/*``` the framework should expand this to ```[’/acme/reactor/foo/in’,’/acme/reactor/foo/out’,’/acme/reactor/foo/inst’]``` **before calling the plugin**.  

If the users asks for ‘/acme/reactor/*/in the framework would request ```[‘/acme/reactor/foo/in’,’/acme/reactor/bar/in’]``` from the framework.

The framework should support a query that includes a tuple.  A query with a tuple would begin and end with brackets and be comma separated (e.g. ```[val1,val2]``` ).  So, as a user I could request ```/acme/reactor/*/[‘out’,’in’]``` and the framework would expand this to ```[‘/acme/reactor/foo/in’,’/acme/reactor/foo/out’,’/acme/reactor/bar/in’,’/acme/reactor/bar/out’]``` **before calling the plugin**. 

The framework should **not** attempt to evaluate a query if the plugin accepts a query at that location.  For example, as a user if I ask for ```/acme/coldfusion/cgroups/*/inst``` the request should not be evaluated by the framework **but passes it as is to the plugin which will evaluate the request at runtime** likely returning many metrics.

Given metric catalog:
```/acme/coldfusion/foo/in
/acme/coldfusion/foo/out
/acme/coldfusion/cgroups/*/inst
/acme/reactor/total_inst
/acme/reactor/foo/in
/acme/reactor/foo/out
/acme/reactor/foo/inst
/acme/reactor/bar/in
/acme/reactor/bar/out
/acme/reactor/bar/inst
```

*Note: Regex support should should also be considered

geauxvirtual:
> If the users asks for ‘/acme/reactor/*/in the framework would request
> [‘/acme/reactor/foo/in’,’/acme/reactor/bar/in’] from the framework.

How would the framework know what to request if the plugin only told the framework about one metric `/acme/reactor/*/in`?

I see this case as just passing `/acme/reactor/*/in` to the plugin and letting the plugin return all metrics that match this.

jcooklin:
@geauxvirtual, If i understand what you are saying correctly that's what I was trying to describe when I said

[...]

Issue #676, opened 2016-01-25
Support for dynamic metrics


andrzej-k:
**Definition:** 
Dynamic metrics is a metric which:

1. can appear or disapear between call to GetMetricTypes and call to CollectMetrics
  - example: Linux process, cgroup, VM, container
2. is not know at the time of GetMetricTypes call (aka. user defined metrics in task manifest)
  - example: metric is defined by user in a task manifest - e.g. User defines SQL query as part of plugin config, then is defining metric which is exposing query result
3. can depend on other dynamic metrics
  - example: specific metrics for specific Linux processes

**Configuration:**
Extensions in per metric (in task manifest), per plugin (in task manifest) and global (config file for snapd) configuration.

1. config per metric shall support more keys than just Version, at least also Source and Tags
  - example: specific metrics retrieved only from specific Sources (hosts), Tags allowing for per metric configuration
2. in single task manifest support having many "collect" sections to allow publishing one set of metrics to one destination and other metrics to other destination without a need of creating separate task manifest.

**Filtering:**
Extensions in filtering capabilities.

1. '*' should not be part of metric namespace, and should not be exposed on a metrics list. It should work on a task manifest level.
  - in case user requests all metrics for server (/intel/server) then framework shall trigger metrics gathering from all plugins with namespace starting with /intel/server
  - in case user requests filtering on pluign level (/intel/server/disk/*/octets_read) then plugin author shall provide support for it
2. it should be possible to replace any element of the namespace with * or other wildcard/reg exp
  - supported wildcards TBD
3. it should be possible to use wildcards multiple times in single namespace

**Metric listing:**
Extensions in metrics listing.

1. it should be possible for plugin to return more, or different, metrics than exposed in metric catalog
2. metric listing should support listing all metrics, listing just /namespace_prefix/* is not enough
  - possible solution: allow defining namespaces as ```/intel/procfs/<process_name>/cpu_utilization``` - then allow to filter on process name and/or return metrics which include names of processes at the time of metric gathering: /intel/procfs/process_A/cpu_utilization, /intel/procfs/process_B/cpu_utilization
3. for user defined metrics return information that metrics are defined by user or enable reading task manifest in GetMetricTypes()

**Returning metrics:**

1. If metrics target (for example: process or cgroup) is not longer available snap should log warning but the task should continue to run

jcooklin:
(In the **Definition** section)

 > 2. is not known at the time of GetMetricTypes call
 
[...]

PR #803, opened 2016-03-24
First implementation of dynamic query support

IzabellaRaulin:
Related to #676 

Summary of changes:
- adding dynamic query support (wildcard and tuple)
- adding validation of metric's namespace:
 - (part 1) not allowed characters, especially that ones which can be invalid interpreted as regexp tokens
 - (part 2) a wildcard in the end of namespace - that has been implemented yet - added after revising of all existing collector plugins; only one collector had such like metrics and was refactored. 

Testing done:
-  checking a query support for plugin with no-dynamic metrics 
-  checking a query support for plugin with dynamic metrics (if plugin accepts wildcards in this location the framework does not attempt to evaluate that)

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -2,7 +2,7 @@
 http://www.apache.org/licenses/LICENSE-2.0.txt
 
 
-Copyright 2015 Intel Corporation
+Copyright 2015-2016 Intel Corporation
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -60,6 +60,7 @@ var (
 
 	// ErrLoadedPluginNotFound - error message when a loaded plugin is not found
 	ErrLoadedPluginNotFound = errors.New("Loaded plugin not found")
+
 	// ErrControllerNotStarted - error message when the Controller was not started
 	ErrControllerNotStarted = errors.New("Must start Controller before calling Load()")
 )
@@ -109,6 +110,8 @@ type managesPlugins interface {
 
 type catalogsMetrics interface {
 	Get([]string, int) (*metricType, error)
+	GetQueriedNamespaces([]string) ([][]string, error)
+	MatchQuery([]string) ([][]string, error)
 	Add(*metricType)
 	AddLoadedMetricType(*loadedPlugin, core.Metric) error
 	RmUnloadedPluginMetrics(lp *loadedPlugin)
@@ -445,10 +448,31 @@ func (p *pluginControl) SwapPlugins(in *core.RequestedPlugin, out core.Cataloged
 	return nil
 }
 
+// MatchQueryToNamespaces performs the process of matching the 'ns' with namespaces of all cataloged metrics
+func (p *pluginControl) MatchQueryToNamespaces(ns []string) ([][]string, serror.SnapError) {
+	// carry out the matching process
+	nss, err := p.metricCatalog.MatchQuery(ns)
+	if err != nil {
+		return nil, serror.New(err)
+	}
+	return nss, nil
+}
+
+// ExpandWildcards returns all matched metrics namespaces with given 'ns'
+// as the results of matching query process which has been done
+func (p *pluginControl) ExpandWildcards(ns []string) ([][]string, serror.SnapError) {
+	// retrieve queried namespaces
+	nss, err := p.metricCatalog.GetQueriedNamespaces(ns)
+	if err != nil {
+		return nil, serror.New(err)
+	}
+	return nss, nil
+}
+
 func (p *pluginControl) ValidateDeps(mts []core.Metric, plugins []core.SubscribedPlugin) []serror.SnapError {
 	var serrs []serror.SnapError
 	for _, mt := range mts {
-		_, errs := p.validateMetricTypeSubscription(mt, mt.Config())
+		errs := p.validateMetricTypeSubscription(mt, mt.Config())
 		if len(errs) > 0 {
 			serrs = append(serrs, errs...)
 		}
@@ -506,7 +530,7 @@ func (p *pluginControl) validatePluginSubscription(pl core.SubscribedPlugin) []s
 	return serrs
 }
 
-func (p *pluginControl) validateMetricTypeSubscription(mt core.RequestedMetric, cd *cdata.ConfigDataNode) (core.Metric, []serror.SnapError) {
+func (p *pluginControl) validateMetricTypeSubscription(mt core.RequestedMetric, cd *cdata.ConfigDataNode) []serror.SnapError {
 	var serrs []serror.SnapError
 	controlLogger.WithFields(log.Fields{
 		"_block":    "validate-metric-subscription",
@@ -515,25 +539,26 @@ func (p *pluginControl) validateMetricTypeSubscription(mt core.RequestedMetric, 
 	}).Info("subscription called on metric")
 
 	m, err := p.metricCatalog.Get(mt.Namespace(), mt.Version())
+
 	if err != nil {
 		serrs = append(serrs, serror.New(err, map[string]interface{}{
 			"name":    core.JoinNamespace(mt.Namespace()),
 			"version": mt.Version(),
 		}))
-		return nil, serrs
+		return serrs
 	}
 
 	// No metric found return error.
 	if m == nil {
 		serrs = append(serrs, serror.New(fmt.Errorf("no metric found cannot subscribe: (%s) version(%d)", mt.Namespace(), mt.Version())))
-		return nil, serrs
+		return serrs
 	}
 
 	m.config = cd
 
 	typ, serr := core.ToPluginType(m.Plugin.TypeName())
 	if serr != nil {
-		return nil, []serror.SnapError{serror.New(err)}
+		return []serror.SnapError{serror.New(err)}
 	}
 
 	// merge global plugin config
@@ -549,19 +574,19 @@ func (p *pluginControl) validateMetricTypeSubscription(mt core.RequestedMetric, 
 	if m.policy.HasRules() {
 		if m.Config() == nil {
 			serrs = append(serrs, serror.New(fmt.Errorf("Policy defined for metric, (%s) version (%d), but no config defined in manifest", mt.Namespace(), mt.Version())))
-			return nil, serrs
+			return serrs
 		}
 		ncdTable, errs := m.policy.Process(m.Config().Table())
 		if errs != nil && errs.HasErrors() {
 			for _, e := range errs.Errors() {
 				serrs = append(serrs, serror.New(e))
 			}
-			return nil, serrs
+			return serrs
 		}
 		m.config = cdata.FromTable(*ncdTable)
 	}
 
-	return m, serrs
+	return serrs
 }
 
 func (p *pluginControl) gatherCollectors(mts []core.Metric) ([]core.Plugin, []serror.SnapError) {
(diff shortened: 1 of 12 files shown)


control/metrics.go
```diff
@@ -160,23 +189,162 @@ func (m *metricType) Timestamp() time.Time {
...
+
+	for _, chars := range notAllowedChars {
+		for _, ch := range chars {
+			if strings.ContainsAny(name, ch) {
+				return errorMetricContainsNotAllowedChars(ns)
+			}
+		}
+	}
+
+	// todo uncomment this after adjusting all plugins
```
geauxvirtual: These comments should be removed as the below lines are not commented out, nor should the lines be commented out.

control/metrics_test.go
```diff
@@ -307,3 +640,28 @@ func TestSubscriptionCount(t *testing.T) {
...
+		Convey("contains not allowed characters", func() {
+			ns := []string{"mock", "foo", "(bar)"}
+			err := validateMetricNamespace(ns)
+			So(err, ShouldNotBeNil)
+		})
+
+		// todo uncomment this after adjusting all plugins
+		// plugin should NOT advertise metrics ending in a wildcard
+		/*
+			Convey("ending in a wildcard", func() {
```
geauxvirtual: This test should be uncommented.

control/metrics.go
IzabellaRaulin: Thanks, I decided to remove this commented out code. 

When I finish revising that all plugins do not exposed a wildcard in the end of namespace + refactoring that one which has such like metrics, I will add this part of validation + unit test which covers that again in another pull request.

Already even plugin has metrics which ends with wildcard, it will works properly,  So, as a first step the plugins will be refactored

control/control.go
```diff
@@ -125,6 +128,31 @@ type managesSigning interface {
 	ValidateSignature([]string, string, []byte) error
 }
 
+type metric struct {
```
jcooklin: You can reuse metricType (defined [here](https://github.com/intelsdi-x/snap/blob/master/control/metrics.go#L63) in metrics.Go) instead of creating this definition.

control/control.go
```diff
@@ -814,7 +848,7 @@ func (p *pluginControl) FetchMetrics(ns []string, version int) ([]core.Cataloged
 }
 
 func (p *pluginControl) GetMetric(ns []string, ver int) (core.CatalogedMetric, error) {
-	return p.metricCatalog.Get(ns, ver)
+	return p.metricCatalog.GetMetric(ns, ver)
 }
 
 func (p *pluginControl) GetMetricVersions(ns []string) ([]core.CatalogedMetric, error) {
```
IRCody: Should GetMatchedMetrics be called here first so we can expand the namespace before calling into metricCatalog.GetVersions()?

control/control.go
```diff
@@ -445,10 +473,16 @@ func (p *pluginControl) SwapPlugins(in *core.RequestedPlugin, out core.Cataloged
 	return nil
 }
 
+// ExpandWildCards expands wildcards in metric namespace `ns' and returns all matched metrics namespaces based on cataloged metrics
+func (p *pluginControl) ExpandWildCards(ns []string) [][]string {
+	nss, _ := p.metricCatalog.GetMatchedMetrics(ns)
```
IRCody: Do we really want to ignore this error?

scheduler/scheduler.go
```diff
@@ -583,11 +584,14 @@ func (s *scheduler) gatherMetricsAndPlugins(wf *schedulerWorkflow) ([]core.Metri
 	)
 
 	for _, m := range wf.metrics {
-		mts = append(mts, &metric{
-			namespace: m.Namespace(),
-			version:   m.Version(),
-			config:    wf.configTree.Get(m.Namespace()),
-		})
+		nss := s.metricManager.ExpandWildCards(m.Namespace())
```
jcooklin: ExpandWildCards is calling control.GetMatchedMetrics and ignoring the error.  Consider returning the err so we can use the metrics directly from the workflow if err != nil.  

I found this issue when I was allowed to create a task with a single invalid metric.  The task creation should have failed with an error indicating an invalid metric was provided.

mgmt/rest/client/client_func_test.go
```diff
@@ -159,7 +159,10 @@ func TestSnapClient(t *testing.T) {
 				Convey("invalid task (missing metric)", func() {
 					tt := c.CreateTask(sch, wf, "baron", "", true)
 					So(tt.Err, ShouldNotBeNil)
-					So(tt.Err.Error(), ShouldContainSubstring, "Metric not found: /intel/mock/foo")
+					// todo
+					// Expected 'Metric not found: /intel/mock/foo' (but it didn't)
+					// Actual 'Plugin not found: type(publisher) name(file) version(-1)' to contain substring
+					//So(tt.Err.Error(), ShouldContainSubstring, "Metric not found: /intel/mock/foo")
```
IRCody: Do we know why this error has changed?

jcooklin: Yes, It's related to my comment in scheduler.go on gatherMetricsAndPlugins.

control/control.go
IzabellaRaulin: I suppose that metricCatalog.GetMetricVersions(ns []string) is called only for the cataloged metrics, not for "metrics to collect" defined in task manifest which can contain wildcard treated as a query.
@danielscottt - what do You think about it? The argument 'ns' does not come from task manifest, does it?

control/control.go
IzabellaRaulin: Thanks, done

control/control.go
IzabellaRaulin: Thanks, currently, as @jcooklin suggested, the err is returned and if err!= nil the metric from the workflow is used directly.

scheduler/scheduler.go
IzabellaRaulin: Thanks, implemented

mgmt/rest/client/client_func_test.go
IzabellaRaulin: Yes, exactly. Now it is resolved.

IzabellaRaulin:
Code after suggested changes. Feel free to comment this PR

control/control.go
jcooklin: I think this stems from the desire to "get" metrics using a wildcard through REST (i.e. `snapctl metric get -m "/intel/psutil/load/*" -v 5`).  @IRCody, is this correct?

Assuming that this is related to adding the ability to get list/get metrics using wild cards through REST we should consider adding that as a separate issue/PR (feature).

IzabellaRaulin: I agree with @jcooklin that this one should be consider as a feature issue and maybe as `snapctl metric **gets**`. 

Currently everything is consistent, because `get` returns details on a single metric as assumed
```
snapctl metric get --help
NAME:
   **get - get details on a single metric**
USAGE:
   command get [command options] [arguments...]

OPTIONS:
   --metric-version, -v '0'     The metric version. Default (0) is latest
   --metric-namespace, -m       A metric namespace
```

IzabellaRaulin: What is more, the dynamic metrics should be taken into consideration - that way the another command should be use.

Example:

Metric catalog:
/intel/procfs/processes/parked
/intel/procfs/processes/waking
/intel/procfs/processes/*/ps_vm
/intel/procfs/processes/*/ps_rss

`snapctl metric get /intel/procfs/processes/*/ps_vm` should return only information about this single metric

IRCody: After thinking about it I agree, it would be better as a separate issue. I opened #810 yesterday that is related.

My thought was just to move handling of dynamic queries out of the API. Currently the API handles queries with a trailing '*'  [here](https://github.com/intelsdi-x/snap/blob/master/mgmt/rest/metric.go#L43) which returns everything that starts with the same concrete namespace that is given.

scheduler/scheduler.go
```diff
@@ -589,11 +590,19 @@ func (s *scheduler) gatherMetricsAndPlugins(wf *schedulerWorkflow) ([]core.Metri
...
+		if err != nil {
+			// use metric directly from the workflow
+			nss = [][]string{m.Namespace()}
+		}
+
+		for _, ns := range nss {
+			mts = append(mts, &metric{
+				namespace: ns,
+				version:   m.Version(),
+				config:    wf.configTree.Get(m.Namespace()),
```
jcooklin: Loading mock1 and running the following task fails.
```
---
  version: 1
  schedule:
    type: "simple"
    interval: "1s"
  workflow:
    collect:
      metrics:
        /intel/*: {}
      config:
        /intel/mock:
          user: "root"
          password: "secret"
``` 

To fix this line should be `config:    wf.configTree.Get(ns),`.

candysmurf:
nice work! Do we have a doc that outlines the current behavior and the differences this enhancement may bring?

scheduler/scheduler.go
IzabellaRaulin: Yes, now it only works for "/intel/mock/*" and for others variations of queries (e.g like /intel/mock/(foo|bar)), but they have to start with "/intel/mock" to get this config. 

I looked on that and I suppose that turning `m.Namespace()` into `ns`  is not the only thing which needs to been done in this case. Look on [scheduler/job.go#L221](https://github.com/intelsdi-x/snap/blob/master/scheduler/job.go#L221) - I do not refactor this file (generally my approach was changing as little as possible), but might this should be changed too - according that, 2 things have to be refactored:
- [in scheduler/job.go func (c *collectorJob) Run()](https://github.com/IzabellaRaulin/snap/blob/dynamic_query_support/scheduler/job.go#L221)
- [in control/control.go func (p *pluginControl) CollectMetrics(..)](https://github.com/IzabellaRaulin/snap/blob/dynamic_query_support/control/control.go#L856)
@jcooklin - what do you think about it?

IzabellaRaulin:
@candysmurf - already there is only that: [docs/TASKS.md](https://github.com/IzabellaRaulin/snap/blob/dynamic_query_support/docs/TASKS.md#collect)

But did you mean a doc for this pull request or generally more detailed section in snap/doc?

scheduler/scheduler.go
marcin-krolik: test

IzabellaRaulin: After refactoring, for metric "/intel/*" in task manifest, the config section is retrieved properly

IzabellaRaulin:
Code after suggested changes by @jcooklin - after refactoring, for metric "/intel/*" in task manifest, the config section is retrieved properly. This fix required (ndirectly) the separation of `ExpandWildcard()` to  two functions, which are:

- `ExpandWildcards([]string ns) ([][]string, serror.SnapError)` -> now it only retrieves the matched metrics namespaces from `metricCatalog.mKey`, does NOT process the matching itself

- `MatchQueryToNamespaces([]string ns) ([][]string, serror.SnapError)` -> it performs the matching process. It is called in scheduler/scheduler.go, in gatherMetricsAndPlugins() when task is creating and metrics are retrieving from task manifest (they are matched with these ones in the catalog). The matching results are held in `metricCatalog.mKey` as before.

BEFORE there was only ExpandWildcards() which performed the matching process when there were not any matched results for this 'metric to collect' which is taken from task manifest.

jcooklin:
LGTM. :ship: it.

jcooklin:
Closes #679

IzabellaRaulin:
I've just added labels to show that this pull request is a good example of using that ones.

PR #808, opened 2016-03-25
Fixes #807 - Resets the lastFireTime on a task.Spin()

### LINKED ISSUE #807: Enabling and starting disabled task sees misses count incremented
When enabling and starting a task after a task becomes disabled, the misses count for the task will be incremented the number of times the task did not run per the schedule due to it being disabled. This would seem to skew statistics of a running task missing it's schedule with misses being incremented when the task wasn't running due to being disabled.

screen shot 2016-03-25 at 9 29 10 am

screen shot 2016-03-25 at 9 29 19 am

lynxbat:
Fixes #807

Summary of changes:
- Resets the lastFireTime for a task on calling Spin()
- Comments

@intelsdi-x/snap-maintainers

### DIFF
--- a/scheduler/task.go
+++ b/scheduler/task.go
@@ -207,6 +207,12 @@ func (t *task) Spin() {
 	// We need to lock long enough to change state
 	t.Lock()
 	defer t.Unlock()
+	// Reset the lastFireTime at each Spin.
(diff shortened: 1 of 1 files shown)


jcooklin:
:+1:

Issue #650, opened 2016-01-05
Source_ , Tags_ etc added in GetMetricTypes() are not visible in CollectMetric()


sandlbn:
I added Source and Tags fields to the PluginMetricType in GetMetricTypes() function. 

```golang
plugin.PluginMetricType{Namespace_: []string{"libvirt", hostname, uuid, "cpu", metric, Source_: hostname, Tags_: some_tags}
```
to avoid unnecessary calls in CollectMetric(), 
but those values are not visible in CollectMetrics function. 

```golang
func (p *Libvirt) CollectMetrics(mts []plugin.PluginMetricType) ([]plugin.PluginMetricType, error) {
    for _, p := range mts {
        fmt.Println(p.Source_)
        fmt.Println(p.Tags_)
    }
}
```
Source and Tags are empty. I can create buffer to cache values but I don't think so that this should be necessary. is it a bug or maybe I should populate Source, Tags and Labels in a different way. ?

jcooklin:
*Is it a bug or maybe I should populate Source, Tags and Labels in a different way?*

>It's currently working as designed.  The thinking being that the tags can change over time thus they are best managed by the plugin author.  This being said I'm not opposed to storing the tags and source in control so they can be sent on future Collect calls.

PR #809, opened 2016-03-25
forward tags and labels from the catalog to the plugin on collection

danielscottt:


geauxvirtual:
Commit needs to be prefixed with issue `Fix #650:`

jcooklin:
LGTM

PR #816, opened 2016-03-30
Update build docs

bodepd:
Fixes #

Summary of changes:
- 
- 
- 

Testing done:
- 

@intelsdi-x/snap-maintainers

### DIFF
--- a/docs/BUILD_AND_TEST.md
+++ b/docs/BUILD_AND_TEST.md
@@ -15,6 +15,7 @@ To build snap you'll need:
         git remote add origin git@github.com:<yourGithubID>/snap.git
         ```
         To push to your fork later, either do a `git push origin master` or do `git checkout -b <someBranchName>` then you can do `git push`.
+
 ## Build
 In the /snap directory there's a `Makefile` that builds all dependencies and snap.
 To get dependencies and build snap just run:  
--- a/docs/BUILD_AND_TEST.md
+++ b/docs/BUILD_AND_TEST.md
@@ -6,22 +6,49 @@ To build snap you'll need:
     * An option to look into is using the [go version manager (gvm)](https://github.com/moovweb/gvm) if you want to easily switch between Go versions.
 * [GNU Make](https://www.gnu.org/software/make/)
 * [git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
+
 * snap:
-    * `go get github.com/intelsdi-x/snap`. This will put snap in your `$GOPATH`.
-    * Fork the repository
-    * `cd` into your /intelsdi-x/snap directory on your machine and do the following to set your fork as the origin and make a branch on your fork
-        ```
-        git remote rename origin upstream
-        git remote add origin git@github.com:<yourGithubID>/snap.git
-        ```
-        To push to your fork later, either do a `git push origin master` or do `git checkout -b <someBranchName>` then you can do `git push`.
+    * Set your `$GOPATH` environment variable.
+
+    ```
+    export GOPATH=$HOME/work`
+    ```
+
+    * Install snap into your `$GOPATH`.
+
+    ```
+    go get github.com/intelsdi-x/snap`.
+    ```
+
+    * `cd` into `$GOPATH/src/github.com/intelsdi-x/snap` and set your fork as a remote repository.
+
+    ```
+    git remote rename origin upstream
+    ```
+
+    ```
+    git remote add origin git@github.com:<yourGithubID>/snap.git
+    ```
+
+    * To push to your fork later, either do a `git push origin master` or do `git checkout -b <someBranchName>` then you can do `git push`.
 
 ## Build
+
+In order to run the make command, you must first configure go and install
+godeps:
+
+```
+export PATH=$GOPATH/bin/
+go get github.com/tools/godep
+```
+
 In the /snap directory there's a `Makefile` that builds all dependencies and snap.
 To get dependencies and build snap just run:  
+
 ```
 make
 ```
+
 It runs the following `make deps` and `make all` commands.
 
 Alternatively, you can run `make` with any of the following targets:
--- a/docs/BUILD_AND_TEST.md
+++ b/docs/BUILD_AND_TEST.md
@@ -30,10 +30,13 @@ To build snap you'll need:
 
 ## Build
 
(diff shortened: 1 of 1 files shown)


docs/BUILD_AND_TEST.md
```diff
@@ -6,21 +6,49 @@ To build snap you'll need:
...
-    * `cd` into your /intelsdi-x/snap directory on your machine and do the following to set your fork as the origin and make a branch on your fork
-        ```
-        git remote rename origin upstream
-        git remote add origin git@github.com:<yourGithubID>/snap.git
-        ```
-        To push to your fork later, either do a `git push origin master` or do `git checkout -b <someBranchName>` then you can do `git push`.
+    * Set your `$GOPATH` environment variable.
+
+    ```
+    export GOPATH=$HOME/work`
```
tiffanyfj: This is assuming that the person has that directory. I definitely don't have my GOPATH there. Instead maybe we should direct people to the Go setup docs.

bodepd: Can I just genericize it to not make any assumptions about the location? I picked that location just b/c it was from the go setup docs.

I included the environment vars as opposed to referring to the golang setup docs just b/c those docs are very long and you only really need to set those couple of env vars to get it to work.

docs/BUILD_AND_TEST.md
```diff
@@ -6,21 +6,49 @@ To build snap you'll need:
...
+
+    * To push to your fork later, either do a `git push origin master` or do `git checkout -b <someBranchName>` then you can do `git push`.
+
 ## Build
+
+In order to run the make command, you must first configure go and install
+godeps:
+
+```
+export PATH=$GOPATH/bin/
```
candysmurf: I was able to make without doing this export.

docs/BUILD_AND_TEST.md
```diff
@@ -6,21 +6,49 @@ To build snap you'll need:
...
+go get github.com/tools/godep
```
candysmurf: this steps seems not needed as well.

docs/BUILD_AND_TEST.md
bodepd: how did godep get into your path? Can you show me `which godep`?

docs/BUILD_AND_TEST.md
bodepd: How did godep get installed. Is it possible that it only worked b/c it was not a completely clean install of go?

candysmurf: good question. I tried to use a new GOPATH, it didn't require me to explicitly get godeps.  I thought our script does that automatically. Correcting me if it does not.

docs/BUILD_AND_TEST.md
candysmurf: hmm, probably GVM did it for me. That's why I never need to do anything specific.

docs/BUILD_AND_TEST.md
bodepd: I added this commit to clarify that you only need to perform that mentioned steps if godeps is not already installed:
3bb2886113f72be1f20418e36370ae5f424847fa

bodepd: it probably just comes with gvm. I was going the straight ago install route

docs/BUILD_AND_TEST.md
bodepd: I added this commit to resolve this concern 11f01b6272fb755cd8de02fc21583c8fe950526e

docs/BUILD_AND_TEST.md
```diff
@@ -1,26 +1,54 @@
  
 # Build and Test
 ## Getting Started
+
 To build snap you'll need:
+
 * [Golang >= 1.4](https://golang.org)
```
candysmurf: do we still support go1.4?

docs/BUILD_AND_TEST.md
```diff
@@ -1,26 +1,54 @@
...
+
+Godeps is a dependency for running the make task(s) required for the build process. If it is not already installed,
+install it into your GOPATH and ensure that its built executables are in your PATH.
+
+```
+# first check to see if it is installed
+which godep
+# install it if necessary
+export PATH=$GOPATH/bin/
+go get github.com/tools/godep
```
candysmurf: After that, I seem always need to run this command. "godep save ./...". Do you like to let users know it?

docs/BUILD_AND_TEST.md
bodepd: I don't know :) I installed 1.6. That's definitely outside of the scope of these changes.

docs/BUILD_AND_TEST.md
bodepd: I didn't have to run that command, I can look into it since I'm doing a fresh build now

docs/BUILD_AND_TEST.md
candysmurf: building snap, it may not need that. That'e necessary for the new plugin repos.

docs/BUILD_AND_TEST.md
geauxvirtual: `godep save` is only required for updating dependencies or saving dependencies that you may have changed that need to be saved in the Godeps.json file. For building snap, you should just need to run `make`. `make` runs two scripts, deps.sh and build.sh. deps.sh will download godeps, so in a since, there is no need to direct users to manually download it.

candysmurf: agreed, for building snap, we don't need "godep save". 
Last time when I only ran "make", it didn't pick up the latest change. I had to run "make all". 
@geauxvirtual , can you elaborate when to use "make" vs "make all"?

geauxvirtual: The only difference between `make` and `make all` is `make` will set the dependencies then run `make all`. `make all` just builds snap and the plugins.

candysmurf:
LGTM

bodepd:
@tiffanyfj would you mind taking a quick look since you had comments before?

docs/BUILD_AND_TEST.md
mjbrender: We don't support 1.4 any further. 1.5.3+ is where I believe we are. Could you confirm @geauxvirtual ?

PR #822, opened 2016-03-30
Fixes #820 - Task fails when metrics come from different versions of …

### LINKED ISSUE #820: A plugin fails to start when a task references metrics from two versions of the same plugin.
Given two different versions of the same plugin where the first plugin contains a metric that is no longer exposed in version 2 the first plugin will not be started and the task fails.

**To recreate the scenario:**
* Start snap with the following config
```
---
control:
  plugins:
    collector:
      mock:
        versions:
          1:
            test: true
```
* Load the mock plugins (versions 1 and 2)
* Create and start the following task
```
---
  version: 1
  schedule:
    type: "simple"
    interval: "1s"
  workflow:
    collect:
      metrics:
        /intel/*: {}
      config:
        /intel/mock:
          user: "root"
          password: "secret"
```

jcooklin:
Fixes # 820

Summary of changes:
- Fixes issue where we override the plugin for a given metric

Testing done:
- Updated workflow test to cover this case
- Manual

Scenario:
* Start snap and load 2 different versions of the same plugin 
  * The lesser version should contain a metric that is not available in the new version
* Load a task that retrieves a metric that is only available in the older plugin

Snapd config which causes `/intel/mock/test` to become available:
```
---
control:
  plugins:
    collector:
      mock:
        versions:
          1:
            test: true
```

Task definition:
```
---
  version: 1
  schedule:
    type: "simple"
    interval: "1s"
  workflow:
    collect:
      metrics:
        /intel/mock/bar: {}
        /intel/mock/foo: {}
        /intel/mock/test: {}
      config:
        /intel/mock:
          user: "root"
          password: "secret"
```
Before:
before
After:
after

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/config.go
+++ b/control/config.go
@@ -84,6 +84,19 @@ func GetDefaultConfig() *Config {
 	}
 }
 
+// NewPluginsConfig returns a map of *pluginConfigItems where the key is the plugin name.
+func NewPluginsConfig() map[string]*pluginConfigItem {
+	return map[string]*pluginConfigItem{}
+}
+
+// NewPluginConfigItem returns a *pluginConfigItem.
+func NewPluginConfigItem() *pluginConfigItem {
+	return &pluginConfigItem{
+		cdata.NewNode(),
+		map[int]*cdata.ConfigDataNode{},
+	}
+}
+
 func newPluginTypeConfigItem() *pluginTypeConfigItem {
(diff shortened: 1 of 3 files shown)


candysmurf:
LGTM

PR #850, opened 2016-04-11
Config based routing

### LINKED ISSUE #539: RFC: Redux on Plugin Selection (Routing) and Caching
## Introduction
When a task is started in snap, it begins by _subscribing_ to a plugin _pool_.  This pool holds references to running instances of a given plugin.  If a subscription results in the need for a new plugin to be started, Snap's plugin runner starts one.  We call this need _eligibility_.  How plugins are chosen inside this pool we call _routing_.

In this specification I propose a redux on the routing module in Snap. This rethinking is predicated on the use cases enumerated in this document.  The existing implementation of plugin routing in Snap is not well suited for these use cases, so the result would be an almost complete overhaul of plugin _selection_ in Snap.

### Snap

Right now, the `apPool` type is the first class object when selecting a plugin.  I propose that a new type (`router`?, `selector`?), be the top-level type, and maintain the responsibilities of the current `apPool`: pooling, subscriptions, eligibility, and routing.  Modularizing the entire pool rather than just the router provides the necessary tools needed for a good implementation.  I propose that this implementation requires __all__ of this state to make intelligent selection decisions.

### Plugins

A plugin writer adds a new `plugin.Meta` key to their plugin's metadata called `routing`.  This key contains a value from an enum defined in the `plugin` package.  When a plugin is loaded its `routing` metadatum is applied to the selector for this plugin.

In order for the router to have access to the state it needs to correctly select the target plugin, new information will also have to be passed into the calls for `Collect`, `Process`, and `Publish`.  I propose that a new parameter be added to these function signatures: `taskId string`. 

### Cache

Related to plugin routing is caching.  Right now, the caching is contained in the client and caches metrics on a per metric namespace and version basis with a global expiration.  

In order for caching to be correct and effective it should be moved within the new type described here.  Further the call to this new type should expose CollectMetrics and not expose the running plugin or cache.  With the cache being moved to this new proposed type a plugin author could expose a minimum cache TTL and snap will instantiate the appropriate cache type based on the routing strategy described here (per plugin type and version, per running plugin or per config).

##### Description

A task needs to return to the same plugin each time, say a processor which is computing moving averages.  These MAs should not be polluted by other tasks data.  Since the implementation of the plugin is not capable of differentiating between one task and another, that task should have exclusive access to this plugin.

##### Proposed Implementation

A plugin writer sets their routing value to `plugin.StickyRouting`.  The task's ID can be used as a key for the plugin.  If the task is new, a new plugin is started.  If the task is new and it would cause the max pool size to be exceeded, an error is returned on task _creation_.  This error should come from an `eligible` call to the routing implementation, and bubble all the way back up to API caller.

##### Limitations

Snap's goal in maintaining the smallest possible footprint has many facets, one of which is the upper bound on the number of plugin instances in a given pool.  With a sticky router, the implementation must choose to either fail the creation of a task if this upper bound is exceeded, or exempt a selector which is of type sticky from this upper bound all together.  I propose the former.

##### Description

Take a publisher which opens, and then maintains a connection to RabbitMQ.  Tasks which share configuration data, like RabbitMQ node, credentials, and even queue/exchange data could share a plugin instance.  

##### Proposed Implementation

A plugin writer sets their routing value to `plugin.ConfigBasedRouting`. This could be achieved by hashing the config data, and then using that hash as a key for selecting a plugin instance.  This results in quite similar behavior to the sticky router described above, however it adds the advantage of avoiding the special exceptions for exceeding the max pool size.

##### Limitations

If a plugin _does_ require differentiating between tasks, the config-based router puts the onus on the plugin author to implement this logic.

marcin-krolik:
Fixes #539 

Summary of changes:
- removed SelectablePlugin interface, since AvailablePlugin interface can be used as well. SelectablePlugin did not introduce any usable abstraction in my opinion.
- aliased map[uint32]AvailablePlugin to MapAvailablePlugin to ease working with keys/values of map
- implemented naive config_based routing based on sticky routing implementation

@intelsdi-x/snap-maintainers


control/strategy/pool.go
```diff
@@ -373,15 +409,11 @@ func (p *pool) generatePID() uint32 {
 // MoveSubscriptions moves subscriptions to another pool
 func (p *pool) MoveSubscriptions(to Pool) []subscription {
 	var subs []subscription
-	// If attempting to move between the same pool
-	// bail to prevent deadlock.
-	if to.(*pool) == p {
-		return []subscription{}
-	}
```
IRCody: What was the reason to remove this?

control/strategy/pool.go
```diff
@@ -373,15 +409,11 @@ func (p *pool) generatePID() uint32 {
...
+
 	p.Lock()
 	defer p.Unlock()
+
 	for task, sub := range p.subs {
-		// ensure that this sub was not bound to this pool specifically before moving
```
IRCody: Same with this line.

control/strategy/pool.go
```diff
@@ -347,21 +359,45 @@ func (p *pool) SubscriptionCount() int {
...
+	}(opts)
+
+	if err != nil {
+		return nil, serror.New(err)
 	}
-	sap, err := p.Select(sp, taskID)
-	if err != nil || sap == nil {
+	aps := p.plugins.Values()
+	ap, err := p.Select(aps, selector)
+	if err != nil || ap == nil {
```
IRCody: If err and ap are nil this will a panic when someone tries to get the error from that serror. 

If both being nil isn't possible, I don't think both checks are needed.

control/strategy/strategy.go
```diff
@@ -28,27 +28,70 @@ import (
...
+		values = append(values, v)
+	}
+	return values
+}
+
+// HasKey checks if key exists in map
+func (sm MapAvailablePlugin) HasKey(key uint32) bool {
+	_, found := sm[key]
+	return found
+}
```
jcooklin: Where are the methods AddMap, RemoveAll, Size, Empty, Keys and HasKey used?  (If they aren't needed they should be removed.)

jcooklin:
> added SelectorValues struct for keeping "selectors" for available strategies in one place. Easy to extend, used to pass taskID and config to pool.SelectAP() method

Multiple arguments vs options struct - where do we draw the line? 
For me it generally comes down to code readability.  In this case I feel that `SelectorValues` decreases code readability.  `pool.Select` accepting the task id and configs is more readable and wouldn’t require the anonymous function in SelectAP.

control/strategy/pool.go
```diff
@@ -347,21 +359,45 @@ func (p *pool) SubscriptionCount() int {
 }
 
 // SelectAP selects an available plugin from the pool
-func (p *pool) SelectAP(taskID string) (SelectablePlugin, serror.SnapError) {
+func (p *pool) SelectAP(opts SelectorValues) (AvailablePlugin, serror.SnapError) {
```
jcooklin: > added SelectorValues struct for keeping "selectors" for available strategies in one place. Easy to extend, used to pass taskID and config to pool.SelectAP() method

Multiple arguments vs options struct - where do we draw the line? 
For me it generally comes down to code readability.  In this case I feel that `SelectorValues` decreases code readability.  `pool.Select` accepting the task id and configs is more readable and wouldn’t require the anonymous function in SelectAP.

marcin-krolik: Sure, I will extend ```pool.SelectAP()``` argument list and remove "selector"

control/strategy/pool.go
marcin-krolik: This change should not be removed. Thanks for noticing!

control/strategy/pool.go
marcin-krolik: Same as above

control/strategy/pool.go
```diff
@@ -227,6 +230,9 @@ func (p *pool) applyPluginMeta(a AvailablePlugin) error {
 	case plugin.StickyRouting:
 		p.RoutingAndCaching = NewSticky(cacheTTL)
 		p.concurrencyCount = 1
+	case plugin.ConfigRouting:
+		p.RoutingAndCaching = NewConfigBased(cacheTTL)
+		p.concurrencyCount = 3
```
jcooklin: Do you want to override the currency count here or allow it to be defined by the plugin author?

control/strategy/config_based_test.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"testing"
+	"time"
+
+	. "github.com/intelsdi-x/snap/control/strategy/fixtures"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestConfigBasedRouter(t *testing.T) {
+	Convey("Given a sticky router", t, func() {
```
jcooklin: Given a config router...

control/strategy/config_based_test.go
```diff
@@ -0,0 +1,55 @@
...
+			sp2, err := router.Select([]AvailablePlugin{p2, p1}, "cfg1")
+			So(err, ShouldBeNil)
+			So(sp1, ShouldNotBeNil)
+			So(sp2, ShouldEqual, p1)
+			// select the other (last) available plugin for task2
+			sp3, err := router.Select([]AvailablePlugin{p2, p1}, "cfg2")
+			So(err, ShouldBeNil)
+			So(sp3, ShouldNotBeNil)
+			So(sp3, ShouldEqual, p2)
+		})
```
jcooklin: It would be interesting to test the case that should error (a third cfg)

control/strategy/pool_test.go
```diff
@@ -0,0 +1,207 @@
...
+					),
+						func() {
+							So(pool.SubscriptionCount(), ShouldEqual, tc.Subscriptions)
+							So(pool.Eligible(), ShouldEqual, tc.Expected)
+						})
+				}
+			})
+		})
+	})
+}
```
jcooklin: I like the use of the fixture.

cc: @tjmcs

control/strategy/pool.go
marcin-krolik: Good point. I agree it could be useful to provide concurrency count in global config. Default value could be 1. 

I think I found an error in ```pool.Eligible()``` that for ```p.concurrencyCount != 1``` function ```pool.Eligibility()``` does not work properly. 

Main condition:

```go
should := p.SubscriptionCount() / p.concurrencyCount
if should > p.Count() && should <= p.max {
    return true
}
return false
```
```p.SubscriptionCount()``` and  ```p.concurrencyCount``` are both ```int```

Let's say ```p.Count() = 1```, ```p.max = 3```, ```p.SubscriptionCount() = 4``` and  ```p.concurrencyCount = 3``` then ```should = 1```

Main condition is ```false```, thus ```pool.Eligible() = false``` while in my opinion it should be ```true```

File: control/strategy/config_based_test.go (hunk 8 above)
Comment (marcin-krolik) [inline control/strategy/config_based_test.go]: Added

File: control/strategy/pool_test.go (hunk 9 above)
Comment (tjmcs) [inline control/strategy/pool_test.go]: So do I; we really need to get this pattern into the other tests we write in order to implement a real set of **small** tests as part of the **Testing Redux** epic; great job @marcin-krolik 👍

**Comment by marcin-krolik** [comment]
Tests fail because of #910

File: control/strategy/config_based.go [hunk 10]
```diff
@@ -0,0 +1,150 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/intelsdi-x/snap/core"
+)
+
+import (
+	"fmt"
+	"time"
+)
+
+// config-based provides a strategy that ... concurrency count is 1
```
danielscottt:
```go
// config-based provides a strategy that selects a plugin based on the given config
```

danielscottt:
Dude, Marcin, I love this! It looks so good.  Would you mind adding a test to `pool_test.go` around `SelectAP`?

marcin-krolik:
@danielscottt : Thanks! I wrote some additional tests for ```pool.SelectAP()```. take a look and say what you think!

danielscottt:
LGTM!

PR #855, opened 2016-04-14
Fix for #811

### LINKED ISSUE #811: The framework should not panic if a plugin (processor, publisher) doesn't accept the given content type

marcin-krolik:
Fixes #811 

Summary of changes:
- Removed panics for incorrect content type in processJob.Run()
- Now when content type is different then expected error is reported
@intelsdi-x/snap-maintainers


jcooklin:
LGTM

PR #872, opened 2016-04-22
Metric schema and metadata

### LINKED ISSUE #871: RFC: Metric schema metadata
This spec propose the following changes related to the metric type.
* Add (long) description describing metrics
* Add (short) description of namespace elements
* Add the field 'unit'  
* Remove the field 'source'
* Replace the field label with the same behavior on the namespace itself

## Key terms and concepts
* **Metric identifier** - The namespace provides the identify for the metric.
* **Namespace** - Uniquely identifies a metric using a tree structure (e.g. /intel/foo/bar/baz).
 * **Static** - Given a collector that exposes '/intel/sys/load' 
   * Query: /intel/sys/load
   * Result: single metric metric{/intel/sys/load, 1}
 * **Dynamic** - Given a collector that expose /intel/cpu/*/llc/misses' 
   * Query: /intel/cpu/*/llc/misses
   * Result: many; []metrics{metric{/intel/cpu/0/llc/misses, 123},metric{/intel/cpu/1/llc/misses, 456},...}
   * Query: /intel/cpu/0/llc/misses
   * Result: single metric metric{/intel/cpu/0/llc/misses, 123}
* **Tags** - Tags are key value pair that can be added to the metric at any point along the collect → process → publish pipeline.  It does not change or impact the metrics identity.  
* **Labels** - Labels assist in describing namespaces that contain data.  They provide a means to decode the namespace identifying components of the namespace that specify grouping and aggregation.    
 * Given the dynamic namespace "/intel/perf/pids/*/llc/misses" the associated label would be "label{name: "process id", index: 3}".  The label is of particular interest when publishing to InfluxDB and other similar databases.  For instance, given this example it is recommended that we publish to a series named /intel/perf/pids/llc/misses so that the metrics would be stored in a single series.  The reason being that you can merge the metrics using using influxdb tags within a series/measurement but it is not possible to merge (or join) across measurements.  In this example the series that would be written to would be /intel/perf/pids/llc/misses where an influxdb tags (process_id: xyz) could be created in the publisher to disambiguate the data points.  In summary, labels provide a means to decode data from the namespace enabling the appropriate transformations to publish data to the given data source.
* **Source** - The source field is the source for confusion (pun intended).  
This field does not contribute to the identity of the metric.  

## Use Cases ##

Use Cases
* As a **user** I should be provided all available meta for each metric in the catalog.
* As a **user** I should be provided a "long" description of the metric 
 * Given a metric in the catalog  the user should be provided what it measures, how it was derived and any other relevant details.
* As a **plugin author** I should have a means to identify if there is data encoded in the namespace.
* As a **plugin author** I should be able to communicate details about the metric that are stored in the metric cataloged and not sent along the collect → process → publish
 * This includes the long description of the metric as well as the (short) description stored in the label.  
* As a **user** I would like to the know the units for a given metric.  
* As a **framework developer** I would like to enforce that collectors provide a tag for units and that they are compliant with http://metrics20.org/spec/#units

## Proposal
* Add (long) description to the cataloged metric
* Merge the behavior achieved through a label into the namespace 
 * The namespace will be an array of NamespaceElements which can have fields for value, name and description.
* Add the field unit to the plugin metric type
* Remove the source field from the plugin metric type
 * What this field was communicating can be more effectively communicated through tags
* Add standard tags
 * Currently the only standard tag is `plugin_running_on`

jcooklin:
Fixes #871 - Metric schema metadata 

Summary of changes:
- [x] Add (long) description to a cataloged metric
- [x] Add (short) description of namespace elements
- [x] Add the field 'unit'
- [x] Remove the field 'source'
- [x] Add standard tags
- [x] Replace the field label with the same behavior on the namespace 

Testing done:
- unit
- functional
- manual

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -1035,20 +1035,20 @@ func (r *requestedPlugin) Config() *cdata.ConfigDataNode {
 // ------------------- helper struct and function for grouping metrics types ------
 
 // just a tuple of loadedPlugin and metricType slice
-type pluginMetricTypes struct {
+type metricTypes struct {
 	plugin      *loadedPlugin
 	metricTypes []core.Metric
 }
 
-func (p *pluginMetricTypes) Count() int {
+func (p *metricTypes) Count() int {
 	return len(p.metricTypes)
 }
 
 // groupMetricTypesByPlugin groups metricTypes by a plugin.Key() and returns appropriate structure
-func groupMetricTypesByPlugin(cat catalogsMetrics, metricTypes []core.Metric) (map[string]pluginMetricTypes, serror.SnapError) {
-	pmts := make(map[string]pluginMetricTypes)
+func groupMetricTypesByPlugin(cat catalogsMetrics, mts []core.Metric) (map[string]metricTypes, serror.SnapError) {
+	pmts := make(map[string]metricTypes)
 	// For each plugin type select a matching available plugin to call
-	for _, incomingmt := range metricTypes {
+	for _, incomingmt := range mts {
 		version := incomingmt.Version()
 		if version == 0 {
 			// If the version is not provided we will choose the latest
@@ -1058,7 +1058,7 @@ func groupMetricTypesByPlugin(cat catalogsMetrics, metricTypes []core.Metric) (m
 		if err != nil {
 			return nil, serror.New(err)
 		}
-		returnedmt := plugin.PluginMetricType{
+		returnedmt := plugin.MetricType{
 			Namespace_:          catalogedmt.Namespace(),
 			LastAdvertisedTime_: catalogedmt.LastAdvertisedTime(),
 			Version_:            incomingmt.Version(),
--- a/control/control.go
+++ b/control/control.go
@@ -1064,6 +1064,7 @@ func groupMetricTypesByPlugin(cat catalogsMetrics, metricTypes []core.Metric) (m
 			Version_:            incomingmt.Version(),
 			Tags_:               catalogedmt.Tags(),
 			Config_:             incomingmt.Config(),
+			Unit_:               catalogedmt.Unit(),
 		}
 		lp := catalogedmt.Plugin
 		if lp == nil {
--- a/control/control.go
+++ b/control/control.go
@@ -542,7 +542,7 @@ func (p *pluginControl) validateMetricTypeSubscription(mt core.RequestedMetric, 
 
 	if err != nil {
 		serrs = append(serrs, serror.New(err, map[string]interface{}{
-			"name":    core.JoinNamespace(mt.Namespace()),
+			"name":    mt.Namespace().String(),
 			"version": mt.Version(),
 		}))
 		return serrs
@@ -611,7 +611,7 @@ func (p *pluginControl) gatherCollectors(mts []core.Metric) ([]gatheredPlugin, [
 		m, err := p.metricCatalog.Get(mt.Namespace(), mt.Version())
 		if err != nil {
 			serrs = append(serrs, serror.New(err, map[string]interface{}{
-				"name":    core.JoinNamespace(mt.Namespace()),
+				"name":    mt.Namespace().String(),
 				"version": mt.Version(),
 			}))
 			continue
@@ -1059,16 +1059,15 @@ func groupMetricTypesByPlugin(cat catalogsMetrics, metricTypes []core.Metric) (m
 			return nil, serror.New(err)
 		}
 		returnedmt := plugin.PluginMetricType{
-			Namespace_:          incomingmt.Namespace(),
+			Namespace_:          catalogedmt.Namespace(),
 			LastAdvertisedTime_: catalogedmt.LastAdvertisedTime(),
 			Version_:            incomingmt.Version(),
 			Tags_:               catalogedmt.Tags(),
-			Labels_:             catalogedmt.Labels(),
 			Config_:             incomingmt.Config(),
 		}
 		lp := catalogedmt.Plugin
 		if lp == nil {
-			return nil, serror.New(errorMetricNotFound(incomingmt.Namespace().Strings()))
+			return nil, serror.New(errorMetricNotFound(incomingmt.Namespace().String()))
 		}
 		key := lp.Key()
 		pmt, _ := pmts[key]
--- a/control/control.go
+++ b/control/control.go
@@ -935,7 +935,7 @@ func (p *pluginControl) CollectMetrics(metricTypes []core.Metric, deadline time.
 			// Reapply standard tags after collection as a precaution.  It is common for
 			// plugin authors to inadvertently overwrite or not pass along the data
 			// passed to CollectMetrics so we will help them out here.
-			for i, _ := range m {
+			for i := range m {
 				m[i] = addStandardTags(m[i])
 			}
 			metrics = append(metrics, m...)
--- a/control/control.go
+++ b/control/control.go
@@ -109,19 +109,19 @@ type managesPlugins interface {
 }
 
 type catalogsMetrics interface {
-	Get([]string, int) (*metricType, error)
-	GetQueriedNamespaces([]string) ([][]string, error)
-	MatchQuery([]string) ([][]string, error)
+	Get(core.Namespace, int) (*metricType, error)
+	GetQueriedNamespaces(core.Namespace) ([]core.Namespace, error)
+	MatchQuery(core.Namespace) ([]core.Namespace, error)
 	Add(*metricType)
 	AddLoadedMetricType(*loadedPlugin, core.Metric) error
 	RmUnloadedPluginMetrics(lp *loadedPlugin)
-	GetVersions([]string) ([]*metricType, error)
-	Fetch([]string) ([]*metricType, error)
+	GetVersions(core.Namespace) ([]*metricType, error)
+	Fetch(core.Namespace) ([]*metricType, error)
 	Item() (string, []*metricType)
 	Next() bool
 	Subscribe([]string, int) error
 	Unsubscribe([]string, int) error
-	GetPlugin([]string, int) (*loadedPlugin, error)
+	GetPlugin(core.Namespace, int) (*loadedPlugin, error)
 }
 
 type managesSigning interface {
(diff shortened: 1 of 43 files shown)


control/metrics.go
```diff
@@ -322,18 +323,22 @@ func (mc *metricCatalog) removeMatchedKey(key string) {
 }
 
 // validateMetricNamespace validates metric namespace in terms of containing not allowed characters and ending with an asterisk
-func validateMetricNamespace(ns []string) error {
-	name := strings.Join(ns, "")
+func validateMetricNamespace(ns core.Namespace) error {
+	// name := strings.Join(ns, "")
```
IRCody: Should remove this.

core/metric.go
```diff
@@ -27,26 +27,132 @@ import (
...
+type Namespace []NamespaceElement
+
+// String returns the string representation of the namespace with "/" joining
+// the elements of the namespace.  A leading "/" is added.
+func (n Namespace) String() string {
+	ns := ""
+	for i, x := range n {
+		if i == 0 {
+			ns = "/"
+		}
```
IRCody: Can we just make `ns := "/"` and remove this or does something rely on String() returning `""` if no namespace is set?

core/metric.go
```diff
@@ -27,26 +27,132 @@ import (
...
+func (n Namespace) Key() string {
+	ns := ""
+	for i, x := range n {
+		ns += x.Value
+		if i < len(n)-1 {
+			ns += "."
+		}
+	}
+	return ns
+}
```
IRCody: This is the behavior of [strings.Join()](https://golang.org/pkg/strings/#Join). Would it make sense to use that here?

core/metric.go
```diff
@@ -27,26 +27,132 @@ import (
...
+
+// NewNamespace takes an array of strings and returns a Namespace.  A Namespace
+// is an array of NamespaceElements.  The provided array of strings is used to
+// set the corresponding Value fields in the array of NamespaceElements.
+func NewNamespace(ns []string) Namespace {
+	n := make([]NamespaceElement, len(ns))
+	for i, ns := range ns {
+		n[i] = NamespaceElement{Value: ns}
+	}
+	return n
```
IRCody: Do we need to worry about the input of []string{}?

core/metric.go
jcooklin: Yep.

core/metric.go
jcooklin: :)... also good call.

core/metric.go
jcooklin: We validate the namespace doesn't contain invalid characters before we add it to the metric catalog.

control/control.go
```diff
@@ -1035,20 +1035,20 @@ func (r *requestedPlugin) Config() *cdata.ConfigDataNode {
 // ------------------- helper struct and function for grouping metrics types ------
 
 // just a tuple of loadedPlugin and metricType slice
-type pluginMetricTypes struct {
+type metricTypes struct {
```
candysmurf: Overloading types here for the different packages. Is it a good practice in Golang?

control/control.go
```diff
@@ -1058,17 +1058,17 @@ func groupMetricTypesByPlugin(cat catalogsMetrics, metricTypes []core.Metric) (m
...
-			Namespace_:          incomingmt.Namespace(),
+		returnedmt := plugin.MetricType{
+			Namespace_:          catalogedmt.Namespace(),
 			LastAdvertisedTime_: catalogedmt.LastAdvertisedTime(),
 			Version_:            incomingmt.Version(),
 			Tags_:               catalogedmt.Tags(),
-			Labels_:             catalogedmt.Labels(),
 			Config_:             incomingmt.Config(),
+			Unit_:               catalogedmt.Unit(),
 		}
```
candysmurf: I'm curious the reason we didn't define "metricType" as following. If that's the case, there is no need to set each field.  Is it because MetricType public and for the rest?

type metricType struct {
	Plugin             *loadedPlugin
	subscriptions      int
	policy             processesConfigData
	config             *cdata.ConfigDataNode
	metric             *plugin.MetricType
}

control/control_test.go
```diff
@@ -123,117 +123,117 @@ func TestPluginControlGenerateArgs(t *testing.T) {
...
-				})
-				Convey("So swapped plugins event should show unloaded plugin version as 2", func() {
-					So(lpe.plugin.UnloadedPluginVersion, ShouldEqual, 2)
-				})
-				Convey("So swapped plugins event should show plugin type as collector", func() {
-					So(lpe.plugin.PluginType, ShouldEqual, int(plugin.CollectorPluginType))
-				})
-			})
-		})
+// func TestSwapPlugin(t *testing.T) {
```
tiffanyfj: This probably was commented out to save testing time or something, but should be uncommented.

control/control_test.go
```diff
@@ -624,33 +624,33 @@ func (m *mc) resolvePlugin(mns []string, ver int) (*loadedPlugin, error) {
...
 		}, nil
 	}
-	return nil, serror.New(errorMetricNotFound(ns))
+	return nil, serror.New(errorMetricNotFound(ns.String()))
 }
 
 func (m *mc) Subscribe(ns []string, ver int) error {
 	if ns[0] == "nf" {
-		return serror.New(errorMetricNotFound(ns))
+		return serror.New(errorMetricNotFound("/" + strings.Join(ns, "/")))
```
tiffanyfj: I think this and line 653 should be ns.String() like in 641.

control/plugin/collector_proxy.go
```diff
@@ -28,22 +28,22 @@ import (
 
 // Arguments passed to CollectMetrics() for a Collector implementation
 type CollectMetricsArgs struct {
```
candysmurf: It seems that struct type is not needed here but a type definition. Maybe there will be more fields be added later?

core/metric.go
```diff
@@ -27,26 +27,122 @@ import (
...
+			idx = append(idx, i)
+		}
+	}
+	return ret, idx
+}
+
+// NewNamespace takes an array of strings and returns a Namespace.  A Namespace
+// is an array of NamespaceElements.  The provided array of strings is used to
+// set the corresponding Value fields in the array of NamespaceElements.
+func NewNamespace(ns []string) Namespace {
```
candysmurf: It seems that only "NamespaceElement.Value" is used now. Would you please elaborate the potential use of Name field?

core/metric.go
```diff
@@ -27,26 +27,122 @@ import (
...
+}
+
+// AddStaticElement adds a static element to the given Namespace.  A static
+// NamespaceElement is defined by having an empty Name field.
+func (n Namespace) AddStaticElement(value string) Namespace {
+	nse := NamespaceElement{Value: value}
+	return append(n, nse)
+}
+
+// NamespaceElement provides meta data related to the namespace.  This is of particular importance when
```
candysmurf: If you can document the difference of Name and Value, that would be awesome.

candysmurf:
👍 all tests passed. Other than my minor questions, LGTM.

core/metric.go
IRCody: 👍

control/control.go
```diff
@@ -932,6 +932,12 @@ func (p *pluginControl) CollectMetrics(metricTypes []core.Metric, deadline time.
 
 	go func() {
 		for m := range cMetrics {
+			// Reapply standard tags after collection as a precaution.  It is common for
+			// plugin authors to inadvertently overwrite or not pass along the data
+			// passed to CollectMetrics so we will help them out here.
+			for i, _ := range m {
```
IRCody: Gofmt complains about this line on travis. Need to omit the _.

control/control_test.go
```diff
@@ -1075,13 +1078,13 @@ func TestCollectDynamicMetrics(t *testing.T) {
 			// pool should be the global cache expiration
 			So(ttl, ShouldEqual, strategy.GlobalCacheExpiration)
 			mts, errs := c.CollectMetrics([]core.Metric{m}, time.Now().Add(time.Second*1), taskID)
-			hits, err := pool.CacheHits(core.JoinNamespace(m.namespace), 2, taskID)
+			hits, err := pool.CacheHits(m.namespace.String(), 2, taskID)
```
tiffanyfj: Should this not just be m.NamespaceAsString() since we have the function? IDK if it really matters...

```go
func (m *metricType) Namespace() core.Namespace {
	return m.namespace
}

func (m *metricType) NamespaceAsString() string {
	return m.Namespace().String()
}
```

control/control_test.go
jcooklin: In both of these cases ns is a []string not a Namespace.

tiffanyfj: Yeah...just realized this right now when looking at the code more...

core/metric.go
jcooklin: I have the intent to write a blog about creating a plugin in which the plugin.MetricType will be discussed in detail.  The Namespace is a type alias for []NamespaceElement.  Namespace has methods AddStaticElement and AddDynamicElement which help authors create a metric.  Specifically

candysmurf: 👍

control/plugin/metric_test.go
```diff
@@ -51,193 +51,193 @@ func TestMetric(t *testing.T) {
...
 			So(m[0].Data(), ShouldResemble, 1)
-			So(strings.Join(m[1].Namespace(), "/"), ShouldResemble, "foo/baz")
+			So(m[1].Namespace().String(), ShouldResemble, "/foo/baz")
 			So(m[1].Data(), ShouldResemble, "2")
 		})
 
 		Convey("error on bad corrupt data", func() {
 			a = []byte{1, 0, 1, 1, 1, 1, 1, 0, 0, 1}
-			m, e = UnmarshallPluginMetricTypes("snap.gob", a)
+			m, e = UnmarshallMetricTypes("snap.gob", a)
```
tiffanyfj: This was from before...but Unmarshal has only one l

control/plugin/metric.go
```diff
@@ -281,7 +286,7 @@ func SwapPluginMetricContentType(contentType, requestedContentType string, paylo
 		}).Error("error while swaping")
```
tiffanyfj: Lol...swaping.

tiffanyfj:
Other than typos that were already there (so could be changed later...) 👍

IzabellaRaulin:
I've just added labels to show that this pull request is a good example of using `breaking-change/` labels and `status/needs-2nd-review` to highlight that specialized and in-depth review was needed

PR #876, opened 2016-04-25
delve might be reasonable alternative

iwankgb:
I have just started working on snap plugin and I thought that it might be reasonable to point users at alternative to GDB that is not very go-friendly.

@intelsdi-x/snap-maintainers


mjbrender:
👍 thanks for sharing @iwankgb. We appreciate having you here! Let me know if you need any help or give a shout through Gitter.

LGTM - merging.

PR #887, opened 2016-04-29
Refactored methods on core.Namespace: Strings() and String()

IzabellaRaulin:
**Before:**
even `len(ns)` was 0, `len(ns.Strings())` was 1

**Now:**
when `len(ns)` is 0, `len(ns.Strings())` also equals 0 and something like below in Fetch() is no more needed:
```
var nss []string
-	if ns.String() != "/" {
-		nss = ns.Strings()
-	}
-	mtsi, err := mc.tree.Fetch(nss)
```

[*] where`ns` is a metric namespace (core.Namespace type)

### Summary of changes:
- refactored methods on core.Namespace: Strings() and String()
- the fragment `strings.Split(ns.String(), "/")` was replaced with `ns.Strings()` 
 
### Testing done:
- run task /examples/task/mock-file.json with mock1 plugin
- checking output from `snapctl metric list` command which calls `Fetch()` 

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/metrics.go
+++ b/control/metrics.go
@@ -409,18 +409,15 @@ func (mc *metricCatalog) Get(ns core.Namespace, version int) (*metricType, error
 func (mc *metricCatalog) GetVersions(ns core.Namespace) ([]*metricType, error) {
 	mc.mutex.Lock()
 	defer mc.mutex.Unlock()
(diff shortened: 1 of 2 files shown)


jcooklin:
LGTM

PR #890, opened 2016-04-29
Fixes #889 - Moves the setting of the log level for snapd into an earlier call.

### LINKED ISSUE #889: Snapd does not honor log level settings from config or command line flags for startup log messages.
```
❯ go run ./snapd.go -t 0 -l 1                                                                                                                                                                                                                                                                     
INFO[0000] Starting snapd (version: unknown)            
INFO[0000] setting GOMAXPROCS to: 1 core(s)             
INFO[0000] Setting work manager queue size               _block=New _module=scheduler value=25
INFO[0000] Setting work manager pool size                _block=New _module=scheduler value=4
INFO[0000] Configuring REST API with HTTPS set to: false  _module=_mgmt-rest
INFO[0000] REST API is enabled                          
INFO[0000] control started                               _block=start _module=control
INFO[0000] module started                                _module=snapd block=main snap-module=control
INFO[0000] scheduler started                             _block=start-scheduler _module=scheduler
INFO[0000] module started                                _module=snapd block=main snap-module=scheduler
INFO[0000] Starting REST API on :8181                    _module=_mgmt-rest
INFO[0000] REST started                                  _block=start _module=_mgmt-rest
INFO[0000] module started                                _module=snapd block=main snap-module=REST
INFO[0000] setting plugin trust level to: disabled      
INFO[0000] auto discover path is disabled               
INFO[0000] snapd started 
```

It looks like the log level filtering is not set till after other operations are handled first.

lynxbat:
Fixes #889

Summary of changes:
- The logrus log level was being set after many of the initial startup calls that log are made.
- This fix just moves the log level setting to just after both command line flags and configuration file are loaded.

Testing done:
- Validation that logging outputs correctly.

@intelsdi-x/snap-maintainers

### DIFF
--- a/snapd.go
+++ b/snapd.go
@@ -221,6 +221,10 @@ func action(ctx *cli.Context) {
 	// same variables in that configuration
 	applyCmdLineFlags(cfg, ctx)
 
+	// Switch log level to user defined
+	log.Info("setting log level to: ", l[cfg.LogLevel])
+	log.SetLevel(getLevel(cfg.LogLevel))
+
 	// If logPath is set, we verify the logPath and set it so that all logging
 	// goes to the log file instead of stdout.
 	logPath := cfg.LogPath
@@ -506,10 +510,6 @@ func action(ctx *cli.Context) {
 			"block":   "main",
 			"_module": "snapd",
 		}).Info("snapd started")
-
-	// Switch log level to user defined
-	log.Info("setting log level to: ", l[cfg.LogLevel])
-	log.SetLevel(getLevel(cfg.LogLevel))
 
 	select {} //run forever and ever
 }


snapd.go
```diff
@@ -221,6 +221,10 @@ func action(ctx *cli.Context) {
 	// same variables in that configuration
 	applyCmdLineFlags(cfg, ctx)
 
+	// Switch log level to user defined
+	log.Info("setting log level to: ", l[cfg.LogLevel])
+	log.SetLevel(getLevel(cfg.LogLevel))
```
IRCody: Should we add the log.Info statement after setting the level so if they have specified a level > info they won't see this info statement?

 Also if you specify a log path this line will be printed to stdout while the rest goes to the logpath.

IRCody:
Because of the ordering this will cause snapd to panic if given an invalid log level (because the validation happens a few lines down @ [line 250](https://github.com/intelsdi-x/snap/blob/master/snapd.go#L250). getLevel() is not safe to call prior to validateLevelSettings() since it panics if the value is not in 1-5.

lynxbat:
You are right on the log validation. I will move it up.

I originally had this just buffering before we know the log level then dumping. I fell back to reordering but now I am thinking of going back to buffering till a Fatal or we get the level.

snapd.go
lynxbat: This validates tat buffering until a Fatal, exit, or the settings are done is a better path. It would honor both the log level and the log output.

lynxbat:
Closing this PR until I get time to refactor with buffering.

Issue #898, opened 2016-05-03
Add validation of keys in config


jcooklin:
When snapd is started with a global config file the user should receive an 
error if invalid keys are present.  Snapd should not start if an invalid 
config is provided.

IRCody:
I think we need a more specific definition for what makes a config invalid. Are all the sections required? If not what is the minimum set of options we are going to require?

Some things that definitely make a config invalid:
* Invalid/unknown keys
* Valid keys in invalid positions

PR #899, opened 2016-05-04
Improves error handling when parsing the global config file

### LINKED ISSUE #897: Need to improve error handling of config file parsing
The current error handling in the process of parsing the global configuration does not provide enough information to users in those cases where YAML or JSON configuration file is malformed. The reason for this is two-fold:

* First, the process that is followed to parse the configuration file as it stands today is somewhat simplistic; the code that parses the configuration file first attempts to parse the byte-stream read from the global configuration file as YAML and then, if that fails (for any reason), attempts to parse it as JSON. The result is that if there are any parsing errors that are the result of attempting to parse a malformed YAML byte-stream, those errors are silently ignored (and the program proceeds moves on and attempts to parse that malformed YAML byte-stream as a JSON byte-stream, which also fails).
* Second, the error that is returned in the case where the byte-stream read from the global configuration file cannot be parsed as either YAML or JSON is, to say the least, inadequate. Currently, if attempts to parse that byte-stream as both YAML and JSON fail, the user is simply presented with an error that states that there is an `Unknown file type` (with no additional information to tell them that it is even an issue with the file type for the configuration file).

We should work to make the error reported back to the user more informative (so that they have a better chance of determining what the error was that caused the parsing of the configuration file to fail) and to improve the parsing process so that it handles the two file types that we currently support (YAML and JSON) are handled better. I think that this can best be accomplished as follows:

* Add code to the current configuration file parsing process that detects the file type and then attempts to parse the file as either YAML or JSON depending on the file type that is detected rather than simply attempting to parse it as YAML (and, if that fails, simply attempting to parse it as JSON).
* Return a more informative error in the case where the format detected is not one of the supported formats (JSON or YAML).
* Modifying the existing code to return the error (if any) that is returned from the parsing routines back to the user. This should provide more information to the user about what the error was that caused the parsing to fail (including the fact that it's an error that occurred when parsing the global configuration file).

With those changes made, I think this issue can be marked as resolved. Without those changes, users will continue to get a rather cryptic `Unknown file type` error back from the program if they provide a malformed configuration file in either of the supported (JSON or YAML) formats.

tjmcs:
Fixes #897 and #898

Summary of changes:
- Improves error handling for both malformed YAML/JSON files and YAML/JSON files that cannot be parsed into our global configuration struct for some other reason (incorrect value types, for example)
- Adds error handling that catches and reports errors when unrecognized keys are detected while parsing the global configuration file

Testing done:
- Tested with a number of malformed YAML and JSON configuration files, configuration files with unrecognized keys, and configuration files that contain incorrect value types (a boolean where an integer is expected, for example) to ensure that all of these types of errors are caught and reported correctly.

With these changes in place, I believe that both of these issues can be considered resolved. We have discussed adding the ability to validate a global configuration file against an externally defined schema as part of Issue #898, but I think this feature would best be done by adding an additional option to the `snapctl` command (and the `snapd` RESTful API) that can be used to check the configuration file (rather than attempting to validate the global configuration file using an externally defined schema), but I could be convinced otherwise if others feel strongly that such validation needs to be added to this pull request before it can be merged.

It should also be noted that the parsing of the section of the global configuration file containing the options for the various plugins (in the `plugins` section under the `control` section of the global configuration file) remains unchanged by this pull request. The dynamic nature of that part of the configuration file makes the sort of error checking we've been adding here (and in the previous pull requests we've done around this global configuration file) a difficult task.

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/config.go
+++ b/control/config.go
@@ -103,23 +103,23 @@ func (c *Config) UnmarshalJSON(data []byte) error {
 		switch k {
 		case "max_running_plugins":
 			if err := json.Unmarshal(v, &(c.MaxRunningPlugins)); err != nil {
-				return err
+				return fmt.Errorf("%v (while parsing 'control::max_running_plugins')", err)
 			}
 		case "plugin_trust_level":
 			if err := json.Unmarshal(v, &(c.PluginTrust)); err != nil {
-				return err
+				return fmt.Errorf("%v (while parsing 'control::plugin_trust_level')", err)
 			}
 		case "auto_discover_path":
 			if err := json.Unmarshal(v, &(c.AutoDiscoverPath)); err != nil {
-				return err
+				return fmt.Errorf("%v (while parsing 'control::auto_discover_path')", err)
 			}
 		case "keyring_paths":
 			if err := json.Unmarshal(v, &(c.KeyringPaths)); err != nil {
-				return err
+				return fmt.Errorf("%v (while parsing 'control::keyring_paths')", err)
 			}
 		case "cache_expiration":
 			if err := json.Unmarshal(v, &(c.CacheExpiration)); err != nil {
-				return err
+				return fmt.Errorf("%v (while parsing 'control::cache_expiration')", err)
 			}
 		case "plugins":
 			if err := json.Unmarshal(v, c.Plugins); err != nil {
--- a/control/config.go
+++ b/control/config.go
@@ -75,6 +75,40 @@ type Config struct {
 	Plugins           *pluginConfig     `json:"plugins,omitempty"yaml:"plugins,omitempty"`
 }
 
+const (
+	CONFIG_CONSTRAINTS = `
+			"control" : {
+				"type": ["object", "null"],
+				"properties": {
+					"plugin_trust_level": {
+						"type": "integer",
+						"minimum": 0,
+						"maximum": 2
+					},
+					"auto_discover_path": {
+						"type": "string"
+					},
+					"cache_expiration": {
+						"type": "string"
+					},
+					"max_running_plugins": {
+						"type": "integer",
+						"minimum": 1
+					},
+					"keyring_paths" : {
+						"type": "string"
+					},
+					"plugins": {
+						"type": ["object", "null"],
+						"properties" : {},
+						"additionalProperties": true
+					}
+				},
+				"additionalProperties": false
+			}
+	`
+)
+
 // get the default snapd configuration
 func GetDefaultConfig() *Config {
 	return &Config{
--- a/control/config.go
+++ b/control/config.go
@@ -125,6 +125,8 @@ func (c *Config) UnmarshalJSON(data []byte) error {
 			if err := json.Unmarshal(v, c.Plugins); err != nil {
 				return err
 			}
+		default:
+			return fmt.Errorf("Unrecognized key '%v' in global config file while parsing 'control'", k)
 		}
 	}
 	return nil
--- a/mgmt/rest/server.go
+++ b/mgmt/rest/server.go
@@ -200,31 +200,31 @@ func (c *Config) UnmarshalJSON(data []byte) error {
 		switch k {
 		case "enable":
 			if err := json.Unmarshal(v, &(c.Enable)); err != nil {
-				return err
+				return fmt.Errorf("%v (while parsing 'restapi::enable')", err)
 			}
 		case "port":
 			if err := json.Unmarshal(v, &(c.Port)); err != nil {
-				return err
+				return fmt.Errorf("%v (while parsing 'restapi::port')", err)
 			}
 		case "https":
 			if err := json.Unmarshal(v, &(c.HTTPS)); err != nil {
-				return err
+				return fmt.Errorf("%v (while parsing 'restapi::https')", err)
 			}
 		case "rest_certificate":
 			if err := json.Unmarshal(v, &(c.RestCertificate)); err != nil {
-				return err
+				return fmt.Errorf("%v (while parsing 'restapi::rest_certificate')", err)
 			}
 		case "rest_key":
 			if err := json.Unmarshal(v, &(c.RestKey)); err != nil {
-				return err
+				return fmt.Errorf("%v (while parsing 'restapi::rest_key')", err)
 			}
 		case "rest_auth":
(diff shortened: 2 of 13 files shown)


tjmcs:
With the most recent commit in place, this is the sort of error output we see in response to a YAML configuration file with errors in three of the fields (two have values that are the wrong type, one has a value that is out of the bounds):
```bash
➜  snap git:(tb/fix-configFile-error-reporting) ✗ go run snapd.go --config ../broken-config2.yaml
ERRO[0000] Validate schema error                         context=(root)::gomaxprocs description=Invalid type. Expected: integer, given: boolean value=false
ERRO[0000] Validate schema error                         context=(root)::control::max_running_plugins description=Invalid type. Expected: integer, given: string value=string
ERRO[0000] Validate schema error                         context=(root)::log_level description=Must be less than or equal to 5 value=110
FATA[0000] Errors found while parsing global configuration file
exit status 1
➜  snap git:(tb/fix-configFile-error-reporting) ✗
```

tjmcs:
(and thanks to @IRCody for the assist on that last commit, too!)

control/config.go
```diff
@@ -75,6 +75,39 @@ type Config struct {
 	Plugins           *pluginConfig     `json:"plugins,omitempty"yaml:"plugins,omitempty"`
 }
 
+const (
+	CONFIG_CONSTRAINTS = `
+	"control" : {
+					"type": ["object", "null"],
+					"properties": {
```
IRCody: Can we fix the indentation here so it's not so much.

mgmt/rest/server.go
```diff
@@ -80,6 +80,38 @@ type Config struct {
...
+													"type": "string"
+									},
+									"rest_certificate": {
+													"type": "string"
+									},
+									"rest_key" : {
+													"type": "string"
+									},
+									"port" : {
+													"type": "integer"
```
IRCody: Are there any constraints we'd want to supply to the port. Maybe a minimum value?

mgmt/tribe/config.go
```diff
@@ -58,6 +59,32 @@ type Config struct {
...
+					"type": ["object", "null"],
+					"properties": {
+									"enable" : {
+													"type": "boolean"
+									},
+									"bind_addr": {
+													"type": "string"
+									},
+									"bind_port": {
+													"type": "integer"
```
IRCody: Same question here about minimum value for the port.

scheduler/config.go
```diff
@@ -36,6 +39,23 @@ type Config struct {
...
+	CONFIG_CONSTRAINTS = `
+	"scheduler": {
+					"type": ["object", "null"],
+					"properties" : {
+									"work_manager_queue_size" : {
+													"type": "integer"
+									},
+									"work_manager_pool_size" : {
+													"type": "integer"
+									}
```
IRCody: We should add minimum value of 1 for all these so we can error on 0/negative values.

snapd.go
```diff
@@ -161,6 +161,41 @@ type Config struct {
...
+                        "minimum": 1,
+                        "maximum": 5
+                },
+                "log_path": {
+                                "description": "path to log file for snapd to use",
+                                "type": "string"
+                },
+                "gomaxprocs": {
+                                "description": "value to be used for gomaxprocs",
+                                "type": "integer"
```
IRCody: Need to add minimum of 1 here as well.

mgmt/rest/server.go
lynxbat: Should be greater than zero at least.

snapd.go
```diff
@@ -551,9 +586,13 @@ func readConfig(cfg *Config, fpath string) {
...
 	}
 
-	err := cfgfile.Read(path, &cfg)
-	if err != nil {
-		log.Fatal(err)
+	serrs := cfgfile.Read(path, &cfg, CONFIG_CONSTRAINTS)
+	if serrs != nil {
+		for _, serr := range serrs {
+			log.WithFields(serr.Fields()).Error(serr.Error())
+			//log.Error(fmt.Sprintf("%v: %+v", serr.Error(), serr.Fields()))
```
IRCody: This line should be removed.

mgmt/rest/server.go
ConnorDoyle: FWIW binding to port 0 usually means any unused ephemeral port in Linux-land.

lynxbat: You are right. It would be a weird thing to do but since we output the port selected no reason to prevent it.

control/config.go
```diff
@@ -75,6 +75,39 @@ type Config struct {
...
+						"maximum": 2
+					},
+					"auto_discover_path": {
+						"type": "string"
+					},
+					"cache_expiration": {
+						"type": "string"
+					},
+					"max_running_plugins": {
+						"type": "integer"
```
IRCody: Probably need a min here also. Do we interpret 0 to mean unlimited? Not sure if it should be 0 or 1.

jcooklin:
There are several tests (control/config_test.go, mgmt/tribe/config_test.go, mgmt/rest/config_test.go) that call cfgfile.Read that need to be updated.  

LGTM, after the tests are fixed.

mgmt/rest/server.go
```diff
@@ -80,6 +80,40 @@ type Config struct {
...
+					"rest_certificate": {
+						"type": "string"
+					},
+					"rest_key" : {
+						"type": "string"
+					},
+					"port" : {
+						"type": "integer",
+						"minimum": 1,
+						"maximum": 65535
```
IRCody: Do we want to allow 0 for dynamic port selection?

jcooklin: Does it makes sense for the REST server to come up on a dynamic port?

IRCody: Idk but that is the current behavior. I think it's actually a bug in the current state b/c if you give it 0 it will select a port for you but then not output what port was selected.

jcooklin: Ok, sounds good.  @tjmcs let's set min for ports to zero (here and in tribe).  The issue #900 captures the current bug related to REST not printing to STDOUT the port that was selected.

IRCody:
LGTM

PR #904, opened 2016-05-06
Start of refactoring of Snap testing framework

tjmcs:
The changes in this pull request represent the start of what should be a number of additional pull requests that will gradually change the current test framework over to use the new `small`, `medium`, and `large` taxonomy. There is no issue in the issues list associated with this pull request that is fixed or resolved since the refactoring of the current test framework into the new taxonomy will take a considerable amount of time (and multiple pull requests).

Summary of changes:
- Modified the existing tests in the Snap codebase so that they are all tagged as `legacy` tests
- Created our first canonical example of a `small` test in the `pgk/cfgfile/cfgfile_small_test.go` file. In this test, there are no dependencies on external resources whatsoever. this new small test provides an example of mocking an external dependency in a function (specifically, mocking the call to the `ReadFile()` method from the `io/ioutil` package in the `cfgfile` package's `Read()` function).
- Modified the existing `scripts/tests.sh` shell script (which is run by the `make check` command in the `.travis.yml` file) so that our newly tagged `legacy` tests are run as any `small` tests (there is only one right now). As more `small` (and `medium`?) tests are added, we plan on phasing out the existing `legacy` tests from the TravisCI runs that are made when a pull request is received

Testing done:
- ran `go test` commands to test execution of both the `legacy` and `small` tests

@intelsdi-x/snap-maintainers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 sudo: false
 language: go
-matrix:
-  include:
-    - go: 1.5.4
-      env: GO15VENDOREXPERIMENT=1
-    - go: 1.6.2
+go:
+- 1.5.4
+- 1.6.2
 before_install:
 - bash scripts/gitcookie.sh
 - go get github.com/tools/godep
@@ -18,6 +16,10 @@ env:
   global:
     - SNAP_SOURCE=/home/travis/gopath/src/github.com/intelsdi-x/snap
     - SNAP_PATH=/home/travis/gopath/src/github.com/intelsdi-x/snap/build
+    - GO15VENDOREXPERIMENT=1
+  matrix:
+    - SNAP_TEST_TYPE=legacy
+    - SNAP_TEST_TYPE=small
 install:
 - export TMPDIR=$HOME/tmp
 - mkdir -p $TMPDIR
--- a/pkg/cfgfile/cfgfile.go
+++ b/pkg/cfgfile/cfgfile.go
@@ -31,41 +31,37 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 )
 
-// Read an input configuration file, parsing it (as YAML or JSON)
-// into the input 'interface{}'', v
-func Read(path string, v interface{}, schema string) []serror.SnapError {
-	// read bytes from file
-	b, err := ioutil.ReadFile(path)
-	if err != nil {
-		return []serror.SnapError{serror.New(err)}
-	}
-	// convert from YAML to JSON (remember, JSON is actually valid YAML)
-	jb, err := yaml.YAMLToJSON(b)
-	if err != nil {
-		return []serror.SnapError{serror.New(fmt.Errorf("error converting YAML to JSON: %v", err))}
-	}
-	// validate the resulting JSON against the input the schema
-	if errors := ValidateSchema(schema, string(jb)); errors != nil {
-		// if invalid, construct (and return?) a SnapError from the errors identified
-		// during schema validation
-		return errors
-	}
-	// if valid, parse the JSON byte-stream (above)
-	if parseErr := json.Unmarshal(jb, v); parseErr != nil {
-		// remove any YAML-specific prefix that might have been added by then
-		// yaml.Unmarshal() method or JSON-specific prefix that might have been
-		// added if the resulting JSON string could not be marshalled into our
-		// input interface correctly (note, if there is no match to either of
-		// these prefixes then the error message will be passed through unchanged)
-		tmpErr := strings.TrimPrefix(parseErr.Error(), "error converting YAML to JSON: yaml: ")
-		errRet := strings.TrimPrefix(tmpErr, "error unmarshaling JSON: json: ")
-		return []serror.SnapError{serror.New(fmt.Errorf("Error while parsing configuration file: %v", errRet))}
-	}
-	return nil
+// we're wrapping a function from the underlying io/ioutil package here so
+// that it's easier to mock when we define our *small* tests for this package...
+// first, define an interface for reading a file (by path)
+type reader interface {
+	ReadFile(s string) ([]byte, error)
 }
 
-// Validate the configuration (JSON) string against the input schema
-func ValidateSchema(schema, cfg string) []serror.SnapError {
+// then define a type (struct) that will implement the ReadFile method
+type cfgReaderType struct{}
+
+// then define the implementation for that type; it simply calls the
+// underlying ioutil.Readfile() method and returns the result
+func (r *cfgReaderType) ReadFile(s string) ([]byte, error) {
+	return ioutil.ReadFile(s)
+}
+
+// then define a private variable that is of type reader (interface)
+var cfgReader reader
+
+// now, to make things even more testable, do the same for the process of
+// validating the schema; first, create an interface to wrap the underlying
+// method that will be used to validate the schema
+type schemaValidator interface {
+	ValidateSchema(schema, cfg string) []serror.SnapError
+}
+
+// then, define a type (struct) that will validate the underlying schema
+type schemaValidatorType struct{}
+
+// and define an implementation for that type that performs the schema validation
+func (r *schemaValidatorType) ValidateSchema(schema, cfg string) []serror.SnapError {
 	schemaLoader := gojsonschema.NewStringLoader(schema)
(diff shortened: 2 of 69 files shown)


pkg/cfgfile/cfgfile.go
```diff
@@ -31,11 +31,76 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 )
 
+// we're wrapping a function from the underlying io/ioutil package here so
+// that it's easier to mock when we define our *small* tests for this package...
+// first, define an interface for reading a file (by path)
+type reader interface {
+	ReadFile(s string) ([]byte, error)
```
IRCody: Do we want an interface that so closely resembles the name of a std library interface, (reader vs Reader)?

tjmcs: Since it's internal to the package, I wasn't that worried about the name clashing; do you think it's an issue?

pkg/cfgfile/cfgfile_small_test.go
```diff
@@ -0,0 +1,207 @@
...
+const (
+	JSON = iota
+	YAML
+	FILE_NOT_FOUND
+	INVALID_YAML
+	UNMATCHED_SCHEMA
+)
+
+// create a type for the entries in our test-table
+type ttEntry struct {
```
IRCody: Maybe just Entry instead of ttEntry?

pkg/cfgfile/cfgfile_small_test.go
```diff
@@ -0,0 +1,207 @@
...
+func newTtEntry(b []byte, e error) ttEntry {
+	return ttEntry{b, e}
+}
+
+// then, define our test-table as a map of integers to ttEntry values
+var tt = map[int]ttEntry{
+	JSON:             newTtEntry(json.Marshal(testConfig{"Tom", "Justin"})),
+	YAML:             newTtEntry(yaml.Marshal(testConfig{"Tom", "Justin"})),
+	FILE_NOT_FOUND:   newTtEntry(nil, errors.New("File not found")),
+	INVALID_YAML:     newTtEntry([]byte("not:\tvalid: YAML"), nil),
```
IRCody: It's not obvious to me why this error would be nil if the entry was invalid YAML. What am I missing?

pkg/cfgfile/cfgfile.go
IRCody: Idk how big of an issue it is, maybe someone else will chime in. My first thought was why are we redefining a well known interface.

pkg/cfgfile/cfgfile_small_test.go
tjmcs: The INVALID_YAML entry is actually setup to return invalid YAML (as if invalid YAML had been read from a file); so the 'nil' error here means that we read the file successfully, not that the YAML was parsed successfully.

pkg/cfgfile/cfgfile_small_test.go
tjmcs: Happy to change it to `entry` instead of `ttEntry` if you feel that it makes things clearer; had named it with the `tt` prefix to try to link the purpose of that type to the `tt` (test table) variable that is defined a bit lower down in this file. Thoughts?

IRCody: I'd probably prefer having it just be testTable but definitely subjective.

pkg/cfgfile/cfgfile.go
```diff
@@ -31,11 +31,76 @@ import (
...
+	return serrors
+}
+
+// then define a private variable that is of type reader (interface)
+var cfgValidator schemaValidator
+
+// and finally, initialize to set the instance of that variable
+func init() {
+	cfgReader = &cfgReaderType{}
+	cfgValidator = &schemaValidatorType{}
```
marcin-krolik: What do you think about this idea. Instead of using package variables, let's define "constructor" functions like

```go
func NewReader() *cfgReaderType {
    return new(cfgReaderType)
}
```

This way, we would call 

```go
b, err := NewReader().ReadFile(path)
```

package variables created only for testing purposes seems somehow ugly to me, but maybe it's just me ;-)

pkg/cfgfile/cfgfile.go
jcooklin: I don't find it an issue.

mjbrender: Great longer explanation of this here too: http://relistan.com/writing-testable-apps-in-go/

pkg/cfgfile/cfgfile.go
jcooklin: @marcin-krolik:  We either need a package variable or the read function needs to accept a reader as an argument.

tjmcs: personally, I'd prefer the former to the latter; modifying the `Read(...)` function to accept an additional input argument just so that we can run our `small` tests seems less clean than defining the instance variables (the instance variables are local after all and are not visible outside of the package)

scripts/test.sh
```diff
@@ -71,13 +71,14 @@ test -z "$(goimports -l -d $TEST_DIRS | tee /dev/stderr)"
 # go test -race ./... - Lets disable for now
 
 # Run test coverage on each subdirectories and merge the coverage profile.
-echo "mode: count" > profile.cov
-
+echo "mode: count (legacy)" > profile.cov
```
geauxvirtual: This behavior of testing legacy or small tests needs to split in the test file and be triggered through changes to the .travis.yml file so that either legacy tests are run or small tests are run. A single test run in travis should not run both. This will make it cleaner to see a distinction between which tests are run and has failures in travis (legacy vs small), and allow us to see an increase in small tests as more tests get committed to the repo.

geauxvirtual:
Can we break this PR up into two separate PRs? 

One PR that tags all existing tests as legacy, updates our test script to run legacy and small tests, and updates .travis.yml to include a test matrix for running legacy and small tests separately.

Second PR that includes documentation around the size of tests and the small test example.

Currently, we can't start adding small tests to the framework until the first requirement is met.

tjmcs:
There are no plans to add additional documentation to the repository at this point to explain this PR, so I'm not sure what the benefit would be to break up this PR into two separate PRs, @geauxvirtual.

We could refactor the changes we're making in this PR to the `scripts/test.sh` script to run the `legacy` and `small` tests separately if you'd like, but this PR includes tagging of all of the existing tests as `legacy` tests, reworking the test matrix to run both the existing, `legacy` tests and the new `small` tests that we've added to the `pkg/cfgfile` package, and documentation around the changes we're making to the test framework to support this new testing taxonomy. As far as I'm concerned it's feature complete.

Is there a specific reason you'd like to see this PR broken out into two separate PRs other than having to wait for work that has already been done in this PR before we can proceed to add more small tests?

tjmcs:
changes made per suggestions from @IRCody and @geauxvirtual; any additional feedback?

.travis.yml
```diff
@@ -1,10 +1,8 @@
 sudo: false
 language: go
-matrix:
-  include:
-    - go: 1.5.4
-      env: GO15VENDOREXPERIMENT=1
-    - go: 1.6.2
+go:
+- 1.5.4
+- 1.6.2
```
IRCody: Was there a reason to remove GO15VENDOREXPERIMENT?

geauxvirtual:
Just need to make sure to capture any documentation changes regarding running of tests as scripts/test.sh now takes an argument when being run. By setting `SNAP_TEST_TYPE`, the makefile will pass in the env variable as the argument to the test.sh file. 

`scripts/test.sh small`
`scripts/test.sh legacy`
`SNAP_TEST_TYPE=small make test`
```
SNAP_TEST_TYPE=small
make test
```

.travis.yml
tjmcs: Just moved it to the global environment section; it doesn't matter to Go 1.6.x and will still be applied to Go 1.5.x. If we didn't move it we'd end up with two blocks labeled matrix with the chances made here to separate out the small and legacy test runs, so we thought we'd just collapse it down to one matrix block by moving that environment variable declaration

tjmcs:
I added a follow on commit with documentation changes, hopefully that's sufficient

IRCody:
LGTM

geauxvirtual:
LGTM

Issue #832, opened 2016-04-05
The task should immediately start collecting the new metrics after a plugin is loaded that shares the same namespace 


IzabellaRaulin:
The scenario which presents that:
Step1:  Load `PluginA`
Step2:  Create/start `Task1` with "intel/\*"
=> Task1 is collecting all metrics from PluginA

Step3a:  Load `PluginB1`
=> Task1 should immediately start collecting the new metrics from PluginB1
In current implementation (PR #803) the metrics are being collected only from pluginA

Step3b:  Load `PluginB2` which requires config
=> Task1 should immediately start collecting the new metrics from PluginB2, but Task1 does not contain any config for the pluginB2, so after 10 times not receiving the metrics from this plugin, the Task1 will fail  (@jcooklin approved that the initial task which was valid isn’t now so it seems reasonable that it would fail)

IzabellaRaulin:
Resolved among others by #1146

PR #907, opened 2016-05-10
Support specifying an instance of dynamic metric as requested metric in task manifest

### LINKED ISSUE #654: Task creation fails when specifying a specific instance of a dynamic metric.
With a dynamic metric, a plugin author can define a metric namespace as being `/foo/bar/*/qux`. When the plugin is loaded into snap, the metric catalog is populated with with a metric namespace of `/foo/bar/*/qux`.

This allows for collecting all instances that have a `qux`metric. However, to create a task that collects the `qux` metric for just `baz` by defining `/foo/bar/baz/qux` in the task manifest, the task will not be created as the metric catalog does not know about the metric `/foo/bar/baz/qux`, just `/foo/bar/*/qux`.

Creation of the task should succeed regardless if `/foo/bar/baz/qux` can be collected from the plugin. The task should fail on collect if the plugin returns an error trying to collect on `/foo/bar/baz/qux`.

IzabellaRaulin:
Fixes #654 and #832

### Summary of changes:
**a) support defining specific instance of dynamic metric in task manifest**

_Example:_ (mock collector plugin)
User might define in task manifest: 
- `/intel/mock/*/bar` to collect metric for all available instance
- or specific dynamic metric:
 - `/intel/mock/host0/bar` (in that case only one metric will be collected)
 - `/intel/mock/(host0|host1)/bar` (that returns two metrics)
 - `/intel/mock/host0/*` (that returns all available metrics for instance "host0" from mock collector)

In situation when there is no such instance `host0` error will be returned.

 **b) changing the way of matching dynamic query due to allowing the specifying of dynamic metrics, but there is no change from user perspective**

 **c) allow to immediately start collecting the new metrics by the running task from loaded plugin**
Read more #832

 **d) adding error message to trying Fetch() on empty metric catalog**

_Example:_
 running `snapctl metric list` when NO PLUGIN has been already loaded 
image

Testing done:
- unit tests (new unit tests added)
- running task with specified dynamic metric for mock plugin
- testing moving subscription to the latest available version of metric
- testing scenario described in #832 

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -109,9 +109,9 @@ type managesPlugins interface {
 }
 
 type catalogsMetrics interface {
-	Get(core.Namespace, int) (*metricType, error)
-	GetQueriedNamespaces(core.Namespace) ([]core.Namespace, error)
-	MatchQuery(core.Namespace) ([]core.Namespace, error)
+	GetMetric(core.Namespace, int) (*metricType, error)
+	GetMetrics(core.Namespace, int) ([]*metricType, error)
+	GetMatchedMetricTypes(core.Namespace, int) ([]*metricType, error)
 	Add(*metricType)
 	AddLoadedMetricType(*loadedPlugin, core.Metric) error
 	RmUnloadedPluginMetrics(lp *loadedPlugin)
@@ -448,40 +448,15 @@ func (p *pluginControl) SwapPlugins(in *core.RequestedPlugin, out core.Cataloged
 	return nil
 }
 
-// MatchQueryToNamespaces performs the process of matching the 'ns' with namespaces of all cataloged metrics
-func (p *pluginControl) MatchQueryToNamespaces(ns core.Namespace) ([]core.Namespace, serror.SnapError) {
-	// carry out the matching process
-	nss, err := p.metricCatalog.MatchQuery(ns)
-	if err != nil {
-		return nil, serror.New(err)
-	}
-	return nss, nil
-}
-
-// ExpandWildcards returns all matched metrics namespaces with given 'ns'
-// as the results of matching query process which has been done
-func (p *pluginControl) ExpandWildcards(ns core.Namespace) ([]core.Namespace, serror.SnapError) {
-	// retrieve queried namespaces
-	nss, err := p.metricCatalog.GetQueriedNamespaces(ns)
-	if err != nil {
-		return nil, serror.New(err)
-	}
-	return nss, nil
-}
-
-func (p *pluginControl) ValidateDeps(mts []core.Metric, plugins []core.SubscribedPlugin) []serror.SnapError {
+func (p *pluginControl) ValidateDeps(mts []core.Metric, cdt *cdata.ConfigDataTree, plugins []core.SubscribedPlugin) []serror.SnapError {
 	var serrs []serror.SnapError
-	for _, mt := range mts {
-		errs := p.validateMetricTypeSubscription(mt, mt.Config())
-		if len(errs) > 0 {
-			serrs = append(serrs, errs...)
-		}
-	}
-	if len(serrs) > 0 {
-		return serrs
+
+	errs := p.validateMetricsTypeSubscriptions(mts, cdt)
+	if len(errs) > 0 {
+		serrs = append(serrs, errs...)
 	}
 
-	//validate plugins
+	//validate plugins (publishers and processors)
 	for _, plg := range plugins {
 		typ, err := core.ToPluginType(plg.TypeName())
 		if err != nil {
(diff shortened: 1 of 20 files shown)


control/control.go
```diff
@@ -530,101 +505,136 @@ func (p *pluginControl) validatePluginSubscription(pl core.SubscribedPlugin) []s
...
 
-	// When a metric is added to the MetricCatalog, the policy of rules defined by the plugin is added to the metric's policy.
-	// If no rules are defined for a metric, we set the metric's policy to an empty ConfigPolicyNode.
-	// Checking m.policy for nil will not work, we need to check if rules are nil.
-	if m.policy.HasRules() {
-		if m.Config() == nil {
-			serrs = append(serrs, serror.New(fmt.Errorf("Policy defined for metric, (%s) version (%d), but no config defined in manifest", mt.Namespace(), mt.Version())))
-			return serrs
+		// merge global plugin config
+		if m.config != nil {
```
candysmurf: It seems that m.config == nil will never happen as in line 540, you handled it.

control/control.go
```diff
@@ -530,101 +505,136 @@ func (p *pluginControl) validatePluginSubscription(pl core.SubscribedPlugin) []s
...
 		}
-		ncdTable, errs := m.policy.Process(m.Config().Table())
-		if errs != nil && errs.HasErrors() {
-			for _, e := range errs.Errors() {
-				serrs = append(serrs, serror.New(e))
+
+		// When a metric is added to the MetricCatalog, the policy of rules defined by the plugin is added to the metric's policy.
+		// If no rules are defined for a metric, we set the metric's policy to an empty ConfigPolicyNode.
+		// Checking m.policy for nil will not work, we need to check if rules are nil.
+		if m.policy.HasRules() {
```
candysmurf: Adding dynamic metric policy rules, is it mandated for the plugin authors? None of our current collectors does it. Please elaborate what kind of policies Snap expects from the plugin authors.

control/control.go
```diff
@@ -886,22 +921,79 @@ func (p *pluginControl) GetMetricVersions(ns core.Namespace) ([]core.CatalogedMe
...
+				Version_:            mt.Version(),
+				Tags_:               mt.Tags(),
+				Config_:             config,
+				Unit_:               mt.Unit(),
+			}
+
+			// group matched metrics by plugin
+			pluginKey := lp.Key()
+			pmt, _ := pmts[pluginKey]
+			pmt.plugin = lp
```
candysmurf: what if "pmt" is nil for the pluginKey?

control/control.go
```diff
@@ -886,22 +921,79 @@ func (p *pluginControl) GetMetricVersions(ns core.Namespace) ([]core.CatalogedMe
...
-func (p *pluginControl) CollectMetrics(metricTypes []core.Metric, deadline time.Time, taskID string) (metrics []core.Metric, errs []error) {
+func (p *pluginControl) CollectMetrics(rmts []core.RequestedMetric, cdt *cdata.ConfigDataTree, deadline time.Time, taskID string) (metrics []core.Metric, errs []error) {
 
-	pluginToMetricMap, err := groupMetricTypesByPlugin(p.metricCatalog, metricTypes)
-	if err != nil {
-		errs = append(errs, err)
+	// gathers metrics to collect grouped by plugin
+	pluginToMetricMap, errs := p.gatherMetricTypesGroupedByPlugin(rmts, cdt)
+
+	if len(errs) > 0 {
```
candysmurf: just a simple return without returning the error?

control/metrics.go
```diff
@@ -202,120 +230,197 @@ func newMetricCatalog() *metricCatalog {
...
+	for index, n := range ns {
+		match := []string{}
+
+		if strings.ContainsAny(n, "*") {
+			// an asterisk covers all tuples cases
+			match = []string{"*"} // to avoid retrieving the same metric more than once
+		} else {
+			// a tuple is equivalent to regexp token (e.q match either a or b: '(a|b)')
+			// so it provides regular expression by itself
+			regex := regexp.MustCompile(n)
```
candysmurf: If this handles the tuple, it compiles on "n" and FindAllString on "n", it'll return an empty slice with the length 0. Are you sure that this is what you intend to do?

control/metrics.go
```diff
@@ -202,120 +230,197 @@ func newMetricCatalog() *metricCatalog {
...
+		}
 
-	// adding matched namespaces to map
-	mc.addItemToMatchingMap(wkey)
+		matchedItems[index] = append(matchedItems[index], match...)
 
-	return mc.matchedNamespaces(wkey)
-}
+		// number of possible combinations increases N=len(match) times
+		numOfPossibleCombinations = numOfPossibleCombinations * len(match)
```
candysmurf: Would you please explain the purpose of calculating the possibleCombinations here? Are you trying to filter returning metrics?

control/metrics.go
```diff
@@ -202,120 +230,197 @@ func newMetricCatalog() *metricCatalog {
...
+	}
+
+	//prepare two dimensional slice representing namespaces
+	nss := make([][]string, numOfPossibleCombinations)
+
+	for index := 0; index < len(ns); index++ {
+		items := matchedItems[index]
+		fillThreshold := len(items)
+
+		for i := 0; i < numOfPossibleCombinations; i++ {
```
candysmurf: numOfPossibleCombinations should be 0 all the time to me in this block.

control/control_test.go
```diff
@@ -1190,12 +1230,12 @@ func TestFailedPlugin(t *testing.T) {
...
 
-				var err []error
+				var errs []error
 				var cr []core.Metric
 				eventMap := map[string]int{}
 				for i := 0; i < MaxPluginRestartCount+1; i++ {
-					cr, err = c.CollectMetrics(m, time.Now().Add(time.Second*1), uuid.New())
-					So(err, ShouldNotBeNil)
+					cr, errs = c.CollectMetrics(rmts, cdt, time.Now().Add(time.Second*1), uuid.New())
+					So(errs, ShouldNotBeEmpty)
```
candysmurf: nice change.

control/control.go
```diff
@@ -911,8 +1003,43 @@ func (p *pluginControl) CollectMetrics(metricTypes []core.Metric, deadline time.
 
 	// For each available plugin call available plugin using RPC client and wait for response (goroutines)
 	for pluginKey, pmt := range pluginToMetricMap {
+
+		// if plugin does not have a pool, try to make new subscription
+		if !p.pluginRunner.AvailablePlugins().hasPool(pluginKey) {
+			// this happens only when the queried metric in task manifest matches with metrics
```
jcooklin: @IzabellaRaulin: It's the plan to move the logic here, dealing with plugins being loaded that may match a query, to the loaded plugin event handler, right?

control/control.go
IzabellaRaulin: By the way, this diff between master and this PR changes is very difficult to read even there is not much change in the code - just  fragment of code related with config has been put in `for` loop over metrics + removing the fragment as below: 

```
if m.Config() == nil {
serrs = append(serrs, serror.New(fmt.Errorf("Policy defined for metric, (%s) version (%d), but no config defined in manifest", mt.Namespace(), mt.Version())))
return serrs
```
because it has never happen - it was unreachable code. But as You caught out, the checking if `m.config !=nil` is exactly the same situation and does no sense any more. So, I will remove that too. Thanks

control/control.go
IzabellaRaulin: @candysmurf - I do not change this fragment of the code, it has been only moved into `for` loop.  But how it works is that if some rules are set in metric's policy, additional checks will be conducted - see it at master branch https://github.com/intelsdi-x/snap/blob/master/control/control.go#L574

So, I think it has to be kept in that form.

control/control.go
IzabellaRaulin: That fragment of code comes from snap master (https://github.com/intelsdi-x/snap/blob/master/control/control.go#L1080), I just renamed this function from **groupMetricTypesByPlugin** to **gatherMetricTypesGroupedByPlugin** to be already more accurate because of retrieving metricTypes from the catalog based on incoming requested metrics namespaces.  But returning to your question - if in "pmts" there is no items (first iteration) or there is no key="pluginKey", **pmt** will be an empty structure. So as the next step pmt.plugin is set (new or just re-set), and **returnedmt** will be appended to **pmt.metricTypes**, and item under **pmts[pluginKey]** will be added/updated. Only thing what can be changed is do not set pmt.lp again if a map contains pluginKey - but it is still not related to feature of specifying instance of dynamic metric. It should be treat  perhaps as improving the snap framework. What do You think?

control/control.go
IzabellaRaulin: As you can see at https://github.com/intelsdi-x/snap/blob/master/control/control.go#L900 I kept in that form as it was before.

control/control_test.go
IzabellaRaulin: :)

control/metrics.go
IzabellaRaulin: Sorry @candysmurf, but I am not rely sure what You ask about... Maybe I will explain how it works in two cases:

1) when the incoming namespace contains such tuples:
`ns = []string{"intel", "mock", ("host0|host1)", "(baz|bar)"}`

Iterate over each elements of slice and **using of an attribute of regular expression that (a|b) is token of regexp itself.**
Matching (the element of ns with oneself):
intel ->    []string{"intel"}
mock -> []string{"mock"}
(host0|host1) ->  []string{"host0", "host1"}
(baz|bar) ->  []string{"baz", "bar"}

So, in the end the number of possibleCombinations is 1\*1\*2\*2 = 4

Results:
 []string{"intel", "mock", "host0", "baz"}`
 []string{"intel", "mock", "host1", "baz"}`
 []string{"intel", "mock", "host0", "bar"}`
 []string{"intel", "mock", "host1", "baz"}`



2) when the incoming namespace contains tuple with an asterisk:   
 `ns = []string{"intel", "mock", ("host0|*)", "bar"}`

This is special case when user put an asterisk inside the tuple

Matching:
intel ->    []string{"intel"}
mock -> []string{"mock"}
(host0|\*) ->  []string{"\*"}   _// an asterisk covers all tuples cases_
(bar) ->  []string{"bar"}

So, in the end the number of possibleCombinations is 1\*1\*1\*1 = 1

Result:
 []string{"intel", "mock", *", "bar"}`

Of course the results are put into [][]string.

@candysmurf, I hope it's helpful, but if there is still some confusion do not hesitate to write to me. Thanks.

control/metrics.go
IzabellaRaulin: Nope, it's stricte math. Above there is explanation how it is counted. That is necessary to combine the elements into results ([][]string)

control/metrics.go
IzabellaRaulin: The minimum number of possible combination is 1 what means that even the incoming **ns** does not contain a tuple, the result ([][]string) has to contain that one **ns**.

control/control.go
IzabellaRaulin: @jcooklin, yes, it stricte related with discussion about pool subscription that we had and this should be removed from CollectMetrics. But the question is if we would like to:
- create new issue about mechanism which will validate metrics type subscriptions, subscribe collector deps and allow to create new pool not - only in the phase of creating task, but also during its running (for case when metrics exposed by already loaded plugin match with requested metrics namespaces)
- suspend this PR until all of that mentioned above will be implemented

control/control.go
candysmurf: Sounds good.

control/control.go
candysmurf: sounds good.

control/control.go
candysmurf: this is fine.

control/metrics.go
candysmurf: I have a doubt for this one. We can discuss when we meet tomorrow.

candysmurf:
good work, Iza. However, I found that the integration with Snap, it gives more metrics than expected. I would like to learn from you if any specifics as a plugin author needs to do in order to get the exact metrics.

IzabellaRaulin:
A plugin must be prepared that in CollectMetrics may receive:
a) metric with an asterisk which means getting metrics from all existing instances, e.g.: `/intel/mock/*/bar`

b) metric with specific name in that place of namespace which means getting metrics only from this instance, e.g.: `/intel/mock/host0/bar`

snap-plugin-collector-processes is a good example, https://github.com/intelsdi-x/snap-plugin-collector-processes/blob/master/processes/processes.go

candysmurf:
@IzabellaRaulin as we agreed in our code review meeting, here is the task manifest I used to reproduce issue we saw.
```
{
    "version": 1,
    "schedule": {
        "type": "simple",
        "interval": "1s"
    },
    "workflow": {
        "collect": {
            "metrics": {
               "/intel/redfish/v1/Chassis/1/Thermal#/Temperatures/1/Temperatures/2/UpperThresholdFatal":{}
            },
             "config": {
            },
            "process": [
                {
                    "plugin_name": "passthru",
                    "process": null,
                    "publish": [
                        {
                            "plugin_name": "file",
                            "config": {
                                "file": "/tmp/collected_redfish2"
                            }
                        }
                    ]
                }
            ]
        }
    }
}
```
Please reach out to me if you made the modification. I'll be happy to test it more for you.

control/control.go
IzabellaRaulin: Dealing with plugins being loaded that may match a query by loaded event handler has been implemented and it is in code review (PR https://github.com/intelsdi-x/snap/pull/997), so when that will be accepted/merged, I return to work on that (rebase code and check/fix issues which @candysmurf  reported)

IzabellaRaulin:
Closed due to #1220

PR #911, opened 2016-05-11
Fix for issue #418

### LINKED ISSUE #418: Specify what address for REST API to bind to
Currently the REST API will bind to all addresses on the system. A configuration parameter should be added to allow specifying what address to start the REST API on.

obourdon:
Fixes #418

Summary of changes:
- added new parameter for allowing binding address+port
- added new test cases
- removed extra space in command line argument specification
- added possibility for default port to be passed as environment variable

Please note that plural was used (bind_addresses and not bind_address) because I have anticipated the fact that we might want to bind on several different addresses + port in the future.
 
Testing done:
- reran all existing tests successfully
- added tests also successful

@intelsdi-x/snap-maintainers

### DIFF
--- a/core/serror/serror.go
+++ b/core/serror/serror.go
@@ -19,6 +19,14 @@ limitations under the License.
 
 package serror
 
+import (
+	"errors"
+)
+
+var (
+	ErrBadAddress = errors.New("Address for binding can not contain comma")
+)
+
 type SnapError interface {
 	error
 	Fields() map[string]interface{}
--- a/core/serror/serror.go
+++ b/core/serror/serror.go
@@ -19,14 +19,6 @@ limitations under the License.
 
 package serror
 
-import (
-	"errors"
-)
-
-var (
-	ErrBadAddress = errors.New("Address for binding can not contain comma")
-)
-
 type SnapError interface {
 	error
 	Fields() map[string]interface{}
--- a/mgmt/rest/server.go
+++ b/mgmt/rest/server.go
@@ -309,11 +309,11 @@ func (s *Server) Name() string {
 	return "REST"
 }
 
-func (s *Server) SetAddress(addrString string, dfltPort int) error {
+func (s *Server) SetAddress(addrString string, dfltPort int) {
 	restLogger.Info(fmt.Sprintf("Setting address to: [%v] Default port: %v", addrString, dfltPort))
 	// In the future, we could extend this to support multiple comma separated IP[:port] values
 	if strings.Index(addrString, ",") != -1 {
-		return serror.ErrBadAddress
+		restLogger.Fatal("Invalid address")
 	}
 	// If no port is specified, use default port
 	if strings.Index(addrString, ":") != -1 {
@@ -322,7 +322,6 @@ func (s *Server) SetAddress(addrString string, dfltPort int) error {
 		s.addrString = fmt.Sprintf("%s:%d", addrString, dfltPort)
 	}
 	restLogger.Info(fmt.Sprintf("Address used for binding: [%v]", s.addrString))
-	return nil
 }
 
 func (s *Server) Start() error {
--- a/mgmt/rest/server.go
+++ b/mgmt/rest/server.go
@@ -52,6 +52,7 @@ const (
 const (
 	defaultEnable          bool   = true
 	defaultPort            int    = 8181
+	defaultBindAddresses   string = ""
 	defaultHTTPS           bool   = false
 	defaultRestCertificate string = ""
 	defaultRestKey         string = ""
@@ -73,6 +74,7 @@ var (
 type Config struct {
 	Enable           bool   `json:"enable,omitempty"yaml:"enable,omitempty"`
 	Port             int    `json:"port,omitempty"yaml:"port,omitempty"`
+	BindAddresses    string `json:"bind_addresses,omitempty"yaml:"bind_addresses,omitempty"`
 	HTTPS            bool   `json:"https,omitempty"yaml:"https,omitempty"`
 	RestCertificate  string `json:"rest_certificate,omitempty"yaml:"rest_certificate,omitempty"`
 	RestKey          string `json:"rest_key,omitempty"yaml:"rest_key,omitempty"`
@@ -107,6 +109,9 @@ const (
 						"type": "integer",
 						"minimum": 0,
 						"maximum": 65535
+					},
+					"bind_addresses" : {
+						"type": "string"
 					}
 				},
 				"additionalProperties": false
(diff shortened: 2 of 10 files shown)


IzabellaRaulin:
Travis tests fail  - please, make `go fmt snapd.go`  and, just in case, also do that for others go files that have been changed in this commit.

lynxbat:
Since it doesn't support multiple addresses yet, we avoid plurals, and for the sake of shorter switches; I would recommend renaming this new switch to `bind-addr`.

lynxbat:
Also - could we instead use an `--ip-addr` switch and have it include port?

```
# Just port
snapd --port 8282
```

```
# IP and Port to bind to
snapd --ip-addr 127.0.0.1:8282
```


```
# If the user provides both we error to make them choose one
snapd --ip-addr 127.0.0.1:8383 --port 8383
> Error: Please use either --ip-addr or --port but not both.
```

lynxbat:
The `ip-addr` example above would also align well to the way to implement multiple listening IPs in golang: https://play.golang.org/p/2IrtRqdGit

obourdon:
Thanks for all remarks. I'll modify this tomorrow French time ;-) and submit the modified version once tested

lynxbat:
This is a great PR. Thanks for coming up with this improvement!

obourdon:
I am almost done with the modifications and tests however I might have "dummy/basic" questions to ask.

- to be completely coherent with both your suggested parameter name (ip-addr) and the already existing port parameters I think it is best to call it addr and not ip-addr because of the possible IP:PORT syntax. However as the -a parameter is already taken by the auto-discover-path I currently left it to be -b (like for previous bind-address[es]) but I do not mind changing it to anything you would consider more appropriate

- I need to have a errors.New("MY MESSAGE") variable which needs to be common between snapd.go and mgmt/rest/server.go. For now, I have included this variable in core/serror/serror.go but I am not convinced that it should really be here. After digging more to see if I could infer something from other parts of the code, I did not manage to find anything which could help me so any suggestion is welcome

- doing the check as requested for removing potential ambiguity between -b IP:PORT1 and -p PORT2 I also need to leave the snapd.go main program "as gracefully as possible" and due to the somewhat old version of codegangsta/cli used here I can not have access to cli.NewExitError(). Again, looking deeper into other part of the code, I could not find anything which could fit this requirement

- My past experiences make me believe that most people are used to the fact that specific value in IP:PORT1 is taking precedence over a more generic -p PORT2 option and that documentation might help making sure people fully understand this but I may be wrong

- last, I have modified the SetAddress function from mgmt/rest/server.go to return an error if for some reason the string passed is not valid

Feel free to comment on all these at will
Thanks again and best regards

snapd.go
```diff
@@ -645,9 +653,37 @@ func setDurationVal(field time.Duration, ctx *cli.Context, flagName string) time
...
 }
 
+//
+func checkCmdLineFlags(ctx *cli.Context) error {
+	// Bind address is specified
+	if ctx.IsSet("api-addr") {
+		addr := ctx.String("api-addr")
+		// Contains a comma
+		if strings.Index(addr, ",") != -1 {
+			return serror.ErrBadAddress
```
jcooklin: Let's consider returning errors.New("Invalid address") here.

mgmt/rest/server.go
```diff
@@ -299,8 +309,20 @@ func (s *Server) Name() string {
 	return "REST"
 }
 
-func (s *Server) SetAddress(addrString string) {
-	s.addrString = addrString
+func (s *Server) SetAddress(addrString string, dfltPort int) error {
+	restLogger.Info(fmt.Sprintf("Setting address to: [%v] Default port: %v", addrString, dfltPort))
+	// In the future, we could extend this to support multiple comma separated IP[:port] values
+	if strings.Index(addrString, ",") != -1 {
+		return serror.ErrBadAddress
```
jcooklin: Let's consider returning errors.New("Invalid address") here.

core/serror/serror.go
```diff
@@ -19,6 +19,14 @@ limitations under the License.
 
 package serror
 
+import (
+	"errors"
+)
+
+var (
+	ErrBadAddress = errors.New("Address for binding can not contain comma")
```
jcooklin: In this case I feel snapd and the rest package should use their own errors.

snapd.go
```diff
@@ -663,6 +699,7 @@ func applyCmdLineFlags(cfg *Config, ctx *cli.Context) {
 	// next for the RESTful server related flags
 	cfg.RestAPI.Enable = setBoolVal(cfg.RestAPI.Enable, ctx, "disable-api", invertBoolean)
 	cfg.RestAPI.Port = setIntVal(cfg.RestAPI.Port, ctx, "api-port")
+	cfg.RestAPI.Address = setStringVal(cfg.RestAPI.Address, ctx, "ip-addr")
```
jcooklin: This should be "api-add" not "ip-addr"

snapd.go
```diff
@@ -645,9 +653,37 @@ func setDurationVal(field time.Duration, ctx *cli.Context, flagName string) time
...
+	}
+	return nil
+}
+
 // Apply the command line flags set (if any) to override the values
 // in the input configuration
 func applyCmdLineFlags(cfg *Config, ctx *cli.Context) {
+	err := checkCmdLineFlags(ctx)
+	if err != nil {
+		panic(err)
```
jcooklin: Let's remove the panic and call log.Fatal.  Calling `log.Fatal` calls `os.Exit(1)` after printing the log message.

mgmt/rest/server.go
jcooklin: The error here isn't used anywhere.  Remove it entirely since the func you added in snapd `checkCmdLineFlags` already checks for this?

snapd.go
obourdon: Good catch thanks

snapd.go
obourdon: Sure will. This answer one of my questions thanks

core/serror/serror.go
obourdon: OK just removing this part then.

snapd.go
obourdon: OK

mgmt/rest/server.go
obourdon: Understood and agreed. However I will still run the test and call log.Fatal because we might enter this code when running tests like PluginRestCalls for instance which will not go through

Considering this, I would therefore like to "factorize/refactor/modify" the checkCmdLineFlags and extract from it the address validation mechanisms so that this can be common to both snapd.go and server.go. However, where should I put such a code in the current code hierarchy ?

obourdon:
Not sure I understand why the TravisCI jobs were successful for Go 1.5 and not for Go 1.6

jcooklin:
LGTM

PR #912, opened 2016-05-11
Display tags in watch output for a task.

geauxvirtual:
Summary of changes:
- Add support for displaying tags during watch by passing a `-v` or `--verbose` to the watch command
- Displays three tags per line, and groups tags together in the `TAGS` column of the display

Testing done:
- Verified `watch` command continues to function without passing `-v` and displays tags when `-v` is passed.

Example output when `-v` is passed to `watch`

screen shot 2016-05-11 at 2 23 59 pm

@intelsdi-x/snap-maintainers


IRCody:
LGTM

Do we need to tag an issue to the commit?

PR #917, opened 2016-05-12
Fixes pool eligibility #910

### LINKED ISSUE #910: Pool eligibility incorrectly calculated for concurrency count > 1
Function ```pool.Eligible()``` does not work properly for cases where ```p.concurrencyCount > 1``` 

Main condition:

```go
should := p.SubscriptionCount() / p.concurrencyCount
if should > p.Count() && should <= p.max {
    return true
}
return false
```
Let's assume maximum pool size to 3 and only one instance of available plugin in the pool 
```go
p.max = 3
p.Count() = 1
``` 
then

|p.SubscriptionCount()|p.concurrencyCount|should|Eligible|Expected|
| :------: |:------:| :------:| :------: | :------: |
| 1 | 2 | 0 |False | False |
| 2 | 2 | 1 |False | False |
| 3 | 2 | 1 |False | True |
| 4 | 2 | 2 |False | True |
| 5 | 2 | 2 |False | True |
| 6 | 2 | 3 |True | True |

Suggested solution:

```go
should := p.SubscriptionCount() > p.concurrencyCount * p.Count()
oversized := p.SubscriptionCount() > p.concurrencyCount * p.max
if should  && !oversized {
    return true
}
return false
```

marcin-krolik:
Fixes #910 

Summary of changes:
- changed condition for eligibility

Testing done:
- unit tests

@intelsdi-x/snap-maintainers

Pool eligibility was not calculated properly for concurrency count > 1

### DIFF
--- a/control/strategy/pool.go
+++ b/control/strategy/pool.go
@@ -270,8 +270,9 @@ func (p *pool) Eligible() bool {
 		return false
 	}
 
-	should := p.SubscriptionCount() / p.concurrencyCount
-	if should > p.Count() && should <= p.max {
+	should := p.SubscriptionCount() > p.concurrencyCount*p.Count()
+	// Check if pool is eligible and number of plugins is less than maximum allowed
+	if should && p.Count() < p.max {
 		return true
 	}
 
--- a/control/strategy/pool.go
+++ b/control/strategy/pool.go
@@ -266,13 +266,12 @@ func (p *pool) Eligible() bool {
 
 	// optimization: don't even bother with concurrency
 	// count if we have already reached pool max
-	if p.Count() == p.max {
+	if p.Count() >= p.max {
 		return false
 	}
 
-	should := p.SubscriptionCount() > p.concurrencyCount*p.Count()
 	// Check if pool is eligible and number of plugins is less than maximum allowed
-	if should && p.Count() < p.max {
+	if p.SubscriptionCount() > p.concurrencyCount*p.Count() {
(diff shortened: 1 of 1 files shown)


control/strategy/pool.go
```diff
@@ -270,8 +270,9 @@ func (p *pool) Eligible() bool {
 		return false
 	}
 
-	should := p.SubscriptionCount() / p.concurrencyCount
-	if should > p.Count() && should <= p.max {
+	should := p.SubscriptionCount() > p.concurrencyCount*p.Count()
+	// Check if pool is eligible and number of plugins is less than maximum allowed
+	if should && p.Count() < p.max {
```
IRCody: I'm wondering if we should change the code at 269 to be:
```go
if p.Count() >= p.max {
        return false
}
``` 
This would capture the case where plugin count went over the max and ensure that when we get to the next statement p.Count() < p.max. Then we could modify this if-statement to check only:
```go
if p.SubscriptionCount() > p.ConcurrencyCount*p.Count() {
       return true
}
```

marcin-krolik: I was thinking about the same. Code looked to verbose, but I decided to leave it that way. But you are right.

IRCody:
LGTM

candysmurf:
LGTM too. But it's better to check with @jcooklin & @geauxvirtual as I had doubt on the calculation before. I'm not sure any reason it did like that.

PR #928, opened 2016-05-17
Fixed example psutil and mock task files

tiffanyfj:
Summary of changes:
- Removed config from psutil example since psutil doesn't have a config policy. Also it was for mock metrics
- Updated "user" to "name" to match mock plugin (however, I think it would make more sense to update mock--thoughts?)

@intelsdi-x/snap-maintainers

### DIFF
--- a/examples/tasks/mock-file.json
+++ b/examples/tasks/mock-file.json
@@ -13,7 +13,7 @@
             },
             "config": {
                 "/intel/mock": {
-                    "user": "root",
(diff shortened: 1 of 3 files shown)


danielscottt:
LGTM

marcin-krolik:
LGTM

geauxvirtual:
This will need to be rebased for tests to be run with the merge of PR #904.

PR #932, opened 2016-05-18
Use govendor to handle dependency

nanliu:


nanliu:
Do not merge, this is used to contrast different ways to handle vendoring. Such as #933.

elemoine:
See https://github.com/intelsdi-x/snap/compare/master...elemoine:vendor for an alternative based on Git submodules in the `vendor` directory.

PR #933, opened 2016-05-18
Vendor godep dependencies

nanliu:


nanliu:
Do not merge, this is used to contrast different ways to handle vendoring. Such as #932.

elemoine:
I am curious to know what the conclusions of this experiment were. Do you think using godep vendor could be a good to manage dependencies for Snap? Thanks for any response.

IRCody:
I'm not sure there was a firm conclusion. There were pros and cons to all the vendoring options and I do not believe a decision was made on which to move forward with.

cc: @nanliu

nanliu:
@elemoine, we haven't made a decision about vendoring, we are going to try other go dependency tools and the final vendoring decision depends on which tool we pick.

elemoine:
To me the first decision to make is whether you want to a) have a copy of the deps in `vendor`, or b) use submodules, or c) use a third-party tool for fetching deps. My choice would be b). This is why I've tested with Vandetta.

PR #938, opened 2016-05-19
Use glide for vendor management.

nanliu:


nanliu:
Do not merge, this is used to contrast different ways to handle vendoring. Such as #932.

PR #941, opened 2016-05-20
Tagging metrics via task manifest

### LINKED ISSUE #827: Tag all metrics collected in a workflow via config
As a user I would like to be able to apply a configuration to a task or a global configuration that will apply a tag to any metric that is collected.  This would would allow additional context to be communicated through to the metrics final destination(s) providing insights like what experiment, user, group, etc this task was associated to.

marcin-krolik:
Fixes #827

Summary of changes:
- Adds tags in ```wmap```, structured in a similar way as configuration
- Enhanced addition of standard tags to also reapply workflow tags to metrics after collection

Testing done:
- unit tests
- new small tests created for ```wmap``` and ```metrics```

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -897,7 +897,19 @@ func (p *pluginControl) MetricExists(mns core.Namespace, ver int) bool {
 // CollectMetrics is a blocking call to collector plugins returning a collection
 // of metrics and errors.  If an error is encountered no metrics will be
 // returned.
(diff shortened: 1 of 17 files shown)


danielscottt:
LGTM!

candysmurf:
👍

PR #943, opened 2016-05-23
Fixes #942: Return schedules in REST API response for task creation for windowed and cron schedules

### LINKED ISSUE #942: REST Response for task creation does not properly return schedules for windowed or cron types
When creating a task through the REST API, the response sent back on will only include the schedule if the task that was created used a simple schedule. If the task was created with a windowed or cron schedule type, the schedule returned back from the API will be empty.

geauxvirtual:
Fixes #942 

Summary of changes:
- Adds returning windowed and cron schedule types on response from REST API on task creation
- Adds Entry() method to cron schedule type to return the entry that was used to create the schedule.

Testing done:
- Verified tests passed.
- Verified task creation response from REST API included the schedule for windowed and cron types.

@intelsdi-x/snap-maintainers

### DIFF
--- a/mgmt/rest/rbody/task.go
+++ b/mgmt/rest/rbody/task.go
@@ -222,8 +222,23 @@ func assertSchedule(s schedule.Schedule, t *AddScheduledTask) {
 			Interval: v.Interval.String(),
 		}
 		return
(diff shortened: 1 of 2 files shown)


danielscottt:
LGTM

PR #945, opened 2016-05-23
sdi-1177 make consistent behavior for RecordFailure

### LINKED ISSUE #858: RecordFailure in scheduler/task.go used inconsistently between job types
Usage of [RecordFailure()](https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L359) is inconsistent between [collect](https://github.com/intelsdi-x/snap/blob/master/scheduler/workflow.go#L319) and [process](https://github.com/intelsdi-x/snap/blob/master/scheduler/workflow.go#L407)/[publish](https://github.com/intelsdi-x/snap/blob/master/scheduler/workflow.go#L449). 

`errors := t.manager.Work(j).Promise.Await()`

For process and publish we are sending errors to RecordFailure(), for collect we are sending j.Errors(). Is this expected/correct? It's not clear to me which one should be sent to RecordFailure() or if some other behavior is desired.

There is also a related issue where RecordFailure will panic if len(e) == 0 since it doesn't check before indexing directly into e. Because on collect it is called with j.Errors(), without first checking j.Errors length (but len(errors) is checked), it is possible for the length to be 0.

candysmurf:
Fixes #858

Summary of changes:
- Send errors to RecordFailure if a job error exists
- Send promise errors to RecordFailure if errors exist

Testing done:
- unit mock test

@intelsdi-x/snap-maintainers

### DIFF
--- a/scheduler/workflow.go
+++ b/scheduler/workflow.go
@@ -318,8 +318,8 @@ func (s *schedulerWorkflow) Start(t *task) {
 	// Block until the job has been either run or skipped.
 	errors := t.manager.Work(j).Promise().Await()
 
-	if len(errors) != 0 {
-		t.RecordFailure(j.Errors())
+	if len(errors) > 0 {
+		t.RecordFailure(errors)
 		event := new(scheduler_event.MetricCollectionFailedEvent)
 		event.TaskID = t.id
 		event.Errors = errors


scheduler/workflow.go
```diff
@@ -310,17 +310,17 @@ func (s *schedulerWorkflow) Start(t *task) {
 	}).Info(fmt.Sprintf("Starting workflow for task (%s\\%s)", t.id, t.name))
 	s.state = WorkflowStarted
 	j := newCollectorJob(s.metrics, t.deadlineDuration, t.metricsManager, t.workflow.configTree, t.id)
+	if len(j.Errors()) > 0 {
```
danielscottt: `j` cannot have an error here, it's just being constructed.  That being said, I think `toRecordFailure` is probably not needed, we just need the changes you made on line 322 and 323 to use `errors` instead of `j.Errors()`.

candysmurf:
@danielscottt , thanks for the review. Changed, please check it again.

scheduler/workflow_test.go
```diff
@@ -422,6 +422,5 @@ func TestWorkJobs(t *testing.T) {
 			// (3*3)
 			So(m3.queue["publisher"], ShouldEqual, 12)
 		})
-
```
IRCody: Add this line back to minimize the diff?

IRCody:
LGTM

PR #944, opened 2016-05-23
Partial refactor of /plugin tests

mjbrender:
I walked through @tjmcs' tutorial to review content. Worth adding in. Summary of changes:

* Including a `main_small_test.go` in each subdirectory for the first time
* Leaving the more complex system testing in, tagged as `legacy` since it depends on `control` and `core` packages that @geauxvirtual is working on separately. Will leave these until we decide on whether it's a medium test or throw away for other testing options.

Testing done:
- After each file change, went to subdirectory of the change and ran: 
```
o test -tags=small -coverprofile=/tmp/coverage.out . && go tool cover -func=/tmp/coverage.out | grep -v '\t0.0%$'
```

Example: 

```
$ cd collector/snap-collector-mock1
$ go test -tags=small -coverprofile=/tmp/coverage.out . && go tool cover -func=/tmp/coverage.out | grep -v '\t0.0%$'
ok  	github.com/intelsdi-x/snap/plugin/collector/snap-collector-mock1	0.013s	coverage: 100.0% of statements
github.com/intelsdi-x/snap/plugin/collector/snap-collector-mock1/main.go:31:	main		100.0%
total:										(statements)	100.0%
```

@intelsdi-x/snap-maintainers

PR #946, opened 2016-05-24
medium tests for snap-collector-mock1

katarzyna-z:
Summary of changes:
- Added medium tests for snap-collector-mock1

@intelsdi-x/snap-maintainers

### DIFF
--- a/plugin/collector/snap-collector-mock1/mock/mock_medium_test.go
+++ b/plugin/collector/snap-collector-mock1/mock/mock_medium_test.go
@@ -1,0 +1,221 @@
+// +build medium
+
+/*
+http://www.apache.org/licenses/LICENSE-2.0.txt
+
+
+Copyright 2016 Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mock
+
+import (
+	"math/rand"
+	"testing"
+	"time"
+
+	"github.com/intelsdi-x/snap-plugin-utilities/str"
+	"github.com/intelsdi-x/snap/control/plugin"
+	"github.com/intelsdi-x/snap/core"
+	"github.com/intelsdi-x/snap/core/cdata"
+	"github.com/intelsdi-x/snap/core/ctypes"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestCollectMetric(t *testing.T) {
+	ns0 := core.NewNamespace("intel", "mock", "test")
+	ns1 := core.NewNamespace("intel", "mock", "foo")
+	ns2 := core.NewNamespace("intel", "mock", "bar")
+	ns3 := core.NewNamespace("intel", "mock").AddDynamicElement("host", "name of the host").AddStaticElement("baz")
+
+	Convey("Testing CollectMetric", t, func() {
+
+		newPlg := new(Mock)
+		So(newPlg, ShouldNotBeNil)
+
+		Convey("with 'test' config variable'", func() {
+
+			node := cdata.NewNode()
+			node.AddItem("test", ctypes.ConfigValueBool{Value: true})
+			cfg := plugin.ConfigType{ConfigDataNode: node}
+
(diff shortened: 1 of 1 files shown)


mjbrender:
FYI - this will conflict with #944 and can be resolved pending the order of merges @intelsdi-x/snap-maintainers

mjbrender:
Hey @katarzyna-z - this looks great. Code overage is hitting the mark: 

```
$ go test -tags=small -coverprofile=/tmp/coverage.out . && go tool cover -func=/tmp/coverage.out | grep -v '\t0.0%$'
ok  	github.com/intelsdi-x/snap/plugin/collector/snap-collector-mock1	0.013s	coverage: 100.0% of statements
github.com/intelsdi-x/snap/plugin/collector/snap-collector-mock1/main.go:31:	main		100.0%
total:										(statements)	100.0%
$ cd mock/
$ go test -tags=small -coverprofile=/tmp/coverage.out . && go tool cover -func=/tmp/coverage.out | grep -v '\t0.0%$'
ok  	github.com/intelsdi-x/snap/plugin/collector/snap-collector-mock1/mock	0.015s	coverage: 100.0% of statements
github.com/intelsdi-x/snap/plugin/collector/snap-collector-mock1/mock/mock.go:50:	CollectMetrics	100.0%
github.com/intelsdi-x/snap/plugin/collector/snap-collector-mock1/mock/mock.go:82:	GetMetricTypes	100.0%
github.com/intelsdi-x/snap/plugin/collector/snap-collector-mock1/mock/mock.go:99:	GetConfigPolicy	100.0%
github.com/intelsdi-x/snap/plugin/collector/snap-collector-mock1/mock/mock.go:111:	Meta		100.0%
github.com/intelsdi-x/snap/plugin/collector/snap-collector-mock1/mock/mock.go:125:	randInt		100.0%
total:											(statements)	100.0%
```

The only catch for our definition of small is that tests cover other packages. Like those in `TestGetMetricTypes`, also cover methods in `node` and `core`. I believe the goal of small tests would be to use nothing from other packages while you test a given method. Did I get that right @tjmcs?

plugin/collector/snap-collector-mock1/mock/mock_small_test.go
```diff
@@ -0,0 +1,220 @@
...
+import (
+	"math/rand"
+	"testing"
+	"time"
+
+	"github.com/intelsdi-x/snap-plugin-utilities/str"
+	"github.com/intelsdi-x/snap/control/plugin"
+	"github.com/intelsdi-x/snap/core"
+	"github.com/intelsdi-x/snap/core/cdata"
+	"github.com/intelsdi-x/snap/core/ctypes"
```
tjmcs: The fact that these other Snap packages are being imported hints at this not being a true small test (since it relies on functionality from these other packages); ideally the calls to the functions or methods in these other packages should be mocked if this really is a set of small tests.

plugin/collector/snap-collector-mock1/mock/mock_small_test.go
```diff
@@ -0,0 +1,220 @@
...
+			So(err, ShouldBeNil)
+			So(len(mts), ShouldEqual, 4)
+
+			Convey("checking namespaces", func() {
+				metricNames := []string{}
+				for _, m := range mts {
+					metricNames = append(metricNames, m.Namespace().String())
+				}
+
+				ns := core.NewNamespace("intel", "mock", "test")
```
tjmcs: Here's an example of calling into something that should probably be mocked up somehow if this truly is a small test. The fact that there are calls being made to functionality in the `core` package (among others) says that this is probably a `medium` rather than a `small` test.

plugin/collector/snap-collector-mock1/mock/mock_small_test.go
marcin-krolik: I don't agree with this approach. What is the point of mocking ```github.com/intelsdi-x/snap/core/ctypes```, when I just need ```ctypes.ConfigValue``` for my test?

plugin/collector/snap-collector-mock1/mock/mock_small_test.go
marcin-krolik: So this will lead, that most of our unit test will need to be ```medium```, because amount of mocking  gymnastics which we will be forced to do to achieve the goal of being tagged as ```small``` will be just to big to accept. I think we should rethink that....

geauxvirtual:
Without updating Travis and the test.sh file to run small tests and medium tests instead of unit and integration, this test won't run in Travis.

katarzyna-z:
I renamed the tests to "medium".

@intelsdi-x/snap-maintainers

tjmcs:
LGTM, @katarzyna-z

katarzyna-z:
@intelsdi-x/snap-maintainers Can this pull request be merged?

danielscottt:
LGTM too.  Merging.

PR #947, opened 2016-05-24
Remove links from committed plugins section

candysmurf:
Summary of changes:
- Removes links from committed plugins section in plugin catalog

Testing done:
- HTML page testing.

@intelsdi-x/snap-maintainers


geauxvirtual:
We should only add plugins to the plugin catalog after the repo has been open sourced.

candysmurf:
Should we remove the whole committed plugin section then?

danielscottt:
LGTM

PR #948, opened 2016-05-24
Fixes error in json unmarshaling of cpolicy node

danielscottt:

### DIFF
--- a/control/plugin/cpolicy/node.go
+++ b/control/plugin/cpolicy/node.go
@@ -78,13 +78,9 @@ func (c *ConfigPolicyNode) UnmarshalJSON(data []byte) error {
 	if err := decoder.Decode(&m); err != nil {
 		return err
 	}
-	if n, ok := m["PolicyNode"]; ok {
-		if pn, ok := n.(map[string]interface{}); ok {
-			if rs, ok := pn["rules"]; ok {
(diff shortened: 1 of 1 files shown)


control/plugin/cpolicy/node.go
```diff
@@ -78,13 +78,9 @@ func (c *ConfigPolicyNode) UnmarshalJSON(data []byte) error {
...
 		return err
 	}
-	if n, ok := m["PolicyNode"]; ok {
-		if pn, ok := n.(map[string]interface{}); ok {
-			if rs, ok := pn["rules"]; ok {
-				if rules, ok := rs.(map[string]interface{}); ok {
-					addRulesToConfigPolicyNode(rules, c)
-				}
-			}
+	if rs, ok := m["rules"]; ok {
```
candysmurf: Do you think it's better to put "rules" as a global constant so that all code references the same instance? If any change in one place, all is changed.

danielscottt: This is actually the only place it's referenced.  I don't think you can interpolate into struct tags.

candysmurf: I see. I was wondering that too.  LGTM

IRCody:
LGTM

PR #956, opened 2016-05-27
Fix for issue #954

### LINKED ISSUE #954: If auto_discover_path is set to /tmp/WHATEVER/.... plugins are erased when snapd stops
To reproduce the issue:
go get github.com/intelsdi-x/snap

Please note that currently this should fail due to latest commits in https://github.com/appc/spec :-(
but once you fixed those, run:

cd $GOPATH/src/github.com/intelsdi-x/snap
make all
mkdir -p /tmp/X/Y/Z
rsync -a build/ /tmp/X/Y/Z
/tmp/X/Y/Z/bin/snapd -t 0 -l 1 -a /tmp/X/Y/Z/plugin
...
^C
`DEBU[0004] stopped                                       _block=start-plugin _module=control-runner
DEBU[0004] Removing plugin                               _module=control-plugin-mgr plugin-name=mock plugin-path=/tmp/X/Y/Z/plugin/snap-collector-mock2 plugin-type=collector plugin-version=2
DEBU[0004] Removing plugin                               _module=control-plugin-mgr plugin-name=passthru plugin-path=/tmp/X/Y/Z/plugin/snap-processor-passthru plugin-type=processor plugin-version=1
DEBU[0004] Removing plugin                               _module=control-plugin-mgr plugin-name=file plugin-path=/tmp/X/Y/Z/plugin/snap-publisher-file plugin-type=publisher plugin-version=3
DEBU[0004] Removing plugin                               _module=control-plugin-mgr plugin-name=mock plugin-path=/tmp/X/Y/Z/plugin/snap-collector-mock1 plugin-type=collector plugin-version=1
`
ls /tmp/X/Y/Z
bin

Only plugins loaded via snapctl should be cleaned up not the one from auto discovery path

obourdon:
Fixes #954 

Summary of changes:
- stronger checking done to decide when to clean up temp space or not

Testing done:
- tests with and without auto_discover_path (-a) command lines option
- /tmp/.... based paths
- other (~user) based path
- all tests passed successfully

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -343,6 +343,7 @@ func (p *pluginControl) returnPluginDetails(rp *core.RequestedPlugin) (*pluginDe
 	details.Path = rp.Path()
 	details.CheckSum = rp.CheckSum()
 	details.Signature = rp.Signature()
+	details.IsAutoLoaded = rp.AutoLoaded()
 
 	if filepath.Ext(rp.Path()) == ".aci" {
 		f, err := os.Open(rp.Path())
--- a/control/plugin_manager.go
+++ b/control/plugin_manager.go
@@ -156,14 +156,15 @@ func (l *loadedPlugins) findLatest(typeName, name string) (*loadedPlugin, error)
 
 // the struct representing a plugin that is loaded into snap
 type pluginDetails struct {
-	CheckSum  [sha256.Size]byte
-	Exec      string
-	ExecPath  string
-	IsPackage bool
-	Manifest  *schema.ImageManifest
-	Path      string
-	Signed    bool
-	Signature []byte
+	CheckSum     [sha256.Size]byte
+	Exec         string
+	ExecPath     string
+	IsPackage    bool
+	IsAutoLoaded bool
+	Manifest     *schema.ImageManifest
+	Path         string
+	Signed       bool
+	Signature    []byte
 }
 
 type loadedPlugin struct {
@@ -484,8 +485,11 @@ func (p *pluginManager) UnloadPlugin(pl core.Plugin) (*loadedPlugin, serror.Snap
 		return nil, se
 	}
 
-	// If the plugin was loaded from os.TempDir() clean up
-	if strings.Contains(plugin.Details.Path, os.TempDir()) {
+	// If the plugin has been uploaded via REST API
+	// aka, was not auto loaded from auto_discover_path
+	// nor loaded from tests
+	// then do clean up
+	if !plugin.Details.IsAutoLoaded {
 		pmLogger.WithFields(log.Fields{
 			"plugin-type":    plugin.TypeName(),
 			"plugin-name":    plugin.Name(),
--- a/control/plugin_manager_test.go
+++ b/control/plugin_manager_test.go
@@ -82,9 +82,10 @@ func loadPlugin(p *pluginManager, path string) (*loadedPlugin, serror.SnapError)
 	var e serror.SnapError
 	var lp *loadedPlugin
 	details := &pluginDetails{
-		Path:     path,
-		ExecPath: filepath.Dir(path),
-		Exec:     filepath.Base(path),
+		Path:         path,
+		ExecPath:     filepath.Dir(path),
+		Exec:         filepath.Base(path),
+		IsAutoLoaded: true,
 	}
 	for i := 0; i < 3; i++ {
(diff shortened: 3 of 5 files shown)


control/plugin_manager.go
```diff
@@ -485,7 +485,8 @@ func (p *pluginManager) UnloadPlugin(pl core.Plugin) (*loadedPlugin, serror.Snap
 	}
 
 	// If the plugin was loaded from os.TempDir() clean up
-	if strings.Contains(plugin.Details.Path, os.TempDir()) {
+	subdir := strings.Replace(plugin.Details.ExecPath, os.TempDir()+"/", "", 1)
```
IRCody: I would like to add a more complete comment explaining what is happening here. When looking at this issue it took me a second to know why we were removing plugins loaded from TempDir. Not just the what but the why so that the next time we look at this we know that we are trying to remove plugins loaded via rest (b/c they get written to os.TempDir()) but want to avoid deleting plugins loaded via auto-discover that might also reside in the os.TempDir().

IRCody:
This works for not deleting plugins loaded via auto-discover. However, it also no longer deletes plugins loaded via rest. I put a comment on #954 that explains what is happening and lays out some possible options to fix it.

The short version is that when writing the plugins loaded from rest `ioutil.TempDir("", "")` is used which generates something like $TMPDIR/12341234. We then join this path to the plugin file name. So the method of checking if the plugin exists in a subdir doesn't work unfortunately.

obourdon:
New version uploaded

snapd.go
```diff
@@ -523,6 +524,14 @@ func action(ctx *cli.Context) {
 							}).Error(err)
 						}
 					}
+					if err != nil {
```
IRCody: For this to be an error we would want this to be:
```go
// Check if signature file not found and plugin trust is enabled
if err != nil && cfg.Control.PluginTrust > 0 {
```
Should probably also add some context to the error about it being a failure to read the signature file.

IRCody:
Looking at the build failures I'm not sure they are related to this PR (or at least I don't see the connection) but I am working to identify the cause.

snapd.go
obourdon: You are most probably right however the code your are referring to is not part of my pull request changes ;-)

obourdon:
I think the issue comes from the changed code because running the following fails on my system:

`make all ; 
export SNAP_PATH=$PWD/build ;
go test -v  -tags=legacy ./control -run SwapPlugin`

and once done, the build/plugin directory has disappeared meaning that the clean up code in plugin_manager.go has been called with plugin.Details.IsAutoLoaded set to false :-(

IRCody:
You're right. What is happening is in the test plugins are being loaded directly (not via rest) but since they are not auto-loaded they are being deleted when that instance of control is stopped, which removes them from being available for the next test because the default is to remove them if not marked as autoLoaded. Which causes a lot of failures.

The way to fix this I think, would be to change the default behavior of a loaded plugin from being deleted to not being deleted (maybe change isAutoLoaded to RemoveOnExit?) and then mark plugins coming from rest (somewhere around [here](https://github.com/intelsdi-x/snap/blob/master/mgmt/rest/plugin.go#L151)) with RemoveOnExit = true. This would cause the desired behavior (autoloaded plugins not being deleted) while not requiring changes to all the tests.

obourdon:
I have uploaded a new version earlier today which fixes the remaining issue(s) and tests
Could you please review this new code
Thanks again for all your help

IRCody:
Yea. Sorry I missed that you had uploaded a fix already. I'll take a look at it.

control/control.go
```diff
@@ -271,6 +271,7 @@ func (p *pluginControl) Load(rp *core.RequestedPlugin) (core.CatalogedPlugin, se
 	if serr != nil {
 		return nil, serr
 	}
+	details.IsAutoLoaded = rp.AutoLoaded()
```
IRCody: Is this assignment necessary? I think details.IsAutoLoaded is already set correctly in returnPluginDetails() call on L347.

IRCody:
LGTM other than the comment above.

control/control.go
obourdon: Very good catch, new version available with this line removed

obourdon:
Many thanks for integrating this change

PR #959, opened 2016-05-30
Fixes #799: Exposes collector plugin defaults to the framework

### LINKED ISSUE #799: When a plugin exposes config with defaults those defaults need to be available to framework
For instance, the call to GetMetricTypes is called with config from the pluginConfig which does not include the defaults that were exposed by the plugin.

tiffanyfj:
Fixes #799 

Summary of changes:
- Switched ctypes.ConfigValueX to pass by value instead of reference
- Updated mock1 so the required password doesn't fail all tests since I moved the rules to the root
- Added defaults to GetMetricTypes call and the plugin config

Testing done:
- Ran tests
- Added testing
- Removed check from Apache plugin and it loaded correctly
- Loaded plugins with print statements
     - For mock1 the config passed into GetMetricTypes is cfg : map[name:{bob}]
     - For CollectMetrics name is only being set in the config policy and for "test" it gives "The mock collected data! config data: name={bob} password={testval} test=true"

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/control_test.go
+++ b/control/control_test.go
@@ -788,6 +788,32 @@ func TestMetricConfig(t *testing.T) {
 		time.Sleep(100 * time.Millisecond)
 	})
 
+	Convey("config provided by defaults", t, func() {
+		c := New(GetDefaultConfig())
+		c.Config.Plugins.All.AddItem("password", ctypes.ConfigValueStr{Value: "pwd"})
+		c.Start()
+		lpe := newListenToPluginEvent()
+		c.eventManager.RegisterHandler("Control.PluginLoaded", lpe)
+		_, err := load(c, fixtures.JSONRPCPluginPath)
+		So(err, ShouldBeNil)
+		<-lpe.done
+		cd := cdata.NewNode()
+		m1 := fixtures.MockMetricType{
+			Namespace_: core.NewNamespace("intel", "mock", "foo"),
+		}
+
+		Convey("So metric should be valid with config", func() {
+			errs := c.validateMetricTypeSubscription(m1, cd)
+			So(errs, ShouldBeNil)
+		})
+		Convey("So mock should have name: bob config from defaults", func() {
+			So(c.Config.Plugins.pluginCache["0:mock:1"].Table()["name"], ShouldResemble, ctypes.ConfigValueStr{Value: "bob"})
+		})
+
+		c.Stop()
+		time.Sleep(100 * time.Millisecond)
+	})
+
 	Convey("nil config provided by task", t, func() {
 		config := GetDefaultConfig()
 		config.Plugins.All.AddItem("password", ctypes.ConfigValueStr{Value: "testval"})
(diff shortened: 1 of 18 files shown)


control/plugin_manager.go
```diff
@@ -355,10 +355,45 @@ func (p *pluginManager) LoadPlugin(details *pluginDetails, emitter gomit.Emitter
 	lPlugin.ConfigPolicy = cp
 
 	if resp.Type == plugin.CollectorPluginType {
+		cfgNode := p.pluginConfig.getPluginConfigDataNode(core.PluginType(resp.Type), resp.Meta.Name, resp.Meta.Version)
+
+		if lPlugin.ConfigPolicy != nil {
+			// Get config policy from root/plugin defaults
+			ncd := lPlugin.ConfigPolicy.Get([]string{""})
```
IRCody: What does ncd stand for?

control/control_test.go
```diff
@@ -838,6 +838,32 @@ func TestMetricConfig(t *testing.T) {
...
+			errs := c.validateMetricTypeSubscription(m1, cd)
+			So(errs, ShouldBeNil)
+		})
+		c.Config.Plugins.All.AddItem("name", ctypes.ConfigValueStr{Value: "jane"})
+		Convey("So mock should have name: bob config from defaults", func() {
+			So(c.Config.Plugins.pluginCache["0:mock:1"].Table()["name"], ShouldResemble, ctypes.ConfigValueStr{Value: "bob"})
+		})
+
+		c.Stop()
+		time.Sleep(100 * time.Millisecond)
```
IRCody: Why do we need to sleep ?

tiffanyfj: I copied it from above. So...not sure?

control/plugin_manager.go
tiffanyfj: Also no clue...got it from control.go

tiffanyfj: Node config data? Throwing things out there... @geauxvirtual or @danielscottt would know according to git blame.

IRCody: That's probably it.

IRCody:
LGTM

control/plugin_manager.go
lynxbat: Yup, this was me.

control/plugin_manager.go
```diff
@@ -355,10 +355,45 @@ func (p *pluginManager) LoadPlugin(details *pluginDetails, emitter gomit.Emitter
...
+					fields[idx] = e.Error()
+				}
+				fields["name"] = resp.Meta.Name
+				fields["version"] = resp.Meta.Version
+				se.SetFields(fields)
+				return nil, se
+			}
+
+			// Update config policy with defaults
+			cp, err = c.GetConfigPolicy()
```
jcooklin: Is this block needed given https://github.com/tiffanyfj/snap/blob/00a75a229df3f72ef5a4f5c41865a55699481945/control/plugin_manager.go#L343-L355

control/control_test.go
danielscottt: Can we find a way to test this without time?

control/plugin_manager.go
tiffanyfj: Hmm so it mentions collector in there...but I thought it is used for all types.

tiffanyfj:
Added part of @jcooklin's commit for a future grpc PR here.

control/control_test.go
tiffanyfj: Have a suggestion besides just removing it from the code and seeing how Travis etc like it?

tiffanyfj: Seems they were added for Travis issues and I just copied it from another line...

https://github.com/intelsdi-x/snap/commit/da78c23ebea8a5b183e8086e8f8e8171268dfb16
@geauxvirtual

geauxvirtual:
LGTM.

PR #961, opened 2016-05-31
Move plugin auto discovery into control module

obourdon:
Summary of changes:
- moved code from snapd.go into control/control.go as suggested in [https://github.com/intelsdi-x/snap/issues/950]

Testing done (using GO 1.6.2):
- successfully launched with auto-discovery-path set
- all autoloaded standard plugins (mock, file, passthrough) functional
- same tests without auto-discovery-path being set and same plugins loaded via CLI
- all tests passed in legacy, small, medium and large configs

@intelsdi-x/snap-maintainers

### DIFF
--- a/mgmt/rest/rest_func_test.go
+++ b/mgmt/rest/rest_func_test.go
@@ -56,10 +56,11 @@ var (
 	// Switching this turns on logging for all the REST API calls
 	LOG_LEVEL = log.WarnLevel
 
-	SNAP_PATH         = os.Getenv("SNAP_PATH")
-	MOCK_PLUGIN_PATH1 = SNAP_PATH + "/plugin/snap-collector-mock1"
-	MOCK_PLUGIN_PATH2 = SNAP_PATH + "/plugin/snap-collector-mock2"
-	FILE_PLUGIN_PATH  = SNAP_PATH + "/plugin/snap-publisher-file"
+	SNAP_PATH              = os.Getenv("SNAP_PATH")
+	SNAP_AUTODISCOVER_PATH = os.Getenv("SNAP_AUTODISCOVER_PATH")
+	MOCK_PLUGIN_PATH1      = SNAP_PATH + "/plugin/snap-collector-mock1"
+	MOCK_PLUGIN_PATH2      = SNAP_PATH + "/plugin/snap-collector-mock2"
+	FILE_PLUGIN_PATH       = SNAP_PATH + "/plugin/snap-publisher-file"
 
 	CompressedUpload = true
 	TotalUploadSize  = 0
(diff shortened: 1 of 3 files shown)


mgmt/rest/rest_func_test.go
```diff
@@ -582,6 +583,12 @@ func TestPluginRestCalls(t *testing.T) {
 				cfg := getDefaultMockConfig()
 				err := cfgfile.Read("../../examples/configs/snap-config-sample.json", &cfg, MOCK_CONSTRAINTS)
 				So(err, ShouldBeNil)
+				if len(SNAP_AUTODISCOVER_PATH) == 0 {
+					if len(SNAP_PATH) != 0 {
+						SNAP_AUTODISCOVER_PATH = fmt.Sprintf("%s/plugin", SNAP_PATH)
+					}
+				}
+				cfg.Control.AutoDiscoverPath = SNAP_AUTODISCOVER_PATH
```
IRCody: This causes failures if the SNAP_AUTODISCOVER_PATH is set to something where the plugins fail to load (since failures on when autoloading plugins are fatal). 

What is the reasoning behind adding this/the SNAP_AUTODISCOVER_PATH env variable to this test?

obourdon: In fact, the tests were first failing because in the examples/configs/snap-config-sample.json file the path is set to a dummy value "auto_discover_path": "/some/directory/with/plugins", therefore I had the choice to either ignore/warn about this error (I guess 'ignore' was the previous behaviour) either make sure that tests could run properly by using SNAP_PATH (which is taken by default if SNAP_AUTODISCOVERY_PATH and as it is set automatically by the make check command this was nice to use) or if the user running test has a separate location for plugins, then he is able overwrite this to test proper loading of his plugins. This is our case as we are building some plugins ourselves to use in Docker containers therefore requiring (for now at least) support for configurable /proc file system. 

Hope this makes sense

IRCody: Thank you for explaining. I knew there was a reason I just didn't see it. I understand why the failures were happening now though. The previous behavior was that auto-loading never happened here because this calls control directly and autoloading only happened in snapd. 

I'm not sure I understand the rest about using it to test plugins loading from another location. What is the rationale behind that?

obourdon: In fact, we are building plugins with some modifications (basically to support proper running in Docker containers) which I forked in my personal github repository and we'd like to ensure that whenever we modify snap these plugins still load therefore the need of this variable
Is this a bit clearer ?

Do the other parts of this PR sounds OK to you ?

Thanks for everything

IRCody: The other parts LGTM. I tested it locally also and didn't notice any issues.

My concern is that someone would unintentionally have this environment variable set and that it might cause tough to decipher build errors. I think it could be ok as-is with a log.Warn() if the environment variable is set letting the user know that this could cause test failures.

obourdon: Ok, got your point and totally agree with it. New version of PR submitted
Thanks for testing all this

IRCody:
LGTM once the tests pass.

PR #972, opened 2016-06-06
Fixes #900: Removes '0' from the list of valid ports

### LINKED ISSUE #900: No output for which port was selected dynamically
When specifying a rest api port of '0', a dynamic port is selected but the port is not output to the user.

```
INFO[0000] Starting REST API on :0                       _module=_mgmt-rest
```

tjmcs:
Fixes #900

Summary of changes:
- Removes '0' from the list of valid ports that the RESTful API can be started on (as such, since a non-zero port must be specified or the default port used, the RESTful API will not have a port dynamically assigned when `snapd` starts up so there is no need to display the port that was chosen in the log file)
- Refactored how the command-line flags are set so that the command-line flags for each package are set locally in a `flags.go` file for that package rather than having the command-line flags for the options from some packages set in a package-level `flags.go` file and while the command-line flags for the options for other packages are set in the `snapd.go` file
- Modified the code that sets the command-line flags so that those that are displaying an example value display the default value for that option in the `--help` output for `snapd` (rather than displaying values like '0' for command-line options like the `--api-port`)

Testing done:
- Tested by attempting to start `snapd` with the `port` set to zero in the global configuration file and observing the error that results, then by starting it with a non-zero port in that same configuration file and observing the successful startup. Also ensured that the `--help` output for `snapd` could be generated successfully with the refactored command-line option code in this PR.

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/flags.go
+++ b/control/flags.go
@@ -1,0 +1,56 @@
+/*
+http://www.apache.org/licenses/LICENSE-2.0.txt
+
+
+Copyright 2015 Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


jcooklin:
LGTM

Issue #965, opened 2016-06-01
Add grpc support


jcooklin:
Adding support for [grpc](http://www.grpc.io) will provide benefits including
- performant client libs in many languages
- streaming RPC
- deadlines and cancellations
- auth (built-in and custom)

jcooklin:
The preferred RPC mechanism between Snap and it's plugins is now gRPC. 

The plugin libs should be used to create plugins.  The plugin libs we have so far include:
- https://github.com/intelsdi-x/snap-plugin-lib-go
- https://github.com/intelsdi-x/snap-plugin-lib-cpp
- https://github.com/intelsdi-x/snap-plugin-lib-py

(Closed by https://github.com/intelsdi-x/snap/pull/1012 )

PR #973, opened 2016-06-07
Enables grpc/common to be reused in the plugin package

jcooklin:
These changes are needed so grpc/common can be reused in #965.


control/fixtures/fixtures.go
```diff
@@ -0,0 +1,127 @@
+/*
```
IRCody: If this is only used in tests, should we add tags to it so that it's only built when tests are being built?

PR #974, opened 2016-06-07
Only use version for specific tags.

nanliu:
Since we are interating on snap, we don't want users to think
0.14.0-sha is semantic version of 0.14.0. This ensure the build script
only label version for specific git tags and use branch-sha to
indicate development versions.


jcooklin:
LGTM

PR #975, opened 2016-06-08
Fixes #971: Rechecks config option constraints after CLI flags applied

### LINKED ISSUE #971: Constraints on CLI option values are not applied on snapd startup
The constraints defined in the package-level configuration files are only used to validate the values retrieved from a global configuration file. The same constraints are not applied to values passed in on the command-line using the command-line flags defined for `snapd` or values passed in as environment variables (and these values override the values from the global configuration file).

tjmcs:
Fixes #971

Summary of changes:
- Modified the `pkg/cfgfile/cfgfile.go` public interface by renaming the `ValidateSchema` method in the underlying `schemaValidator` interface to `validateSchema`, then adding a new `ValidateSchema` function that calls through to that underlying `validateSchema` method; this provides an externally visible function that can be used to check the configuration (after applying the command-line flags) against the constraints specified in the schema from within the `action()` method in the `snapd.go` file
- Modified the `small` tests that previously tested the `ValidateSchema` method from the `schemaValidator` interface so that they test the new `ValidateSchema` function instead
- Removed the `omitempty` tags from the fields in the `Config` structs defined in the `control/config.go`, `mgmt/rest/server.go`, `mgmt/tribe/config.go`, `scheduler/config.go`, and `snapd.go` files; with the `omitempty` tags included files with `zero` values were stripped out of the configuration when we marshalled the resulting configuration as JSON in order to check the configuration against the constraints in the schema (resulting in an incomplete check of the configuration since only `non-zero` fields were checked)
- Added code to check the configuration (after the command-line options had been applied) using the new `ValidateSchema` function and report any errors found; if errors are found, then `snapd` exits with a fatal error

Testing done:
- Verified that the existing small tests from the `pkg/cfgfile` package ran without error after the refactoring was complete
- Started `snapd` using a global configuration file that set the RESTful API port to `0` and verified that an error was reported (the port for the RESTful API must be greater than or equal to 1)
- Started `snapd` using a global configuration file that set the RESTful API port to `8282` and verified that the application started correctly and that the RESTful API was listening on port `8282`
- Started `snapd` with the same global configuration file used in the previous test (with the port set to `8282`) but overrode that port assignment with a CLI option of `--api-port 0`; verified that an error was reported
- Started `snapd` with the same global configuration file used in the previous test, but overrode that port assignment with a CLI option of `--api-port 8080`; verified that the application started correctly and that the RESTful API was listening on port `8080`

@intelsdi-x/snap-maintainers

### DIFF
--- a/snapd.go
+++ b/snapd.go
@@ -104,13 +104,13 @@ const (
 //         UnmarshalJSON method in this same file needs to be modified to
 //         match the field mapping that is defined here
 type Config struct {
-	LogLevel   int               `json:"log_level,omitempty"yaml:"log_level,omitempty"`
-	GoMaxProcs int               `json:"gomaxprocs,omitempty"yaml:"gomaxprocs,omitempty"`
-	LogPath    string            `json:"log_path,omitempty"yaml:"log_path,omitempty"`
-	Control    *control.Config   `json:"control,omitempty"yaml:"control,omitempty"`
-	Scheduler  *scheduler.Config `json:"scheduler,omitempty"yaml:"scheduler,omitempty"`
-	RestAPI    *rest.Config      `json:"restapi,omitempty"yaml:"restapi,omitempty"`
-	Tribe      *tribe.Config     `json:"tribe,omitempty"yaml:"tribe,omitempty"`
+	LogLevel   int               `json:"log_level"yaml:"log_level"`
+	GoMaxProcs int               `json:"gomaxprocs"yaml:"gomaxprocs"`
+	LogPath    string            `json:"log_path"yaml:"log_path"`
+	Control    *control.Config   `json:"control"yaml:"control"`
+	Scheduler  *scheduler.Config `json:"scheduler"yaml:"scheduler"`
+	RestAPI    *rest.Config      `json:"restapi"yaml:"restapi"`
+	Tribe      *tribe.Config     `json:"tribe"yaml:"tribe"`
 }
 
 const (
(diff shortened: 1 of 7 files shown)


lmroz:
LGTM
However I noticed something (but haven't checked if I'm right). In case of port I think it's possible to set it to `0` using `--api-addr localhost:0` and also command line flags have more strict validation (if you specify port as separate flag you are not allowed to give it again using bind addr). But it's not really related to this commit.

tjmcs:
I agree completely, @lmroz; in fact @jcooklin and I were chatting yesterday about perhaps removing the ability to specify the port as part of the `--api-addr` value (or the associated `restapi:port` value in the global configuration file).

Personally, I feel that it's confusing to be able to specify the port to use for the RESTful API in both places; and in spite of the fact that there seems to be code in place to check if the port is specified using both the `--api-addr` and `--api-port` command-line arguments, it's entirely possible to define two different ports using the `restapi:port` and `restapi:addr` parameters in the global configuration file. As an aside, it's also completely unclear which will port will be used if you do define two separate ports in the global configuration file; as it turns out, the port included as part of the RESTful API address (using the `restapi:addr` parameter) overrides the a port specified using the `restapi:port` parameter in the global configuration file.

In short, I feel that catching situations where the port is specified as zero using the `address:port` form for the value of the `--api-addr` command-line argument (and throwing an appropriate error) is a separate issue that we should resolve in a subsequent pull request...thoughts from anyone else on this?

snapd.go
```diff
@@ -204,6 +204,18 @@ func action(ctx *cli.Context) {
 	// same variables in that configuration
 	applyCmdLineFlags(cfg, ctx)
 
+	// test the resulting configuration to ensure the values it contains still pass the
+	// constraints after appling the environment variables and command-line parameters;
```
jcooklin: s/appling/applying/

jcooklin:
I agree about creating an issue on the multiple ways to configure the port we will bind to.

jcooklin:
LGTM

snapd.go
tjmcs: good catch, @jcooklin; have pushed up a fix

tjmcs:
Pushed up a fix to that spelling error, @jcooklin; should be good to go now

Issue #949, opened 2016-05-25
Publishing metric extended from dynamic elemenets


marcin-krolik:
There are uses cases where parts of a metric namespace is undesirable in published metric. 

Let's say that some ```foo``` collector provides such ```/intel/foo/*/power```, where wildcard is a dynamic metric element which expands to host IP address. This way what is returned from ```CollectMetrics``` is expanded to ```/intel/foo/1.1.1.1/power```, ```/intel/foo/1.1.1.2/power``` and ```/intel/foo/1.1.1.3/power```. 
As a end user I don't want to have my timeseries DB populated with such metrics. I prefer to have simpler metric ```/intel/foo/power``` tagged with ```host_ip: 1.1.1.x```. 

Before introduction of new metric schema and ```core.Namespace```, it was possible to provide ```metric.Label_```, which in turned was used in timeseries publishers like ```snap-plugin-publisher-influxdb``` to remove unwanted namespace elements from metric and save them as tags. 

Currently it seems we don't have any mechanism which would allow similar action.

### Solution 1
Introduction of additional field in ```NamespaceElement``` to allow publisher to "see" unwanted namespace elements and turn them into tags.

```go
type NamespaceElement struct {
	Value       string
	Description string
	Name        string
	Removable   bool
}

func (n Namespace) AddRemovableElement(value string) Namespace {
	nse := NamespaceElement{Value: value, Removable: true}
	return append(n, nse)
}
```
Above function would need to be used in collectors whenever dynamic element is expanded to set of available instances.

Additionally, all timeseries publishers would need to be enhanced by a generic function which would remove all ```removable``` elements from the namespace and turn them into tags.

Pros:
- generic solution which will work every time

Cons:
- requires change in the framework
- additional effort to implement generic function in publishers
- additional effort to implement change in some collector plugins

### Solution 2
New processor plugin which would serve as a "namespace filter". It would require regex expression provided in configuration and based on the regex it would remove namespace element which matches and turn it into a tag.

Pros:
- no changes to framework
- no changes to plugins

Cons:
- playing with regex gives ability to damage published metrics (namespace changed unintentionally)

cc: @intelsdi-x/snap-maintainers , @intelsdi-x/plugin-maintainers @PatrykMatyjasek @mkuculyma

marcin-krolik:
As for the second solution, processor is ready [ns-filter](https://github.com/intelsdi-x/snap-plugin-processor-ns-filter/pull/1)

[...]

PR #977, opened 2016-06-08
SDI-1394 a metric can be changed from static to dynamic by setting na…

### LINKED ISSUE #937: A metric can be changed from static to dynamic by setting Name parameter on NamespaceElement
Currently, the only determination if a metric namespace is dynamic or static is if `Name` is set on a NamespaceElement(s). 

After creating a metric namespace via `core.NewNamespace("intel", "mock", "foo")`, there is nothing that prevents setting the Name field on the three NamespaceElements that make up this Namespace. Initially, `IsDynamic()` on this Namespace would return false. However, if a plugin were to go the extra step to set the `Name` field on each NamespaceElement, like so

```
ns[0].Name = "org"
ns[1].Name = "plugin"
ns[2].Name = "metric"
```
`IsDynamic()` will now return true on this metric namespace. Any tooling that would consume this information and display dynamic metrics different from static metrics would see an issue. As an example, this is what snapctl would display with the changes in PR #931 

screen shot 2016-05-19 at 11 24 53 am

candysmurf:
Fixes #937 

Summary of changes:
- checks the name along with the value

Testing done:
- unit test

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/metrics.go
+++ b/control/metrics.go
@@ -83,6 +83,14 @@ func errorMetricEndsWithAsterisk(ns string) error {
 	return fmt.Errorf("Metric namespace %s ends with an asterisk is not allowed", ns)
 }
 
+func errorMetricStaticElementHasName(value, name, ns string) error {
+	return fmt.Errorf("A static element %s should not define name %s for namespace %s.", value, name, ns)
+}
+
+func errorMetricDynamicElemnetHasNoName(value, ns string) error {
+	return fmt.Errorf("A dynamic element %s requires a name for namespace %s.", value, ns)
+}
+
 // listNotAllowedChars returns list of not allowed characters in metric's namespace as a string
 // which is used in construct errorMetricContainsNotAllowedChars as a recommendation
 // exemplary output: "brackets [( ) [ ] { }], spaces [ ], punctuations [. , ; ? !], slashes [| \ /], carets [^], quotations [" ` ']"
@@ -338,19 +346,28 @@ func (mc *metricCatalog) removeMatchedKey(key string) {
 
 // validateMetricNamespace validates metric namespace in terms of containing not allowed characters and ending with an asterisk
 func validateMetricNamespace(ns core.Namespace) error {
-	name := ""
+	value := ""
 	for _, i := range ns {
-		name += i.Value
+		// A dynamic element requires the name while a static element does not.
+		if i.Name != "" && i.Value != "*" {
+			return errorMetricStaticElementHasName(i.Value, i.Name, ns.String())
+		}
+		if i.Name == "" && i.Value == "*" {
+			return errorMetricDynamicElemnetHasNoName(i.Value, ns.String())
+		}
+
+		value += i.Value
 	}
+
 	for _, chars := range notAllowedChars {
 		for _, ch := range chars {
-			if strings.ContainsAny(name, ch) {
+			if strings.ContainsAny(value, ch) {
 				return errorMetricContainsNotAllowedChars(ns.String())
 			}
 		}
 	}
 	// plugin should NOT advertise metrics ending with a wildcard
-	if strings.HasSuffix(name, "*") {
+	if strings.HasSuffix(value, "*") {
 		return errorMetricEndsWithAsterisk(ns.String())
 	}
 
@@ -363,7 +380,7 @@ func (mc *metricCatalog) AddLoadedMetricType(lp *loadedPlugin, mt core.Metric) e
 			"_module": "control",
 			"_file":   "metrics.go,",
 			"_block":  "add-loaded-metric-type",
-			"error":   fmt.Errorf("Metric namespace %s contains not allowed characters", mt.Namespace()),
+			"error":   fmt.Errorf("Metric namespace %s is invalid", mt.Namespace()),
 		}).Error("error adding loaded metric type")
 		return err
 	}
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -1,0 +1,124 @@
+
+# Snap Static and Dynamic Metrics
+
+Snap Framework supports two types of metrics. They are static and dynamic metrics. A Snap metric consists of a namespace and value pair.
+
+### Static Metrics
+A namespace having no wildcard and only string literals separated by slashes is a static metric.
+
+String representation examples:
+```
+/intel/cassandra/node/zeus/type/Cache/scope/KeyCache/name/Requests/OneMinuteRate
+/intel/cassandra/node/zeus/type/ClientRequest/scope/CASRead/name/Unavailables/FiveMinuteRate
+/intel/cassandra/node/apollo/type/Cache/scope/RowCache/name/Hits/OneMinuteRate
+/intel/cassandra/node/apollo/type/ClientRequest/scope/RangeSlice/name/Latency/OneMinuteRate
+```
+### Dynamic Metrics
+A namespace including at least one wildcard is a dynamic metric.
+
+String representation examples:
+```
+/intel/cassandra/node/*/type/*/scope/*/name/*/OneMinuteRate
+/intel/cassandra/node/*/type/*/scope/*/name/*/FiveMinuteRate
+/intel/cassandra/node/*/type/*/keyspace/*/name/*/OneMinuteRate
+/intel/cassandra/node/*/type/*/keyspace/*/name/*/FiveMinuteRate
+```
+### Namespace Element
+Namespace element in Snap is defined as a struct. Both static and dynamic elements share the same definition.
+ ```
+ type NamespaceElement struct {
+	Value       string
+	Description string
+	Name        string
+}
+ ```
+The _`NamespaceElement`_ forms each cell of a namespace and those cells are separated by slashes. 
+
+Create a dynamic element:
+```
+ns := core.NewNamespace("intel", "cassandra", "node")
+    .AddDynamicElement("node name", "description")
+```
+
+Create multiple dynamic elements:
+```
+ns := core.NewNamespace("intel", "cassandra", "node")
+    .AddDynamicElement("node name", "description")
+    .AddStaticElement("type")
+    .AddDynamicElement("type value", "description")
+    .AddStaticElement("scope")
+    .AddDynamicElement("scope value", "description")
+    .AddStaticElement("name")
+    .AddDynamicElement("name value", "description")
+    .AddStaticElement("50thPercentile")
+```
+
+### Why Dynamic Element
+By defining dynamic elements inside a namespace, you'll have the capability to treat them differently at a later time. For instance, you have metric namespaces:
+```
+/foo/bar/host-alice/status
+/foo/bar/host-bob/status
+/foo/bar/host-nicole/status
+/foo/bar/host-david/status
+```
+To get the measurement `/foo/bar/status`, specifying the _`hostname`_ element as a dynamic element and create filtering friendly _`hostname`_ tags.
+
+>The key takeaway is that the _`named element`_ is a _`dynamic element`_. A static element has an _`empty Name`_ field.
+
+### Collector Plugins
+Building a Snap collector plugin involves two primary tasks. One is to create a collector metric catalog. Another is to collect metric data.
+
+##### Create Metric Catalog
+Creating a collector having dynamic metric catalog by utilizing the following methods from Snap _`core`_ package, Snap CLI(snapctl) verbose output could display the definition and the description of a wildcard along with a namespace's `Measurement Unit` if it's defined.
+
+Methods:
+```
+(n Namespace) AddStaticElement(value string) Namespace
+(n Namespace) AddDynamicElement(name, description string) Namespace
+(n Namespace) AddStaticElements(values ...string) Namespace
+```
+
+Create Metric Catalog:
+```
+metricType := plugin.MetricType{
+    Namespace_: ns,
+    Unit_:   <namespace measurement unit>,
+}
+```
+
+Snap CLI verbose output:
+```
+$ $SNAP_PATH/bin/snapctl metric list --verbose
+
+/intel/cassandra/node/[node name]/type/[type value]/scope/[scope value]/name/[name value]/OneMinuteRate  float64
+/intel/cassandra/node/[node name]/type/[type value]/scope/[scope value]/name/[name value]/FiveMinuteRate  float64
+/intel/cassandra/node/[node name]/type/[type value]/keyspace/[keyspace value]/name/[name value]/OneMinuteRate  float64
+/intel/cassandra/node/[node name]/type/[type value]/keyspace/[keyspace value]/name/[name value]/FiveMinuteRate  float64
+```
+##### Collecting metric data
+While collecting metric data, collector plugin authors may leverage Snap framework by specifying `Name` field for every _`dynamic`_ element but leaving an empty `Name` field for each static element.
+
+>If you use the _`(n Namespace) AddDynamicElement(value string) Namespace`_ method to build dynamic elements for a metric, remember setting the actual metric element values.
+
+### Publisher Plugins
+Snap publisher plugins may leverage following methods from Snap to determine if an element is dynamic or static _`if and only if`_ a collector correctly defined the _`Name`_ field for every dynamic element.
+
+Get all dynamic element positions:
+```
+isDynamic, indexes := ns.IsDynamic() 
+```
+Where `isDynamic` is a bool type which indicates if a namespace contains at least one dynamic element and `indexes` is an array of dynamic element positions.
+
+Loop through each element:
+```
+for _, elt := range ns {
+    elt.IsDynamic() {
+        // strip off dynamic element to create searchable metric tags
+    }
+}
+```
+It's even better that plugin [utilities](https://github.com/intelsdi-x/snap-plugin-utilities/blob/master/mts/mts.go#L32) may save your time.
+
+> Snap checks the Name field of a namespace element to determine if an item is static or dynamic.  The _`Name`_ field should be empty as a static element but not empty as a dynamic element.
+
+Appropriately define your dynamic element to leverage Snap framework!
--- a/control/metrics_test.go
+++ b/control/metrics_test.go
@@ -561,3 +561,30 @@ func TestMetricNamespaceValidation(t *testing.T) {
 		})
 	})
 }
+
+func TestMetricStaticDynamicNamespace(t *testing.T) {
+	Convey("validateStaticDynamic()", t, func() {
+		Convey("has static elements only", func() {
+			ns := core.NewNamespace("mock", "foo", "bar")
+			err := validateMetricNamespace(ns)
+			So(err, ShouldBeNil)
+		})
+		Convey("had both static and dynamic elements", func() {
+			ns := core.NewNamespace("mock", "foo", "*", "bar")
+			ns[2].Name = "dynamic element"
+			err := validateMetricNamespace(ns)
+			So(err, ShouldBeNil)
+		})
+		Convey("has name for a static element", func() {
+			ns := core.NewNamespace("mock", "foo")
+			ns[0].Name = "static element"
+			err := validateMetricNamespace(ns)
+			So(err, ShouldNotBeNil)
+		})
+		Convey("has * but no name", func() {
+			ns := core.NewNamespace("mock", "foo", "*", "bar")
+			err := validateMetricNamespace(ns)
+			So(err, ShouldNotBeNil)
+		})
+	})
+}
--- a/docs/PLUGIN_AUTHORING.md
+++ b/docs/PLUGIN_AUTHORING.md
@@ -87,6 +87,10 @@ Example:
 
 Snap validates the metrics exposed by plugin and, if validation failed, return an error and not load the plugin.
 
+##### c) static and dynamic metrics
+Snap supports both static and dynamic metrics. Do you like to know the differences and the recommendation for plugins?
+See details [here](./METRICS.md).
+
 ### Mandatory packages
 There are three mandatory packages that every plugin must use. Other than those three packages, you can use other packages as necessary. There is no danger of colliding dependencies as plugins are separated processes. The mandatory packages are:
 ```


lmroz:
This is bad fix.
Plugins exposing dynamic metrics change the `Value` field from `*` to it's realization:
https://github.com/intelsdi-x/snap/blob/master/plugin/collector/snap-collector-mock1/mock/mock.go#L58

Then this is used to strip such namespace element from metric name:
https://github.com/intelsdi-x/snap-plugin-publisher-influxdb/blob/master/influx/influx.go#L163

candysmurf:
@Imroz, thanks for reviewing it. From my understanding, IsDynamic only applies to the metric types, not metric data themselves. I don't see two issues you linked. Thoughts?

lmroz:
@candysmurf Dynamic metric as exposed by `GetMetricTypes` is defined as a set of metrics that may change during runtime. So every named element (formely just `*`) contains some kind of data. In mock plugin this is name of a host that we pretend to measure.

So example scenario is:

- plugin exposes namespace `/org/[host]/power_consumption`
- user requests metric `/org/*/power_consumption`
- plugin returns 2 metric types `/org/host1/power_consumption = 10W`, `/org/host2/power_consumption = 15W`

Now, publisher is going to put such rows in database:
```
| MetricName             | Host  | Value |
|------------------------|-------|-------|
| /org/power_consumption | host1 | 10W   |
| /org/power_consumption | host2 | 15W   |
```
It does this by inspecting namespace elements using `.IsDynamic()`, then striping them from metric name and putting them as separate column in DB.

But notice `*` is present in namespace only before `CollectMetrics`, so after your fix such data will be generated in DB:
```
| MetricName                   | Value |
|------------------------------|-------|
| /org/host1/power_consumption | 10W   |
| /org/host2/power_consumption | 15W   |
```

candysmurf:
@lmroz, thanks for your examples. However, I tried with Snap master where it does not have my change. It does not do what you expected.

IMHO, in the publishers, we know if it's a dynamic or static metric and the position of wildcard "*"  The wildcard may be in many places beyond your examples. We could have a scenario where stripping some of them and keeping a few of them. In another word, publishers should do diligent based on use cases. For example if you like to store "/org/power_consumption" instead of "/org/host1/power_consumption".

In summary, IsDynamic should only check for metric types that defined in the task manifest. It has no other purpose. If it does, it's wrong itself. Please share your thoughts.

marcin-krolik:
@candysmurf Emily, Lukasz is right. This fix will break handling of dynamic metrics further in the pipeline. You can't leave "*" in dynamic metric after collection. You have to change the ```Value``` to particular instance. Please take a closer look at mock plugin and how it is handled [there](https://github.com/intelsdi-x/snap/blob/master/plugin/collector/snap-collector-mock1/mock/mock.go#L54-L65)

candysmurf:
@marcin-krolik, @lmroz:  I cannot agree that the IsDynamic function is useful in collectors. For a publisher plugin, if we have to depend on IsDynamic to decide what to do for the incoming metric data, does it mean that we'll need to populate the name for each dynamic element when we're collecting the data?  Is there any better and scalable way to achieve this when we're trying to address this bug. Share your thoughts.

candysmurf:
@lmroz , @marcin-krolik, @jcooklin, @geauxvirtual , @intelsdi-x/plugin-maintainers:

I propose to remove IsDynamic function entirely. It's flawed. To use your example, when we create a namespace, it should be in the following format:

/org/host/host1/power_consumption
/org/host/host2/power_consumption

A publisher can strip off both "/host/host1" and "/host/host2" to create tags.  Another even more obvious approach could be 

/org/host=host1/power_consumption
/org/host=host2/power_consumption

or

/org/host*host1/power_consumption
/org/host*host2/power_consumption

So that you definitely know which one to remove and create tags. 
Please share your thoughts.

elemoine:
I am failing to understand what issue this PR is attempting to fix. I did read #937, but I don't understand why a plugin would want to give names to namespace elements (e.g. `ns[0].Name = "org"`). The approach stating that elements with names are dynamic elements makes sense to me. Plugins should just be aware of that.

marcin-krolik:
@candysmurf Problem with your PR is that it tries to address reported bug incorrectly. I had the same confusion when I reported #949 . 

`IsDynamic()` is useful especially in publisher or processors. Metric which leaves collector **cannot have** wildcard (`*` character) as `Value` in any of `NamespaceElement`. On the other hand **we need** to have some sort of mechanism to recognize metrics which have "dynamic" origin (metric with wildcard was send to `CollectMetrics()`). This is achieved by substituting `NamespaceElement.Value` with particular "instance" like `host1`, or `10.1.1.2` etc., but leaving `NamespaceElement.Name` as it was. This way we can still identify metric as dynamic, because `IsDynamic()` checks for non-empy value of `Name`. 

In short, you were asked to collect such metric
`/intel/*/foo` 

This metric was created with namespace as 
```go
metric.Namepace_ = core.NewNamespace("intel").AddDynamicElement("host_ip", "IP address of the host").AddStaticElement("foo")
```

In `CollectMetric()` you need to handle this metric. You have two ways of doing that

1. with static namespace
2. with dynamic namespace

If you will choose 1 like:
```go
metric.Namespace_: core.NewNamespace("intel", "10.1.1.2", "foo").
```
then you will loose information of dynamic origin of the metric. Publisher will have no way to get rid of unwanted namespace element which is "10.1.1.2"

If you go with 2, follow the example from [mock collector](https://github.com/intelsdi-x/snap/blob/master/plugin/collector/snap-collector-mock1/mock/mock.go#L56-L64) then **you don not loose** dynamic origin of the metric, because second `NamespaceElement` still have `Name` set, so it can be  recognized as dynamic in publishers/processors.

Hope it helps.

candysmurf:
@elemoine, @marcin-krolik : Thanks for your reply. I understood that my PR will not work for some of publishers and processors as there are dependencies on "IsDynamic" checking and can only check for the name field of the struct.

Please read issue #937, users can set the "name" for static elements. Purely checking if the name is not empty to assume the element is dynamic itself is flawed.

Note that if collectors didn't set the name during the data collection, even keeping the current IsDynamic checking for publishers or processors, it'll not achieve the result you expected.

I have my proposal in above comment to decouple the dependency on "IsDynamic" Please share your thoughts and if you still like to keep the coupling between Snap, Collectors, Processors, Publishers through "IsDynamic".

marcin-krolik:
@candysmurf I'm aware of #937. At the same time I don't think either removal of `IsDynamic` or introducing another breaking change in namespace creation (which would require to reimplement all plugins which offer dynamic metrics) will help solve this issue. 

First solution which comes to my mind could based on introduction of interface instead of `core.NamespaceElement` with two different structs implementing such interface. One being static, another dynamic. For static setting `Name` would not be possible. It would not break plugins and `IsDynamic` would still serve it's purpose.

candysmurf:
@marcin-krolik. It's a good suggestion to use interface. Let's think along this line. 

I'm afraid that many plugins will still need to be changed as any dynamic element will need to provide the name field in order for IsDynamic to work appropriately. The amount of refactoring may be not any less.  

Anyway, thanks for your suggestion, I like it!

jcooklin:
>It's a good suggestion to use interface
An interface cannot be decoded which will cause some headaches

I would consider addressing this issue with plugin authoring documentation.  I added a comment to #937 suggesting the same.

lmroz:
I see two options here:

- add to namespace validation a check if namespace element has a `Name` then have to have `*` as a Value. Perhaps this check after `CollectMetrics` should also be done to make sure no element changed it's definition (btw, maybe some info should be stripped before collection, like `Description` and added later to avoid pushing such data back and forth)
- allow namespace element to be Named or Dynamic or both. That would also solve problem when list of targets is known (for example, number of cpus), but it's presented in different way in metric list.

candysmurf:
@marcin-krolik, interface idea is brilliant and idiomatic to me. I was able to use custom marshal and unmarshal to address the polymorphism element slice. Unfortunately, GOB was not happy with the following error. e.g.
```
Failures:

  * /Users/egu/.gvm/pkgsets/go1.6.2/snap-latest/src/github.com/intelsdi-x/snap/scheduler/workflow_test.go
  Line 144:
  Expected: nil
  Actual:   'gob: wrong type (core.Namespace) for received field MetricType.Namespace_'
```
I tried to stick to the original variable and method name as much as possible. I'm not sure if Joel meant the same thing. 

@imroz, I didn't get what you suggest. Please elaborate. thanks.

core/metric.go
```diff
@@ -70,88 +172,100 @@ func (n Namespace) Key() string {
...
+
+// UnmarshalJSON unmarshals JSON into a Namespace
+func (n Namespace) UnmarshalJSON(data []byte) error {
+	objs := []*NamespaceJSON{}
+	return json.Unmarshal(data, &objs)
+}
+
+func init() {
+	gob.Register(&StaticElement{})
+	gob.Register(&DynamicElement{})
```
IRCody: I don't know if we want to have register going on in multiple places. 

We also need to use gob.RegisterName() to avoid import path issues in the future.

core/metric.go
```diff
@@ -45,8 +53,102 @@ type Metric interface {
...
+
+// Namespace defines array of namespace elements
 type Namespace []NamespaceElement
 
+// DynamicElement defines the dynamic element of a namespace
+type DynamicElement struct {
+	Value_       string
+	Description_ string
+	Name_        string
+}
```
IRCody: Should we have the functions for DynamicElement next to it's definition?

I also think it probably makes sense to have all the values be unexported since they are all exposed/modifiable via functions.

core/metric.go
candysmurf: Agreed, where do you suggest to put them?

core/metric.go
candysmurf: Agreed as well but they were unexported but what I learned is that GOB needs all fields to be exported. As I thought there is GOB issue, I exposed all of them now. Correct me if I miss anything.

core/metric.go
```diff
@@ -70,88 +172,100 @@ func (n Namespace) Key() string {
...
 	}
-	return false
+	return json.Marshal(objs)
+}
+
+// UnmarshalJSON unmarshals JSON into a Namespace
+func (n Namespace) UnmarshalJSON(data []byte) error {
+	objs := []*NamespaceJSON{}
+	return json.Unmarshal(data, &objs)
+}
```
IRCody: Do we need the json methods/struct? Where are these used at?

candysmurf: yes, we need them as they override json.Marshal and Unmarshal. Otherwise we'll not able to marshal the polymorphic namespace element slice.

control/plugin/metric.go
```diff
@@ -90,8 +90,8 @@ func NewPluginConfigType() ConfigType {
 // Represents a metric type. Only used within plugins and across plugin calls.
 // Converted to core.MetricType before being used within modules.
 type MetricType struct {
-	// Namespace is the identifier for a metric.
-	Namespace_ []core.NamespaceElement `json:"namespace"`
+	// Namespace_ is the identifier for a metric.
+	Namespace_ core.Namespace `json:"namespace"`
```
IRCody: Will this break compatibility with plugins?

candysmurf: Hopefully not. But it could require massive changes. We'll need to find it out if it works with GRPC first.

control/plugin/client/native.go
```diff
@@ -181,6 +182,8 @@ func (p *PluginNativeClient) GetMetricTypes(config plugin.ConfigType) ([]core.Me
 
 	args := plugin.GetMetricTypesArgs{PluginConfig: config}
 
+	fmt.Println("---args NATIVE---", args)
+
```
IRCody: This was for debugging?

control/plugin/metric.go
IRCody: Why did it need to be changed? Is there an example of how it would be used in the plugins? I think we need to be careful/thoughtful about changes that effect plugin authors.

candysmurf: @IRCody, It started from fixing a bug. The change shouldn't affect plugin authors much, but it'll make our code less vulnerable.

Of course, the gain is not much either as plugin authors can still ignore the feature Snap framework tries to provide and use dynamic metric as a static metric unless they specify the name for the dynamic metrics correctly. 

We can document this part. At least, Snap framework supports it.  

If all of you votes against the fix, we can document it as Joel suggested.

lmroz:
First solution I proposed is simply adding piece of code like this:
```
for _, elem := range ns {
	if elem.Name != "" && elem.Value != "*" {
		return meaningful error
	}
}
```
in these places:
https://github.com/intelsdi-x/snap/blob/master/control/metrics.go#L340
https://github.com/intelsdi-x/snap/blob/master/control/control.go#L1063

Second one is IMO better. We would have 3 types of `NamespaceElement`
- Static Anonymous Element - this is element that have `Value` other than `*` and `Name` empty, currently created with `AddStaticElement`
- Dynamic Named Element - element with `Value` = `*` and non-empty `Name`, currently created with `AddDynamicElement`
- Static Named Element - the one that caused "problem", has `Value` other than `*` and non-empty `Name`, this would allow asking plugin about concrete metric (not all metrics) and simultaneously enable publishers to treat this element as data and strip it from namespace. Additionaly user would have been informed in a consistent way about available metrics (that requires additional code to list such metric only once, not per each instance).

Anyway, validation like in first solution has to be added to disallow unwanted combination of empty `Name` and `Value` = `*`

candysmurf:
@lmroz, thanks for the details. 

The first option will not require a lot of changes if only two places that we need to patch. Please confirm this.

Agreed, the second one is better and I believe it's doable technically. It requires a lot of changes inside Snap.  

I already spent some effort similar to the second approach. I still need to tackle GRPC part and change a lots "[]core.Namespace" to []core.NamespaceElement" and test.  Changes will be extensive. Before I put all my effort into it, I would like to get your inputs. Basically, we have following suggestions:

1. Document it only (simplest, no code change)
2. Patch a few places (Lukas's option 1, I have no knowledge if it's only a few patchable places.)
3. Utilizes Interface, a lot of code changes 
4. Symbolic separation between dynamic and static elements (Emily's simple and dirty)

lmroz:
99.9% sure it's only these 2 places (AFAIK we don't touch binary data that goes out of processor), but remember that in CollectMetrics you have to check in metric catalog if namespace element was static and then collector assigned a `Name` (or the opposite).

candysmurf:
@lmroz, thanks for the confirmation and two links.

Basically, two scenarios we'll throw an error.

  Name != "" && Value != "*"
  Name == "" && Value == "*"

These two checks will be done for every metric collected. Do you have any concern regarding 
  1. performance
  2. throw errors

If not, this probably is the simplest fixable solution although it's not ideal.  Please share your thoughts.

candysmurf:
@lmroz, @marcin-krolik: Luaksz's proposal #1 is implemented. Please review and share your thought if Snap really wants to enforce the validation in control.go. I didn't throw errors instead logged them.

elemoine:
If I may chime in… This validation implies a loop over every collected metric, and for each metric a copy of every namespace element. As @jcooklin [commented](https://github.com/intelsdi-x/snap/issues/937#issuecomment-225390793) the issue can be addressed with documentation. For those reasons I personally would not add validation. My 2 cents.

candysmurf:
@elemoine, I'll be happy to remove the validation in control.go.  Let's hear from others.

We explored a couple of options this time. I appreciate all your contributions.  It was fun!

The Doc is added, please review it as well. thank you.

elemoine:
Thank you @candysmurf. Wouldn't it make sense to add the static/dynamic metric doc to the PLUGIN_AUTHORING file?

marcin-krolik:
@elemoine We definitely need such document since there is a lot of confusion around these two types of metrics, especially for people starting with Snap.

elemoine:
> @elemoine We definitely need such document since there is a lot of confusion around these two types of metrics, especially for people starting with Snap.

You mean you would create a separate document as @candysmurf has done?

marcin-krolik:
@candysmurf I went through the change you proposed in this PR and although I appreciate your work, you did great job, I'm not convinced by this solution.Do we really need to check every single metric and every single `NamespaceElement` in metrics' namespace just to make sure that plugin author did not set `NamespaceElement.Name` while `NamespaceElement.Value != "*"`? I think it is not necessary. I would rather stay with suggestion @jcooklin made and prepare proper documentation with examples.

marcin-krolik:
@elemoine I have not noticed this part somehow. There's no need for separate document apart from one created by @candysmurf . I guess it could be extended by adding some code examples based on mock collector implementation.

elemoine:
> @candysmurf I went through the change you proposed in this PR and although I appreciate your work, you did great job, I'm not convinced by this solution.Do we really need to check every single metric and every single NamespaceElement in metrics' namespace just to make sure that plugin author did not set NamespaceElement.Name while NamespaceElement.Value != "*"? I think it is not necessary. I would rather stay with suggestion @jcooklin made and prepare proper documentation with examples.

Exactly my opinion (expressed in https://github.com/intelsdi-x/snap/pull/977#issuecomment-226400267).

candysmurf:
##### Summary of the PR:

1.Added a check in metric.go to disallow setting the `Name` field for a static cell but requires it for a dynamic element.
2.Added docs/METRICS.md and linked it to docs/PLUGIN_AUTHORING.md

##### Summary of discussed solutions:
* Symbolic separation of dynamic metrics
    - Pros
        * Simple and enforce the `Name` stays with the `Value` of a dynamic element
        * No dependency on IsDynamic() method
        * No need to specify `Name` for every dynamic element
        * Publisher plugins have the freedom to handle the differences between dynamic and static metrics
    - Cons
        * Requires collector plugins to change the symbolic format
        * Requires publisher plugins to parse data on the new symbolic format
        * Deals with the dynamic metrics in each plugin could be entirely different
* Interface
    - Pros
        * Idiomatic and object oriented
        * two different types of static and dynamic elements
        * no dependency on `Name` field checking
        * standardizes the IsDynamic() checking for publishers
    - Cons
        * A lot of changes inside Snap including custom Marshal/Unmarshal, GRPC, Encoding/Decoding
        * Each collector plugin still _`have to`_ define dynamic element `Name` for using IsDynamic correctly
* Validation from Snap 
    - Procs
        * Simple, only two places
        * Fixes a static element shouldn't have `Name` defined issue
        * Checks if the `Name` field is defined for a dynamic element 
    - Cons
        * Costly in that it checks each element of every namespace
        * Dependency on setting the `Name` of each dynamic element correctly for collector plugins.
* Document Static vs Dynamic
    - Pros
        * Simple and always needed no matter which solution
    - Cons
        * Issue is not fixed and the design is vulnerable

control/metrics.go
```diff
@@ -357,7 +357,25 @@ func validateMetricNamespace(ns core.Namespace) error {
 	return nil
 }
 
+// validateDynamic validates a dynamic namespace qualification.
+// A dynamic element requires the name field while a static element does not.
+func validateDynamic(ns core.Namespace) error {
```
elemoine: I find the function name a bit confusing. To me it suggests that the function validates that the namespace is dynamic. Couldn't this validation be part of the `validateMetricNamespace` function instead?

docs/METRICS.md
```diff
@@ -0,0 +1,108 @@
...
+
+Snap Framework supports two types of metrics. They are static and dynamic metrics. A Snap metric consists of a namespace and value pair.
+
+### Static Metrics
+A namespace having no wildcard and only string literals separated by slashes is a static metric. For example:
+```
+/intel/cassandra/node/zeus/type/Cache/scope/KeyCache/name/Requests/OneMinuteRate
+/intel/cassandra/node/zeus/type/ClientRequest/scope/CASRead/name/Unavailables/FiveMinuteRate
+/intel/cassandra/node/apollo/type/Cache/scope/RowCache/name/Hits/OneMinuteRate
+/intel/cassandra/node/apollo/type/ClientRequest/scope/RangeSlice/name/Latency/OneMinuteRate
```
elemoine: Very minor but isn't the convention to use the snake notation in metric namespaces?

docs/METRICS.md
```diff
@@ -0,0 +1,108 @@
...
+```
+### Dynamic Metrics
+A namespace included at least one wildcard is a dynamic metric. For example:
```
elemoine: s/included/including ?

docs/METRICS.md
```diff
@@ -0,0 +1,108 @@
...
+ ```
+The _`NamespaceElement`_ forms each cell of a namespace and those cells are separated by slashes. 
+
+Create a new instance of a namespace:
+```
+1. // contains all string literals
+2. ns := core.NewNamespace("intel", "cassandra", "node")
+
+3. // contains a wildcard
+4. ns := core.NewNamespace("intel","cassandra", "node", "*")
```
elemoine: (nit) Space is missing after first comma.

But I think would not even advertise this. I would just advertise the `AddDynamicElement` function for the creation of dynamic elements.

docs/METRICS.md
```diff
@@ -0,0 +1,108 @@
...
+```
+The line #4 has the dynamic element _`*`_. If you leave it like that, you cannot fully leverage what Snap framework provides. Instead of doing line #4, you can do:
+```
+ns := core.NewNamespace("intel", "cassandra", "node")
+    .AddDynamicElement("node name", "description")
+```
+Another alternative for not providing the description:
+```
+ns := core.NewNamespace("intel","cassandra", "node", "*")
+ns[3].Name = "node name"
```
elemoine: I would not advertise this.

elemoine:
@candysmurf I've added a number of comments in the code. Feel free to ignore me if my comments are not valid/relevant :)

docs/METRICS.md
candysmurf: @elemoine, what's the snake notation? is it [1][2][3][4]?

elemoine: sorry_for_not_being_more_explicit :) 

https://en.m.wikipedia.org/wiki/Snake_case

candysmurf: @elemoine: thanks for the clarification! Agreed that snake_case could be easier to read comparing to CamelCase.  In my example, I copied them from the Cassandra collector. Those camel cases were from MX4J APIs. 

Are you suggesting that Snap should standardize cases in namespaces to Snake Case?

lynxbat: We do not have a standardization on this for namespaces. I prefer whatever is closest to the resources. In this case @candysmurf has then coupled well with the Cass collector which is good.

candysmurf:
@elemoine, thanks for your feedback! Please take another look and let me know if anything other suggestions. thank you.

docs/METRICS.md
lynxbat: I think it should hopefully be consistent inside a plugin. But it is not something we can control.

candysmurf:
I got an idea to enhance collectors so that they're TimeDB friendly. The good thing is that it'll not have a dependency on "IsDynamic()". There will be no modification to publishers and minimum changes to collectors. I'm implemented it in both Cassandra and Redfish collectors. They'll be ready for review soon.

candysmurf:
@jcooklin, Updating the doc. Please review this PR and let me know if anything you like me to change. thank you.

candysmurf:
@jcooklin , Well done! Thanks for the collaboration!

control/metrics.go
```diff
@@ -83,6 +83,14 @@ func errorMetricEndsWithAsterisk(ns string) error {
 	return fmt.Errorf("Metric namespace %s ends with an asterisk is not allowed", ns)
 }
 
+func errorMetricStaticElementHasName(value, name, ns string) error {
+	return fmt.Errorf("A static element %s should not define name %s for namespace %s.", value, name, ns)
+}
+
+func errorMetricDynamicElemnetHasNoName(value, ns string) error {
```
danielscottt: There's a typo here. should be
```go
func errorMetricDynamicElementHasNoName(value, ns string) error
```

candysmurf: corrected. Thanks!

danielscottt:
:+1:

PR #981, opened 2016-06-09
Better logs

obourdon:
Summary of changes:
- Added log truncation option
- Updated logrus version to match same than most plugins
- Adding log coloring command line option parameter

Testing done:
- small & legacy tests ran without error

@intelsdi-x/snap-maintainers

### DIFF
--- a/snapd.go
+++ b/snapd.go
@@ -64,6 +64,10 @@ var (
 		Name:  "log-truncate",
 		Usage: "Log file truncating mode. Default is false => append (true => truncate).",
 	}
+	flLogColors = cli.BoolTFlag{
+		Name:  "log-colors",
+		Usage: "Log file coloring mode. Default is true => colored (--log-colors=false => no colors).",
+	}
 	flLogLevel = cli.IntFlag{
 		Name:   "log-level, l",
 		Usage:  "1-5 (Debug, Info, Warning, Error, Fatal)",
@@ -101,6 +105,7 @@ const (
 	defaultGoMaxProcs  int    = 1
 	defaultLogPath     string = ""
 	defaultLogTruncate bool   = false
+	defaultLogColors   bool   = true
 	defaultConfigPath  string = "/etc/snap/snapd.conf"
 )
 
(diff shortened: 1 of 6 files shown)


obourdon:
I have created a brand new Go environment on my systems to see why the go dependencies in TravisCI fail but on my side everything runs perfectly fine. Could someone tell me what is wrong here? Thanks

obourdon:
Seems like I finally found out a way to reproduce this. Submitted new PR. CI currently running

obourdon:
Strangely enough the small profile test jobs have returned OK (0 value return code) however seems like they have not returned yet :-(

snapd.go
```diff
@@ -60,6 +60,14 @@ var (
 		Usage:  "Path for logs. Empty path logs to stdout.",
 		EnvVar: "SNAP_LOG_PATH",
 	}
+	flLogTruncate = cli.BoolFlag{
+		Name:  "log-truncate",
+		Usage: "Log file truncating mode. Default is false => append (true => truncate).",
+	}
+	flLogColors = cli.BoolTFlag{
+		Name:  "log-colors",
+		Usage: "Log file coloring mode. Default is true => colored (--log-colors=false => no colors).",
```
jcooklin: Do you also want `Value: defaultLogColors,` here?

obourdon: Seems like it is not possible with the cli library to specify the Value: for booleans.
BoolFlag => false by default
BoolTFlab => true by default

Please note also that the usage implied by this cli library is not symetrical:
BoolFlag log-truncate => false by default => --log-truncate sets it to true
I would have expected the behaviour to be
BoolTFlag log-colors => true by default => --no-log-colors sets it to false
However the way to get it working is by using --log-colors=false :-(

IRCody:
@obourdon: What conflicts are showing up with this?

obourdon:
@IRCody I did merge with latest sources so as to resolve conflicts. Everything was fine on my side and I do not understand why CI failed for GO 1.5 legacy.

IRCody:
@obourdon; Thanks for updating. There are some intermittent failures that we are still trying to fix with the build. I believe this was the failure you saw. I restarted that test.

obourdon:
(Re-)Rebased current work to get rid of conflicts. However still one error caught during CI test run. Is this one transient too ?

IRCody:
Yes 👎 . I re-triggered the build for.

IRCody:
LGTM

IRCody:
Thanks for adding the doc changes. LGTM.

PR #984, opened 2016-06-09
Distributed workflow

IRCody:
Adds the capability to remotely target nodes in a workflow. 

This completes the first phase for adding distributed workflow.

@intelsdi-x/snap-maintainers


danielscottt:
I looked very closely at the earlier PR that landed this into the distributed-workflow branch, as well as tested on my own machine.

LGTM 🎉

PR #989, opened 2016-06-11
Fixes #988 - Plugins with multiple config policy nodes may cause a panic

### LINKED ISSUE #988: Plugins with multiple config policy nodes may cause a panic
When there are multiple config policy nodes 
**and** there exists a metric which does not fall under any of the policies
**and** the root node does not contain a policy 

The following example causes a panic since no policies apply to `/intel/abc/baz`. 
**metric types**
/intel/abc/foo
/intel/def/bar
/intel/abc/baz

**config at**
/intel/abc/foo 
/intel/def/bar 

**root**
/intel

jcooklin:
When there are multiple config policy nodes 
**and** there exists a metric which does not fall under any of the policies
**and** the root node does not contain a policy 

The following example causes a panic since no policies apply to `/intel/abc/baz`. 
**metric types**
/intel/abc/foo
/intel/def/bar
/intel/abc/baz

**config at**
/intel/abc/foo 
/intel/def/bar 

**root**
/intel

----
- [x] @tiffanyfj to add test

### DIFF
--- a/pkg/ctree/tree_test.go
+++ b/pkg/ctree/tree_test.go
@@ -193,7 +193,7 @@ func TestConfigTree(t *testing.T) {
 			So(g, ShouldNotBeNil)
 		})
 
-		Convey("get is inbetween two nodes in tree", func() {
+		Convey("get is in between two nodes in tree", func() {
 			d1 := newMockNode()
 			d1.data = "a"
 			d2 := newMockNode()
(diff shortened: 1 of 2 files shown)


pkg/ctree/tree_test.go
```diff
@@ -219,6 +219,21 @@ func TestConfigTree(t *testing.T) {
...
 
+		Convey("doesn't panic on ns where the root and ns don't have a policy", func() {
+			d1 := newMockNode()
+			d1.data = "a"
+			d2 := newMockNode()
+			d2.data = "b"
+			c := New()
+			c.Add([]string{"intel", "foo", "sdi-x", "cody"}, d1)
+			c.Add([]string{"intel", "foo", "sdi-x", "nan"}, d2)
+			c.Freeze()
```
IRCody: I believe Get calls freeze for you on the first get so I'm not sure this is needed.

IRCody:
LGTM

PR #990, opened 2016-06-13
Fixes #986 Kubernetes dependency.

### LINKED ISSUE #986: Fix dependencies related to Kubernetes
There is a dependency in the snap godep file that no longer exists in the kubernetes repo. This is causing godep restore to fail as it cannot change into this directory and try to checkout the dependency at the revision in the Godeps.json file. Also, it appears the commit kubernetes(and related dependencies) is pinned to in snap Godeps.json file is to a commit that has been reverted. These dependencies should be updated.

nanliu:


nanliu:
Fixes #986, we can't move to appc 0.8.4 yet which eliminates several nested dependency due to: https://github.com/appc/spec/issues/624

tiffanyfj:
LGTM

PR #997, opened 2016-06-16
Fix immediately start collecting the new metrics after new plugin which shares the same namespace is loaded

### LINKED ISSUE #832: The task should immediately start collecting the new metrics after a plugin is loaded that shares the same namespace

mkleina:
Fixes #832

> Summary of changes (16-Jun-2016):
> - Add RefreshTaskSubscription() method to scheduler, which calls SubscribeDeps() and ValidateDeps() for all task dependencies (code moved from startTask() method and slightly modified)
> - Add SetTaskScheduler() method to pluginControl and pluginRunner for access to scheduler from plugin runner
> - Call scheduler's RefreshTaskSubscription() for all tasks from control_event.LoadPluginEvent event in runner
> - Add new mock plugin anothermock1 (based on mock2) with another name and namespace for testing purposes (plugin is used in RefreshTaskSubscription() unit test)

#### Summary of changes NEW VERSION (05-Jul-2016):
- Add new mock plugin anothermock1 (based on mock2) with another name and namespace for testing purposes
- Add taskIDData map to pluginControl. Map stores task IDs, config tree and metrics of every task, so there is no need to call scheduler's GetTasks() method (as scheduler should be unknown for control).
- Add methods AddTaskIDData and RemoveTaskIDData to manage task map on control site
- Call AddTaskIDData and RemoveTaskIDData from scheduler's createTask and removeTask methods to track current tasks list on control site
- Call ValidateDeps and SubscribeDeps in LoadPluginEvent for every task from map, which metrics matches loaded plugin metrics
- newCollectorJob gets now managesMetrics instead of collectsMetrics as argument to be able to call ValidateDeps before every run, so task immidiately goes invalid when it for example doesn't contain config for newly loaded plugin. ValidateDeps is called for every remote manager.
- Add some protobuf messages to implement gRPC calls in the future (for AddTaskIDData and RemoveTaskIDData).

Testing done:
- Manual
```
1. Load plugin anothermock1
2. Create and start task which collects all metrics from /intel/* and contains config data for both plugins
3. Load second plugin mock1 - snap immidiately collects data from mock1 (and anothermock1)
4. Load third plugin mock2 - snap immidiately switches to and collects data from mock2 (and anothermock1)
```
- Unit tests

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -224,9 +224,6 @@ func New(cfg *Config) *pluginControl {
 	c.pluginRunner.SetMetricCatalog(c.metricCatalog)
 	c.pluginRunner.SetPluginManager(c.pluginManager)
 
-	// Pass runner events to control main module
-	c.eventManager.RegisterHandler(c.Name(), c)
-
 	c.tasks = &taskDataMap{data: make(map[string]*taskData)}
 
 	// Start stuff
@@ -244,83 +241,6 @@ func New(cfg *Config) *pluginControl {
 	}
 
 	return c
-}
-
-func (p *pluginControl) HandleGomitEvent(e gomit.Event) {
-	switch v := e.Body.(type) {
-	case *control_event.LoadPluginEvent:
-		// Ignore plugins with type other than collector
-		if core.PluginType(v.Type).String() != "collector" {
-			return
-		}
-
-		// Get all known tasks data
-		for taskID, taskData := range p.tasks.data {
-			// Expand every task metric namespaces (with "*") to corresponding metric calalog namespaces
-			for _, tm := range taskData.metrics {
-
-				matchedNss, err := p.metricCatalog.MatchNamespaces(tm.Namespace())
-
-				if err != nil {
-					log.WithFields(log.Fields{
-						"_block":         "control",
-						"event":          v.Namespace(),
-						"plugin-name":    v.Name,
-						"plugin-version": v.Version,
-						"plugin-type":    core.PluginType(v.Type).String(),
-					}).Error("error matching task namespace with metric catalog")
-					continue
-				}
-
-				if len(matchedNss) > 0 {
-					depMts := []core.Metric{}
-					for _, ns := range matchedNss {
-
-						// Get expanded namespace data from metric catalog
-						if m, err := p.metricCatalog.Get(ns, tm.Version()); err == nil {
-							// Check if expanded namespace belongs to currently loaded plugin
-							if m.Plugin.TypeName() == core.PluginType(v.Type).String() && m.Plugin.Name() == v.Name {
-								depMts = append(depMts, &metric{
-									namespace: ns,
-									version:   tm.Version(),
-									config:    taskData.configTree.Get(ns.Strings()),
-								})
-							}
-						}
-					}
-
-					// Validate and subscribe deps for loaded plugin
-					errs := p.ValidateDeps(depMts, taskData.plugins)
-					if len(errs) > 0 {
-						log.WithFields(log.Fields{
-							"_block":         "control",
-							"event":          v.Namespace(),
-							"plugin-name":    v.Name,
-							"plugin-version": v.Version,
-							"plugin-type":    core.PluginType(v.Type).String(),
-						}).Error("error validating dependencies")
-						continue
-					}
-
-					cps := returnCorePlugin(taskData.plugins)
-					errs = p.SubscribeDeps(taskID, depMts, cps)
-					if len(errs) > 0 {
-						log.WithFields(log.Fields{
-							"_block":         "control",
-							"event":          v.Namespace(),
-							"plugin-name":    v.Name,
-							"plugin-version": v.Version,
-							"plugin-type":    core.PluginType(v.Type).String(),
-						}).Error("error subscribing dependencies")
-						p.UnsubscribeDeps(taskID, depMts, cps)
-						continue
-					}
-
-					p.metricCatalog.UpdateQueriedNamespaces(tm.Namespace())
-				}
-			}
-		}
-	}
 }
 
 func (p *pluginControl) AddTaskIDData(taskID string, metrics []core.RequestedMetric, configTree *cdata.ConfigDataTree, plugins []core.SubscribedPlugin) {
@@ -536,6 +456,8 @@ func (p *pluginControl) Load(rp *core.RequestedPlugin) (core.CatalogedPlugin, se
 		pl.Details.ExecPath = ""
 	}
 
+	p.refreshPluginSubscriptions(pl.Meta)
+
 	// defer sending event
 	event := &control_event.LoadPluginEvent{
 		Name:    pl.Meta.Name,
@@ -545,6 +467,66 @@ func (p *pluginControl) Load(rp *core.RequestedPlugin) (core.CatalogedPlugin, se
 	}
 	defer p.eventManager.Emit(event)
 	return pl, nil
+}
+
+func (p *pluginControl) refreshPluginSubscriptions(meta plugin.PluginMeta) {
+	f := map[string]interface{}{
+		"_block": "refresh-plugin-subscriptions",
+	}
+
+	// Ignore plugins with type other than collector
+	if core.PluginType(meta.Type).String() != "collector" {
+		return
+	}
+
+	// Get all known tasks data
+	for taskID, taskData := range p.tasks.data {
+		// Expand every task metric namespaces (with "*") to corresponding metric calalog namespaces
+		for _, tm := range taskData.metrics {
+
+			matchedNss, err := p.metricCatalog.MatchNamespaces(tm.Namespace())
+
+			if err != nil {
+				log.WithFields(f).Error("error matching task namespace with metric catalog")
+				continue
+			}
+
+			if len(matchedNss) > 0 {
+				depMts := []core.Metric{}
+				for _, ns := range matchedNss {
+
+					// Get expanded namespace data from metric catalog
+					if m, err := p.metricCatalog.Get(ns, tm.Version()); err == nil {
+						// Check if expanded namespace belongs to currently loaded plugin
+						if m.Plugin.TypeName() == core.PluginType(meta.Type).String() && m.Plugin.Name() == meta.Name {
+							depMts = append(depMts, &metric{
+								namespace: ns,
+								version:   tm.Version(),
+								config:    taskData.configTree.Get(ns.Strings()),
+							})
+						}
+					}
+				}
+
+				// Validate and subscribe deps for loaded plugin
+				errs := p.ValidateDeps(depMts, taskData.plugins)
+				if len(errs) > 0 {
+					log.WithFields(f).Error("error validating dependencies")
+					continue
+				}
+
+				cps := returnCorePlugin(taskData.plugins)
+				errs = p.SubscribeDeps(taskID, depMts, cps)
+				if len(errs) > 0 {
+					log.WithFields(f).Error("error subscribing dependencies")
+					p.UnsubscribeDeps(taskID, depMts, cps)
+					continue
+				}
+
+				p.metricCatalog.UpdateQueriedNamespaces(tm.Namespace())
+			}
+		}
+	}
 }
 
 func (p *pluginControl) verifySignature(rp *core.RequestedPlugin) (bool, serror.SnapError) {
--- a/control/control.go
+++ b/control/control.go
@@ -259,51 +259,54 @@ func (p *pluginControl) HandleGomitEvent(e gomit.Event) {
 						"plugin-version": v.Version,
 						"plugin-type":    core.PluginType(v.Type).String(),
 					}).Error("error matching task namespace with metric catalog")
-				} else {
-					if len(matchedNss) > 0 {
-						depMts := []core.Metric{}
-						for _, ns := range matchedNss {
+					continue
+				}
 
-							// Get expanded namespace data from metric catalog
-							if m, err := p.metricCatalog.Get(ns, tm.Version()); err == nil {
-								// Check if expanded namespace belongs to currently loaded plugin
-								if m.Plugin.TypeName() == core.PluginType(v.Type).String() && m.Plugin.Name() == v.Name {
-									depMts = append(depMts, &metric{
-										namespace: ns,
-										version:   tm.Version(),
-										config:    taskData.configTree.Get(ns.Strings()),
-									})
-								}
-							}
-						}
+				if len(matchedNss) > 0 {
+					depMts := []core.Metric{}
+					for _, ns := range matchedNss {
 
-						// Validate and subscribe deps for loaded plugin
-						errs := p.ValidateDeps(depMts, taskData.plugins)
-						if len(errs) > 0 {
-							log.WithFields(log.Fields{
-								"_block":         "control",
-								"event":          v.Namespace(),
-								"plugin-name":    v.Name,
-								"plugin-version": v.Version,
-								"plugin-type":    core.PluginType(v.Type).String(),
-							}).Error("error validating dependencies")
-						} else {
-							cps := returnCorePlugin(taskData.plugins)
-							errs = p.SubscribeDeps(taskID, depMts, cps)
-							if len(errs) > 0 {
-								log.WithFields(log.Fields{
-									"_block":         "control",
-									"event":          v.Namespace(),
-									"plugin-name":    v.Name,
-									"plugin-version": v.Version,
-									"plugin-type":    core.PluginType(v.Type).String(),
-								}).Error("error subscribing dependencies")
-								p.UnsubscribeDeps(taskID, depMts, cps)
-							} else {
-								p.metricCatalog.UpdateQueriedNamespaces(tm.Namespace())
+						// Get expanded namespace data from metric catalog
+						if m, err := p.metricCatalog.Get(ns, tm.Version()); err == nil {
+							// Check if expanded namespace belongs to currently loaded plugin
+							if m.Plugin.TypeName() == core.PluginType(v.Type).String() && m.Plugin.Name() == v.Name {
+								depMts = append(depMts, &metric{
+									namespace: ns,
+									version:   tm.Version(),
+									config:    taskData.configTree.Get(ns.Strings()),
+								})
 							}
 						}
 					}
+
+					// Validate and subscribe deps for loaded plugin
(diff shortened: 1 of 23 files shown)


plugin/collector/snap-collector-anothermock1/anothermock/anothermock.go
```diff
@@ -0,0 +1,145 @@
...
+	// Name of plugin
+	Name = "anothermock"
+	// Version of plugin
+	Version = 1
+	// Type of plugin
+	Type = plugin.CollectorPluginType
+)
+
+// Mock collector implementation used for testing
+type AnotherMock struct {
```
marcin-krolik: I'm not sure if I understand the need for another mock collector. Could you elaborate on that?

mkleina: It is needed in unit tests. There was already mock1 and mock2, but both of them have the same name mock, so framework just switches to new version instead of adding new metrics to running task (this was supported before). Another mock has another name and namespace, so task with /intel/*/foo metric will collect both /intel/mock/foo and /intel/anothermock/foo after this PR.

control/runner.go
```diff
@@ -66,11 +67,17 @@ type executablePlugin interface {
 	WaitForResponse(time.Duration) (*plugin.Response, error)
 }
 
+type scheduler interface {
```
marcin-krolik: I'm not sure if `scheduler` is not misleading here. Maybe would be good to think of another, more suitable name for this interface

marcin-krolik: I'm not sure if `scheduler` is not misleading here. Maybe would be better to think for another name for the interface

marcin-krolik:
`profile-legacy.cov` should not be included in the commit

marcin-krolik:
:+1:

mkleina:
@marcin-krolik 
> profile-legacy.cov should not be included in the commit
Ok, I just amended "Add unit tests" commit, removed profile-legacy.cov file and modified "profile.cov" with "profile*.cov" in .gitignore.

snapd.go
```diff
@@ -249,13 +249,11 @@ func action(ctx *cli.Context) {
...
-
-	coreModules = []coreModule{}
-
-	coreModules = append(coreModules, c)
 	s := scheduler.New(cfg.Scheduler)
+	c.SetTaskScheduler(s)
 	s.SetMetricManager(c)
-	coreModules = append(coreModules, s)
+
+	coreModules = []coreModule{c, s}
```
IRCody: Are we sure we want to pass control into scheduler and scheduler into control? This seems very odd.

lynxbat: Why does control ever need to talk to the scheduler? Listening for events on the control side makes sense. But control should never need to *know* the scheduler or that one even exists.

That would be a major architecture decision.

mkleina: I need access to scheduler from control to update task deps, what is another way to access tasks from control package?

mkleina: I need to talk to scheduler from control to update task deps on plugin load event. Is there any way to know that the new plugin was just loaded from scheduler point of view?

scripts/build.sh
```diff
@@ -17,7 +17,7 @@
 #See the License for the specific language governing permissions and
 #limitations under the License.
 
-GITVERSION=`git describe --always`
+GITVERSION=`git describe --always --exact-match 2> /dev/null || echo "$(git symbolic-ref HEAD 2> /dev/null | cut -b 12-)-$(git log --pretty=format:"%h" -1)"`
```
IRCody: This [commit](https://github.com/intelsdi-x/snap/blob/master/scripts/build.sh#L20) already made it into master. I think you can safely rebase it out of this PR.

snapd.go
IRCody: >So why talking in opposite way c.SetTaskScheduler(s) can't be implemented?

It's not that it can't it's that, at least imo, it shouldn't. Control shouldn't need to know/care/think about tasks. 

>Maybe just push this 2-way communication into already existing s.SetMetricManager(c), so after setting metric manager, scheduler will be able to talk to metric manager (as it works now) and vice versa?

What are you thinking is the way to add 2-way communication to that interface?

mkleina: > What do you think is the way to add 2-way communication to that interface?
I thought just about removing SetTaskScheduler() from control, modify scheduler's SetMetricManager() to something like:
```go
// Set metricManager for scheduler
func (s *scheduler) SetMetricManager(mm managesMetrics) {
	s.metricManager = mm
	mm.SetTaskScheduler(s) // <-- Just add this line
	schedulerLogger.WithFields(log.Fields{
		"_block": "set-metric-manager",
	}).Debug("metric manager linked")
}
```
And leave snapd.go unmodified, but in the end there is actually no difference as you said that plugin manager shouldn't know about scheduler.

Only possible way here would be to create something between scheduler and plugin manager, that could exchange communication between these modules (for example package with interface both for scheduler and plugin manager). But creating whole module just for exchanging data between two other modules makes a little mess, I think.

The best possible way in my opinion would be some sort of event manager shared between all modules (now every module has it's own event management), so it would be easy to pass events between modules. But on the other hand, this is really huge modification.

What do you think will be best way to do?

scripts/build.sh
mkleina: I just rebased to master.

snapd.go
IRCody: I'm not sure what the best way is. My concern isn't with where control gets a reference to scheduler, but that it has one at all. It seems to go against the idea of them being separate packages.

pluginControl.CollectMetrics gets a list of metrics to collect, if it's dynamic it could expand that to include all metrics currently in the metric catalog (including plugins loaded since the last task fire).  The example in the linked issue was:

>The scenario which presents that:
Step1:  Load `PluginA`
Step2:  Create/start `Task1` with "intel/\*"
=> Task1 is collecting all metrics from PluginA

>Step3a:  Load `PluginB1`
=> Task1 should immediately start collecting the new metrics from PluginB1
In current implementation (PR #803) the metrics are being collected only from pluginA

>Step3b:  Load `PluginB2` which requires config
=> Task1 should immediately start collecting the new metrics from PluginB2, but Task1 does not contain any config for the pluginB2, so after 10 times not receiving the metrics from this plugin, the Task1 will fail  (@jcooklin approved that the initial task which was valid isn’t now so it seems reasonable that it would fail)
 
In this example, does CollectMetrics not have all the information it needs to collect? If I send control a metric `/intel/*`, and the results of that dynamic query has changed from the previous collect,  it would be transparent to the scheduler.

I feel like maybe I am missing a reason why it wouldn't work this way?

danielscottt:
Hi @mkleina,

The limitation we're up against here is that Snap's `scheduler` pkg is holding the state of the namespace queries _with_ the tasks.  I believe this is probably not "right".  That being said, your solution is on the right path given the current State of Snap; I don't see another way.  However, I think we have an opportunity here to straighten out some of these separation-of-concerns issues.

Aside from the philosophical issue of the tight coupling of these two components, we also need to think about this solution's ramifications on the gRPC relationship `scheduler` and `control` have.  Having a reference to a `scheduler` inside `control` would require an implementation of a gRPC server for `scheduler`, and, `control` would then need to keep track of `scheduler` clients.  You can see [examples of this client tracking in the `scheduler` now](https://github.com/mkleina/snap/blob/3ee90eadc479b8bd74ceefd27cb7904c2860b7f1/scheduler/task.go#L461-L506). Right now, this relationship is one-way, as you know, and this affords us the simplicity of not needing to track clients on both sides.

I think to preserve the delineation of `control` and `scheduler`, moving that state into `control` is the best solution.  I think this can be accomplished by having each task hold only `core.Namespace`s rather than `core.Metric`s, and then have `control` cache queries via a map of query => query results.  As an added performance optimization, the structure holding the results on the right side of the map could also hold some state analogous to plugin's existing subscription mechanism.  This optimization could provide prefetched lookups against the metric trie, and preserve memory space when the cached lookup is no longer necessary!

When a new plugin is loaded or unloaded, `control` could catch its own event and update this map accordingly.

This, of course, requires significant refactoring of the `scheduler.task` structure, and would involve modifying `control.CollectMetrics`'s signature, but I think not only does it serve as a solution to this problem, but it also is better architecturally than what is currently in place :smile:.

mkleina:
@danielscottt 
I just implemented your conception of this PR related to last sync (store tasks data in map on control side), can you look at it? There is now totally no connection from control to scheduler and everything works correctly:
https://github.com/mkleina/snap/tree/fix-subscribe-after-plugin-load-v2

danielscottt:
Thanks for the update, @mkleina.  I'm taking a look now.

danielscottt:
Actually, would you mind opening a new PR, or overwriting the history on this one, so I can leave comments?

mkleina:
Yes, I will overwrite history of this PR, so you can leave comments here.

mkleina:
@danielscottt 
Done, PR history overwritten.

mkleina:
@danielscottt 
Done, PR history overwritten.

control/control.go
```diff
@@ -225,6 +239,83 @@ func New(cfg *Config) *pluginControl {
...
+
+				if err != nil {
+					log.WithFields(log.Fields{
+						"_block":         "control",
+						"event":          v.Namespace(),
+						"plugin-name":    v.Name,
+						"plugin-version": v.Version,
+						"plugin-type":    core.PluginType(v.Type).String(),
+					}).Error("error matching task namespace with metric catalog")
+				} else {
```
IzabellaRaulin: Consider using here `continue` instead of nesting the next things in _else_. 

See similar case in [/control/runner.go#L345](https://github.com/intelsdi-x/snap/blob/master/control/runner.go#L345)

control/control.go
```diff
@@ -225,6 +239,83 @@ func New(cfg *Config) *pluginControl {
...
+						// Validate and subscribe deps for loaded plugin
+						errs := p.ValidateDeps(depMts, taskData.plugins)
+						if len(errs) > 0 {
+							log.WithFields(log.Fields{
+								"_block":         "control",
+								"event":          v.Namespace(),
+								"plugin-name":    v.Name,
+								"plugin-version": v.Version,
+								"plugin-type":    core.PluginType(v.Type).String(),
+							}).Error("error validating dependencies")
```
IzabellaRaulin: As above, consider using here `continue` instead of nesting the next things in _else_.

control/control.go
```diff
@@ -225,6 +239,83 @@ func New(cfg *Config) *pluginControl {
...
+						} else {
+							cps := returnCorePlugin(taskData.plugins)
+							errs = p.SubscribeDeps(taskID, depMts, cps)
```
IzabellaRaulin: Please, correct me if I'm wrong, but in taskData.plugins there are only procesor and publisher plugins requested in task manifest (so, loading a new plugin does not change anything in context of processors/publishers). So, there is no need to (re)subscribe them.  As I know, there should be made only subscription of collector plugins - normally collector plugins are gathered based on metrics namespace in `SubscribeDeps` (see [snap/control/control.go#L757](https://github.com/intelsdi-x/snap/blob/master/control/control.go#L757)) but in this case the step of `gatherCollectors(mts)` could be skipped, so implementing such function for example, `SubcribeCollectorDeps(taskId string, collectorPlugins []core.Plugin)` which will do exactly the same as in lines [762-785](https://github.com/intelsdi-x/snap/blob/master/control/control.go#L762). So, that can be used here too to subscribe already loaded collector which metrics matches with requested query of metrics.   I'm also considering that all of that in `case *control_event.LoadPluginEvent` should be execute only when there is no pool for this plugin. Otherwise, when there is a pool for such plugin, but in different version, it should work as it's done before [see snap/control/runner.go#327](https://github.com/intelsdi-x/snap/blob/master/control/runner.go#L327). What do You think?

scheduler/job.go
```diff
@@ -239,16 +242,32 @@ func (c *collectorJob) Run() {
...
-		nss, err := c.collector.ExpandWildcards(rmt.Namespace())
+		nss, err := c.metricManager.ExpandWildcards(rmt.Namespace())
 		if err != nil {
 			// use metric directly from the workflow
 			nss = []core.Namespace{rmt.Namespace()}
 		}
 
 		for _, ns := range nss {
-			config := c.configDataTree.Get(ns.Strings())
+			config := c.workflow.configTree.Get(ns.Strings())
```
IzabellaRaulin: Could You explain me this change? Also, I would like to be sure that this configTree represents both Global and Task Configuration.

IzabellaRaulin:
Plese, remove binary files: plugin/snap-collector-anothermock1, plugin/snap-collector-mock1, and so on.

scheduler/job.go
mkleina: I need whole workflow to pass into gatherMetricsAndPlugins function, which is needed to revalidate dependencies on job run. Previously there was just configTree, which is already inside workflow.

Before:

mkleina: I need whole workflow to pass into gatherMetricsAndPlugins function, which is needed to revalidate dependencies on job run. Previously there was just configTree, which is already inside workflow.

mkleina:
@IzabellaRaulin 
Sure, I just removed binaries from Update proto buffer sources and unit tests commit.

scheduler/job.go
IzabellaRaulin: Ok, that's fine

mkleina:
@jcooklin @IRCody 
As we were talking on last sync, I removed plugin validation inside job run.

mkleina:
@jcooklin @IRCody 
As we were talking on last sync, I removed plugin validation inside job run.

mkleina:
@jcooklin @IRCody 
As we were talking on last sync, I removed plugin validation inside job run.

scheduler/scheduler_test.go
```diff
@@ -384,3 +389,83 @@ func testInboundContentType(node interface{}) {
...
+	mock1Path := path.Join(PluginPath, "snap-collector-mock1")
+	anotherMock1Path := path.Join(PluginPath, "snap-collector-anothermock1")
+
+	Convey("RefreshTaskSubscriptions()", t, func() {
+		Convey("Create new scheduler and set metric manager", func() {
+			s = New(GetDefaultConfig())
+			s.SetMetricManager(c)
+			err := s.Start()
+			So(err, ShouldBeNil)
+			time.Sleep(100 * time.Millisecond)
```
jcooklin: It appears that all of the sleeps in this test can be removed if you register and use an event handler.

control/metrics.go
```diff
@@ -453,6 +453,14 @@ func (mc *metricCatalog) RmUnloadedPluginMetrics(lp *loadedPlugin) {
 	mc.mutex.Lock()
 	defer mc.mutex.Unlock()
 	mc.tree.DeleteByPlugin(lp)
+
+	// Update metric catalog keys
+	mc.keys = []string{}
+	mts := mc.tree.gatherMetricTypes()
+	for _, m := range mts {
+		mc.keys = append(mc.keys, m.Namespace().Key())
+	}
```
jcooklin: It's interesting, I noticed just yesterday that unloading a running plugin does not kill it.  I was going to investigate this more tomorrow as I do not recall if the original intent was for the plugin to continue to run but not

lynxbat: I definitely think they should be killed if unloaded. In case of a security risk or bug an unload is a safety valve for "do not use".

jcooklin: Issue #1061 has been opened.

mkleina: I think that this issue is not related to my PR, so maybe future PR will fix that, right?

mkleina: I think that this issue is not related to my PR, so maybe separate PR will fix that, right?

jcooklin: Agreed, it will be fixed in a separate PR.

jcooklin:
@mkleina, will you confirm that this PR addresses #1052

mkleina:
@jcooklin Yes, if I reproduce #1052 steps, task starts back without problem (because of my fix in RmUnloadedPluginMetrics). But as it is described in #1052, it is needed to stop the task, load new plugin version, unload old one and start task again, or just unload old version before task fails 10 times, it doesn't work immediately on next collect job.

marcin-krolik:
@mkleina I did some exercises with your changes checking below scenarios

Scenario 1
- load `snap-collector-mock1` (with `/intel/mock/foo` metric), 
- start task collecting `/intel/*`
- stop task
- remove task
- unload `snap-collector-mock1` 
- load modified `snap-collector-mock1` (without `/intel/mock/foo` metric), 
- start new task

Scenario 2
- load `snap-collector-mock1` (with `/intel/mock/foo` metric), 
- start task collecting `/intel/*`
- stop task
- unload `snap-collector-mock1`
- load modified `snap-collector-mock1` (without `/intel/mock/foo` metric), 
- restart task

Scenario 3
- load `snap-collector-mock1` (with `/intel/mock/foo` metric), 
- start task collecting `/intel/*`
- unload `snap-collector-mock1` (this disables the task)
- load modified `snap-collector-mock1` (without `/intel/mock/foo` metric), 
- enable task and restart it

All above scenarios works fine with your changes. What is unclear to me, it's the second part of your last comment, especially 

>
it doesn't work immediately on next collect job.

Is there something I'm missing here?

mkleina:
@marcin-krolik I thought about scenario, when task is not disabled, stopped or removed, and plugin version is upgraded during task run, so on next collect new metric set is loaded, without task going to error state.

mkleina:
@marcin-krolik I thought about scenario, when task is not disabled, stopped or removed, and plugin version is upgraded during task run, so on next collect new metric set is loaded, without task going to error state. But I don't know if it is really needed.

IzabellaRaulin:
I've just added label `needs-2nd-review` to show that this pull request is a good example of using that ones - it needed specialized and in-depth review.

PR #1006, opened 2016-06-20
Customizable number of failures before snap disable the task

### LINKED ISSUE #967: Add the ability to configure the number of task retries
The task manifest should include an optional field for setting the number of times the task will be retried before failing.

The rest interface is currently setting retries to 10 ([here](https://github.com/intelsdi-x/snap/blob/master/mgmt/rest/task.go#L98)).

Providing a value of -1 should indicate that the task should never go into a disabled state.

iceycake:
Fixes #967 

Summary of changes:
- Adding a new optional field "failure" under Snap task header.  This field defines the number of consecutive failures when snap disable the task.  If the field is missing, the default value will be 10.
- Update the task documentation to reflect the change
- Adding --failure for the command line

Testing done:
- Creating task via task manifest using the example folder

@intelsdi-x/snap-maintainers

### DIFF
--- a/cmd/snapctl/flags.go
+++ b/cmd/snapctl/flags.go
@@ -126,7 +126,7 @@ var (
 		Name:  "deadline",
 		Usage: "The deadline for the task to be killed after started if the task runs too long (All tasks default to 5s)",
 	}
-	flTaskFailures = cli.IntFlag{
+	flTaskMaxFailures = cli.IntFlag{
 		Name:  "max-failures",
 		Usage: "The number of consecutive failures before snap disable the task (Default 10 consective failures)",
 		Value: DefaultMaxFailures,
--- a/cmd/snapctl/flags.go
+++ b/cmd/snapctl/flags.go
@@ -21,6 +21,10 @@ package main
 
 import "github.com/codegangsta/cli"
 
+const (
+	DefaultMaxFailures = 10
+)
+
 var (
 
 	// Main flags
@@ -121,6 +125,11 @@ var (
 	flTaskDeadline = cli.StringFlag{
 		Name:  "deadline",
 		Usage: "The deadline for the task to be killed after started if the task runs too long (All tasks default to 5s)",
+	}
+	flTaskFailures = cli.IntFlag{
+		Name:  "max-failures",
+		Usage: "The number of consecutive failures before snap disable the task (Default 10 consective failures)",
+		Value: DefaultMaxFailures,
 	}
 
 	// metric
--- a/cmd/snapctl/task.go
+++ b/cmd/snapctl/task.go
@@ -210,7 +210,7 @@ func createTaskUsingWFManifest(ctx *cli.Context) error {
 
 	// Deadline for a task
 	dl := ctx.String("deadline")
-	failure := uint(ctx.Int("failure"))
+	maxFailures := ctx.Int("max-failures")
 
 	var sch *client.Schedule
 	// None of these mean it is a simple schedule
@@ -279,7 +279,7 @@ func createTaskUsingWFManifest(ctx *cli.Context) error {
 		}
 	}
 	// Create task
-	r := pClient.CreateTask(sch, wf, name, dl, !ctx.IsSet("no-start"), failure)
+	r := pClient.CreateTask(sch, wf, name, dl, !ctx.IsSet("no-start"), maxFailures)
 	if r.Err != nil {
 		errors := strings.Split(r.Err.Error(), " -- ")
 		fmt.Println("Error creating task:")
--- a/cmd/snapctl/task.go
+++ b/cmd/snapctl/task.go
@@ -78,11 +78,12 @@ func trunc(n int) string {
 }
 
 type task struct {
-	Version  int
-	Schedule *client.Schedule
-	Workflow *wmap.WorkflowMap
-	Name     string
-	Deadline string
+	Version     int
+	Schedule    *client.Schedule
+	Workflow    *wmap.WorkflowMap
+	Name        string
+	Deadline    string
+	MaxFailures int `json:"max-failures"`
 }
 
 func createTask(ctx *cli.Context) error {
@@ -134,7 +135,14 @@ func createTaskUsingTaskManifest(ctx *cli.Context) error {
 		fmt.Println("Invalid version provided")
 		return errCritical
 	}
-	r := pClient.CreateTask(t.Schedule, t.Workflow, t.Name, t.Deadline, !ctx.IsSet("no-start"))
+
+	// If the number of failures does not specific, default value is 10
+	if t.MaxFailures == 0 {
+		fmt.Println("If the number of maximum failures is not specified, use default value of", DefaultMaxFailures)
+		t.MaxFailures = DefaultMaxFailures
+	}
+
+	r := pClient.CreateTask(t.Schedule, t.Workflow, t.Name, t.Deadline, !ctx.IsSet("no-start"), t.MaxFailures)
 
 	if r.Err != nil {
 		errors := strings.Split(r.Err.Error(), " -- ")
@@ -202,6 +210,7 @@ func createTaskUsingWFManifest(ctx *cli.Context) error {
 
 	// Deadline for a task
 	dl := ctx.String("deadline")
+	failure := uint(ctx.Int("failure"))
 
 	var sch *client.Schedule
 	// None of these mean it is a simple schedule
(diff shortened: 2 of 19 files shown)


cmd/snapctl/flags.go
```diff
@@ -122,6 +122,10 @@ var (
 		Name:  "deadline",
 		Usage: "The deadline for the task to be killed after started if the task runs too long (All tasks default to 5s)",
 	}
+	flTaskFailure = cli.StringFlag{
+		Name:  "failure",
+		Usage: "The number of consecutive failures before snap disable the task (Default 10 consective failures)",
```
IRCody: I think we can set a default value here and avoid the checks later that put in a default?

cmd/snapctl/task.go
```diff
@@ -135,7 +136,14 @@ func createTaskUsingTaskManifest(ctx *cli.Context) {
 		fmt.Println("Invalid version provided")
 		os.Exit(1)
 	}
-	r := pClient.CreateTask(t.Schedule, t.Workflow, t.Name, t.Deadline, !ctx.IsSet("no-start"))
+
+	// If the number of failure does not specific, default value is 10
```
IRCody: If the number of failures is not specified, use the default value

cmd/snapctl/task.go
```diff
@@ -135,7 +136,14 @@ func createTaskUsingTaskManifest(ctx *cli.Context) {
...
+	if t.Failure == 0 {
+		fmt.Println("Not specific the number of failures to disable the task.  Use default value: 10")
+		t.Failure = 10
```
IRCody: Since we reference the default failure number in multiple places, could we make it a const in this package so that if we decide to change it it's just in one place.

cmd/snapctl/task.go
```diff
@@ -135,7 +136,14 @@ func createTaskUsingTaskManifest(ctx *cli.Context) {
 		fmt.Println("Invalid version provided")
 		os.Exit(1)
 	}
-	r := pClient.CreateTask(t.Schedule, t.Workflow, t.Name, t.Deadline, !ctx.IsSet("no-start"))
+
+	// If the number of failure does not specific, default value is 10
+	if t.Failure == 0 {
+		fmt.Println("Not specific the number of failures to disable the task.  Use default value: 10")
```
IRCody: "The number of allowed failures not specified. Using default value of 10"

docs/TASKS.md
```diff
@@ -32,9 +33,15 @@ The schedule describes the schedule type and interval for running the task.  The
         "type": "cron",
         "interval" : "0 30 * * * *"
     },
+    "failure": 10,
 ```
 More on cron expressions can be found here: https://godoc.org/github.com/robfig/cron
 
+#### Failure
+By default, snap will disable a task if there is 10 consecutive error from any plugins within the workflow.  The configuration
+can be changed by specific the number of failures value in the task header.
```
IRCody: "can be changing by specifying..."

IRCody:
I am going to look into the CI failures. Looking at the failures I don't believe it is related to the changes in this PR.

iceycake:
@IRCody I defined constant and move the default value using cli flag instead.  Let me know how to fix the CI failures.

docs/TASKS.md
```diff
@@ -32,9 +33,15 @@ The schedule describes the schedule type and interval for running the task.  The
         "type": "cron",
         "interval" : "0 30 * * * *"
     },
+    "failure": 10,
 ```
 More on cron expressions can be found here: https://godoc.org/github.com/robfig/cron
 
+#### Failure
+By default, snap will disable a task if there is 10 consecutive error from any plugins within the workflow.  The configuration
```
ConnorDoyle: "if there are 10 consecutive errors"

docs/TASKS.md
```diff
@@ -32,9 +33,15 @@ The schedule describes the schedule type and interval for running the task.  The
...
+can be changing by specifying the number of failures value in the task header.
```
ConnorDoyle: "can be changed"

ConnorDoyle:
Nit: the field/flag name "failure" seems a little ambiguous. Something like "max-consecutive-failures" or even "max-failures" could buy clarity for just a few more characters.

iceycake:
@ConnorDoyle Changed from failure to max-failures in all the flag, json, and documentation according to the comment above.

mgmt/rest/task.go
```diff
@@ -95,7 +96,7 @@ func (s *Server) addTask(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
 	if tr.Name != "" {
 		opts = append(opts, core.SetTaskName(tr.Name))
 	}
-	opts = append(opts, core.OptionStopOnFailure(10))
+	opts = append(opts, core.OptionStopOnFailure(tr.MaxFailures))
```
IRCody: We need a way to handle the case where MaxFailures is not set and supply a default. I believe you can just check to see if tr.MaxFailures is equal to 0, if it is we can supple the default, if not we pass through the supplied value. 

This seems to be the root cause of the test failures since this particular test does not use the client library.

mgmt/rest/task.go
```diff
@@ -66,6 +66,7 @@ type task struct {
 	FailedCount        uint                    `json:"failed_count,omitempty"`
 	LastFailureMessage string                  `json:"last_failure_message,omitempty"`
 	State              string                  `json:"task_state"`
+	MaxFailure         uint                    `json:"max_failure"`
```
IRCody: Do we want this to be uint? I know as part of #967 we were thinking of using -1 (or I guess anything <0) to represent infinite retries. If that is the case we might just want this to be int?

iceycake: In the 2nd pull request will change that to uint.  Just try to make the minimum change here.

IRCody:
Can you rebase into one commit since the change is mostly one discrete feature addition?

Other than that LGTM.

iceycake:
@IRCody Combining both pull requests into just this one.  Can you close #1015 ?

IRCody:
@iceycake: Can you rebase this into 2 commits -- one adding the ability to set the # of allowed failures for a task, and the other to deal with -1 meaning an infinite # of retries? I think that will make it easier to review/merge.

Thanks!

iceycake:
@IRCody I think I made some mistakes when I squashed the commits into 2.  Some of the changes may be mixed up together a little bit.

scheduler/task.go
```diff
@@ -301,24 +301,31 @@ func (t *task) spin() {
...
+						t.state = core.TaskDisabled
+						t.Unlock()
+						// Send task disabled event
+						event := new(scheduler_event.TaskDisabledEvent)
+						event.TaskID = t.id
+						event.Why = fmt.Sprintf("Task disabled with error: %s", t.lastFailureMessage)
+						defer t.eventEmitter.Emit(event)
+						return
+					}
+				} else if consecutiveFailures > 0 {
```
danielscottt: I don't think we need this additional state in task failures.  If `stopOnFailure` is < 1, then we should just continue to tick the failures, and let the task run as is.  Sleeping an additional second can skew the way the task quanta are calculated.

iceycake: I was thinking to add an incremental delay here.  For example, a publisher may be not available for a long period of time.  It will be nice if the servers being monitored can give some breath room for the remote publisher.

ConnorDoyle: Backoff strategy could be a nice thing to add, but it shouldn't be conflated with the consecutive failure limit in this patch IMO. If added, it should probably be configurable per task. A 1 second delay here seems arbitrary.

iceycake: Agree.  I will remove that part of code for this PR.

danielscottt: >  For example, a publisher may be not available for a long period of time. It will be nice if the servers being monitored can give some breath room for the remote publisher.

:+1: FWIW, that type of failure "buffering" is something we're working on.  An RFC is forthcoming.  I'll make a note to tag you when it lands.

scheduler/task.go
```diff
@@ -301,25 +301,28 @@ func (t *task) spin() {
...
-					t.Lock()
-					t.state = core.TaskDisabled
-					t.Unlock()
-					// Send task disabled event
-					event := new(scheduler_event.TaskDisabledEvent)
-					event.TaskID = t.id
-					event.Why = fmt.Sprintf("Task disabled with error: %s", t.lastFailureMessage)
-					defer t.eventEmitter.Emit(event)
-					return
+				if t.stopOnFailure >= 0 {
```
danielscottt: One last nit, and then I think this is ready to land.  Now that we don't have the extra state, would you mind combining these nested `if`s with an `&&`?

danielscottt:
LGTM!  Thanks @iceycake!

PR #1008, opened 2016-06-21
Vendoring with Git submodules

elemoine:
I am opening this PR to follow up on the discussion in https://github.com/intelsdi-x/snap/issues/994.

Summary of changes:

- Used [Vandetta](https://github.com/dpw/vendetta) to populate the `vendor` directory with Git submodules for Snap's dependencies.
- Changed the submodules' commits based on the current content of `Godeps.json`.
- Removed `Godeps.json` and the associated `deps.sh` script.

This is what I've used for testing:

```
$ export GOPATH=/tmp/go
$ mkdir -p $GOPATH/src/github.com/intelsdi-x
$ cd $GOPATH/src/github.com/intelsdi-x
$ git clone -b vendor --single-branch https://github.com/elemoine/snap
$ cd snap
$ git submodule update --init --recursive
$ go install  # "make all" works as well
```

Vandetta is just used to create and update the submodules' commits. To get, build and install Snap one just uses the standard Go tools (`go get`). And by using submodules we avoid bloating the Snap repo.

PR #1009, opened 2016-06-21
Use go get -d to get Snap

### LINKED ISSUE #994: go get github.com/intelsdi-x/snap does not work
It looks like `go get github.com/intelsdi-x/snap` fails with the following error:

```
# github.com/intelsdi-x/snap/grpc/controlproxy/rpc
.gvm/pkgsets/go1.6/global/src/github.com/intelsdi-x/snap/grpc/controlproxy/rpc/control.pb.go:380: undefined: grpc.SupportPackageIsVersion2
```

elemoine:
Fixes #994

This PR fixes #994 by changing the `BUILD_AND_TEST` doc to use `go get -d` rather than `go get` to download Snap. This is to just download Snap instead of also building and installing it.

Building and installing Snap with `go get` is not safe, because `go get` fetches the latest commits of the dependencies, and Snap may not compile against these commits.

In the future, Snap may rely on the Vendoring functionality, and building it at `go get` time may be possible again. See #932, #933 and #1008 for experiments.

### DIFF
--- a/docs/BUILD_AND_TEST.md
+++ b/docs/BUILD_AND_TEST.md
@@ -30,9 +30,11 @@ To build the Snap Framework you'll need:
 
 The instructions below assume that the `GOPATH` environment variable has been set properly. Many of us use [go version manager (gvm)](https://github.com/moovweb/gvm) to easily switch between Go versions.
 
-Now you can install Snap into your `$GOPATH`:
+Now you can download Snap into your `$GOPATH`:
+
 ```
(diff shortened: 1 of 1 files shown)


mjbrender:
Checked edits and I like the explanation, especially with comments included. +1 to it if all agree that `-d` is the best option on build. @intelsdi-x/snap-maintainers

IRCody:
LGTM

PR #1015, opened 2016-06-21
If max-failures value is -1, snap will never disable a task with consecutive failure

### LINKED ISSUE #967: Add the ability to configure the number of task retries

iceycake:
Fixes #967 

Summary of changes:
- This is part 2 of issue#967 which will never disable a task with consecutive failure when max-failures is negative value

Testing done:
- Using a task manifest of max-failures value is -1

@intelsdi-x/snap-maintainers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,5 @@ tags
 
 # OSX stuff
 .DS_Store
+
+*.cov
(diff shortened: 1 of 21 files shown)


iceycake:
@IRCody This is part 2 of the pull request on issue#967

PR #1023, opened 2016-06-22
all signatures of cli commands changed to return error

mkhajeh:
All `ActionFunc` signatures has been updated to return `error` according to the latest changes in [cli](github.com/codegangsta/cli) package.
In this case, all the such functions return a predefined error which will be handled in `main()` function.
This commit eliminates the `DEPRECATED Action signature` error when executing a `snapctl` command.
The next essential step would be creating a more clear solution for raising and handling the errors.

@intelsdi-x/snap-maintainers

### DIFF
--- a/cmd/snapctl/config.go
+++ b/cmd/snapctl/config.go
@@ -52,7 +52,7 @@ func (c *config) loadConfig(path string) error {
 	return nil
 }
 
-func getConfig(ctx *cli.Context) {
+func getConfig(ctx *cli.Context) error {
 	pDetails := filepath.SplitList(ctx.Args().First())
 	var ptyp string
 	var pname string
@@ -66,7 +66,7 @@ func getConfig(ctx *cli.Context) {
 		if err != nil {
 			fmt.Println("Can't convert version string to integer")
 			cli.ShowCommandHelp(ctx, ctx.Command.Name)
-			os.Exit(1)
+			return errCritical
 		}
 	} else {
 		ptyp = ctx.String("plugin-type")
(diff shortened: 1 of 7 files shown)


cmd/snapctl/config.go
```diff
@@ -77,24 +77,24 @@ func getConfig(ctx *cli.Context) {
 	if ptyp == "" {
 		fmt.Println("Must provide plugin type")
 		cli.ShowCommandHelp(ctx, ctx.Command.Name)
-		os.Exit(1)
+		return errCritical
```
thomastaylor312: Won't `cli` handle printing the error text? It may be better to move the `fmt.Println` text into a custom error message here instead of returning a generic one.

cmd/snapctl/main.go
```diff
@@ -48,7 +50,9 @@ func main() {
 	app.Commands = append(commands, tribeCommands...)
 	sort.Sort(ByCommand(app.Commands))
 	app.Before = beforeAction
-	app.Run(os.Args)
+	if app.Run(os.Args) != nil {
```
thomastaylor312: This would be the place to print the errors passed back instead of with `fmt.Println`

mkhajeh:
@thomastaylor312 I agree, but the problem is that in different parts of the code, error handling behavior is different. For instance somewhere you need to print the error message alone and somewhere else you should show a help message alongside.
I thought that we can divide the problem to two parts. At first we eliminate the warning message that cli shows to us, and then we try to refactor to a cleaner solution for handling errors.

thomastaylor312:
@mkhajeh Good point. I agree. This LGTM.

danielscottt:
LGTM too. Landing.

PR #1025, opened 2016-06-22
Fix #1011: Group dynamic metrics in cache under appropriate keys

### LINKED ISSUE #1011: Dynamic metrics are not properly cached when returned from a plugin
When a dynamic metric is part of a task manifest, the metric will have the form of `/foo/bar/*/baz`. Prior to collecting this metric from a plugin, the cache will be checked to see if `/foo/bar/*/baz` exists. If the metric does not exist, then the metric will be collected from the plugin and the result stored in the cache.

When the dynamic metric is returned from the plugin, the metric now has a namespace in the form of `/foo/bar/hosta/baz`. The `updateCache` method will check to see if the metric's namespace is dynamic, and if it is dynamic, then the metric will be grouped with other metrics that fit the dynamic namespace. However, the current implementation of `updateCache` does not replace `hosta` with `*`, thus causing every metric returned from the collection of the dynamic metric `/foo/bar/*/baz` to be stored individually in the cache instead of as a grouping under the dynamic metric namespace. This also causes subsequent lookups in the cache for the metric `/foo/bar/*/baz` to return nil as `/foo/bar/*/baz` does not exist in the cache.

geauxvirtual:
Fixes #1011

Summary of changes:
- Adds control/strategy/cache_small_test.go file
- Adds TestUpdateCache method to cache_small_test.go to test adding static and dynamic metric entries to the cache
- Reimplement updateCache method to properly store dynamic metrics in cache under the metrics dynamic namespace instead of the namespace returned from the plugin.

Testing done:
- Wrote new tests to verify functionality of code changes.

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/strategy/cache.go
+++ b/control/strategy/cache.go
@@ -153,27 +153,31 @@ type listMetricInfo struct {
 }
 
 func (c *cache) updateCache(mts []core.Metric) {
-	dc := map[string]listMetricInfo{}
+	dc := map[string]*listMetricInfo{}
 	for _, mt := range mts {
-		isDynamic, _ := mt.Namespace().IsDynamic()
-		if isDynamic == false {
-			// cache the individual metric
-			c.put(mt.Namespace().String(), mt.Version(), mt)
-		} else {
-			// collect the dynamic query results so we can cache
-			key := fmt.Sprintf("%v:%v", mt.Namespace().String(), mt.Version())
+		isDynamic, idx := mt.Namespace().IsDynamic()
+		if isDynamic {
+			// cache dynamic metrics
+			dynNS := make(core.Namespace, len(mt.Namespace()))
+			copy(dynNS, mt.Namespace())
+			for _, v := range idx {
+				dynNS[v].Value = "*"
+			}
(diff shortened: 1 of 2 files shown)


control/strategy/cache.go
```diff
@@ -153,29 +153,32 @@ type listMetricInfo struct {
...
+					version:   mt.Version(),
 				}
 			}
-			var tmp = dc[key]
-			tmp.metrics = append(dc[key].metrics, mt)
-			tmp.namespace = mt.Namespace().String()
-			tmp.version = mt.Version()
-			dc[key] = tmp
+			dc[key].metrics = append(dc[key].metrics, mt)
+			for _, v := range dc {
```
danielscottt: For each dynamic metric, we are doing <code>n<sup>2</sup></code> writes where `n` is the number of metrics which are a result of the query.

This loop should be moved outside of the outer `mts` loop, and  done _iff_ `len(dc) > 0`.

danielscottt:
LGTM

Landing.

PR #1027, opened 2016-06-24
Update snap build scripts

nanliu:
* make SNAP_PATH optional.
* remove options around build-plugins.
* fix docker test so it works regardless of directory.
* simplify travis.ci config.


scripts/test.sh
```diff
@@ -17,85 +17,60 @@
...
+test_dirs=$(find . -type f -name '*.go' -not -path "./.*" -not -path "*/_*" -not -path "./Godeps/*" -not -path "./vendor/*" -print0 | xargs -0 -n1 dirname| sort -u)
+export test_dirs
+
+# shellcheck source=scripts/common.sh
+. "${__dir}/common.sh"
+
+_debug "script directory ${__dir}"
+_debug "project directory ${__proj_dir}"
+
+[[ "$SNAP_TEST_TYPE" =~ ^(small|medium|large|legacy)$ ]] || _error "invalid TEST_TYPE (value must be 'small', 'medium', 'large', or 'legacy', recieved:${SNAP_TEST_TYPE}"
```
geauxvirtual: Probably should say "Invalid SNAP_TEST_TYPE", though should a different error be thrown if SNAP_TEST_TYPE is not set?

And at the end of the line, recieved -> received.

nanliu:
@geauxvirtual updated error message per request.

geauxvirtual:
When running tests, seeing this in the output. A message is displayed goimports is not found. It doesn't stop from running other commands, not sure if we want that behavior though.
```
2016-06-30 16:56:00 UTC [     info] running goimports
/home/pulse/test991/src/github.com/intelsdi-x/snap/scripts/common.sh: line 82: goimports: command not found
```

scripts/common.sh
```diff
@@ -0,0 +1,123 @@
...
+
+  _path_prepend "${go_bin_path}"
+
+  _debug "GOPATH: ${GOPATH}"
+  _debug "PATH: ${PATH}"
+}
+
+_goimports() {
+  _go_get golang.org/x/tools/cmd/goimports
+  test -z "$(goimports -l -d $(find . -type f -name '*.go' -not -path "./vendor/*") | tee /dev/stderr)"
```
geauxvirtual: Should probably prefix any tool that is grabbed via `go get` with $GOPATH/bin so the tool can be run.

nanliu: Updated the test script to verify the path at the start, so we don't need to check the path later on per command. Also seems like our current test is missing folders since gofmt didn't catch errors in plugins dir.

scripts/deps.sh
```diff
@@ -17,18 +17,57 @@
...
+  local _dep=${_dep:=$(_detect_dep)}
+  _info "ensuring ${_dep} is available"
+  case $_dep in
+    godep)
+      _go_get github.com/tools/godep
+      ;;
+    glide)
+      _go_get github.com/Masterminds/glide
+      ;;
+    govendor)
```
geauxvirtual: Only other comment I have would be to remove glide and govendor from install_go_dep and restore_go_dep and reintroduce glide with the other PR.

nanliu: sure, that's all cleanup and removed now.

nanliu:
+1 Per discussion with Justin

PR #1036, opened 2016-06-28
Fix #1013: Standardisation for setting time when an entry is added to the cache

### LINKED ISSUE #1013: Adding cache entries should set time consistently.
There is a current mix use of chrono.Chrono.Now() and time.Now() for setting time when an entry is added to the cache.

https://github.com/intelsdi-x/snap/blob/master/control/strategy/cache.go#L101-L118

kindermoumoute:
Fixes #1013 

Standardisation of chrono.Chrono.Now() for setting time when an entry is added to the cache.

@intelsdi-x/snap-maintainers


danielscottt:
LGTM

PR #1039, opened 2016-06-29
Fix two test files to run tests only when legacy test type is specified.

geauxvirtual:
Summary of changes:
- Removes `/` from first comment in two files that was causing the build flag for testing to not be seen by  `go test`. This was running `legacy` tests for any test type specified.

Testing done:
- Verified tests files are only run for appropriate test type.

@intelsdi-x/snap-maintainers


candysmurf:
LGTM

PR #1047, opened 2016-07-01
Altered plugin test files and helper.go

kjang96:
Fixes #

Summary of changes:
- Got rid of BuildPlugin func in helper.go, replaced with CheckPluginBuilt
- ^because of that, plugin main_test.go no longer unnecessarily rebuilds plugins 
- instated check to main_test.go to check if plugin executables exist

Testing done:
- 

@intelsdi-x/snap-maintainers

### DIFF
--- a/plugin/publisher/snap-publisher-file/main_test.go
+++ b/plugin/publisher/snap-publisher-file/main_test.go
@@ -22,7 +22,6 @@ limitations under the License.
 package main
 
 import (
-	"fmt"
 	"os"
 	"path"
 	"testing"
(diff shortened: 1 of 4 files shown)


scripts/deps.sh
```diff
@@ -32,4 +32,4 @@ _go_path
 _go_get github.com/tools/godep
 
 _info "restoring dependency with godep"
-(cd "${__proj_dir}" && godep restore)
+(cd "${__proj_dir}" && godep restore)
```
IRCody: I think we should keep this newline.

plugin/publisher/snap-publisher-file/main_test.go
```diff
@@ -44,20 +43,18 @@ func TestFilePublisherLoad(t *testing.T) {
...
-		// Helper plugin trigger build if possible for this plugin
-		helper.BuildPlugin(PluginType, PluginName)
-		//
+
+	SkipConvey("make sure plugin has been built", t, func() {
+		err := helper.CheckPluginBuilt(SnapPath, PluginName)
+		So(err, ShouldBeNil)
 		//TODO cannot test this locally. We need AMQP and integration tests.
-		SkipConvey("ensure plugin loads and responds", t, func() {
+		SkipConvey("ensure plugin loads and responds", func() {
```
IRCody: I don't think these need to be SkipConvey's. [SkipConvey](https://github.com/smartystreets/goconvey/wiki/Skip) has the effect that the test is not run, we should be able to still verify that the plugin loads at the minimum.

I know that's what was there before but I think it makes sense to test that plugin is loadable as a minimal test.

IRCody:
Nan's commit should be removed from this PR since it was already merged in #1027.

IRCody:
LGTM

PR #1059, opened 2016-07-08
Fixes #1058: Adds code to ensure Snap restarts in response to a SIGHUP

### LINKED ISSUE #1058: Snap should restart in response to a SIGHUP
We should be able to use a `kill -HUP [PROC_ID]` (where the `[PROC_ID]` value is the process ID of the Snap daemon) to force the Snap daemon to restart (and pick up any changes that might have been made to the configuration in it's global configuration file).

tjmcs:
Fixes #1058

Summary of changes:
- Adds code to ensure that the gRPC server and RESTful server both shut down cleanly when their respective `Stop()` methods are called; previously these two servers only shut down because the Snap server exited in response to a SIGTERM, SIGINT, or SIGKILL (killing these threads)
- Adds code to ensure that any requests being processed by the gRPC and RESTful servers finish before their respective servers shut down (previously, these servers would shut down when the Snap server exited in response to the signals shown above, and the Snap server would not wait for any requests that were "in flight" to complete before exiting)
- Adds code to restart the Snap server in response to a SIGHUP (shutting down the Snap server and it's modules cleanly, then restarting the Snap server); the result will be that the Snap server can be forced to restart and pick up any changes that may have been made to it's global configuration file by sending a SIGHUP signal to the Snap process (eg. by running a `kill -HUP [PROC_ID]` where the `[PROC_ID]` value is the process ID of the Snap process).
- refactored the `mgmt/rest/tls.go` file, renaming it as `mgmt/rest/snapTLS.go` and renaming the `tls` struct it contained to use the name `snapTLS` instead; this helped us avoid shadowing of the `crypto/tls` library from the Go standard libraries with our `mgmt/rest/tls.go` file

Testing done:
- Manually started the Snap server as both an HTTP and HTTPS server (using an autogenerated certificate for the HTTPS server) where Snap's global configuration file was configured to set the Snap server log level to `INFO`
- verified that the server log level was set to `INFO` and that the server could be accessed in both of these scenarios using the `snapctl -u "[SNAP_URL]" plugin list` and `snapctl -u "[SNAP_URL]" --insecure plugin list` commands (respectively)
- modified the global configuration file being used so that it was configured to set the Snap server log level to `DEBUG`
- restarted the server using a `kill -HUP [PROC_ID]` command
- verified that the server log level was set to `DEBUG` and that the server could still be accessed in both of these scenarios using the `snapctl -u "[SNAP_URL]" plugin list` and `snapctl -u "[SNAP_URL]" --insecure plugin list` commands (respectively)

@intelsdi-x/snap-maintainers

### DIFF
--- a/docs/SNAPD_CONFIGURATION.md
+++ b/docs/SNAPD_CONFIGURATION.md
@@ -216,6 +216,17 @@ The same configuration settings above can also be provided in a JSON formatted c
 }
 ```
 
+## Restarting snapd to pick up configuration changes
+If changes are made to the configuration file, `snapd` must be restarted to pick up those changes. Fortunately, this is a simple matter of sending a `SIGHUP` signal to the `snapd` process. For example, the following command will restart the `snapd` process on the local system:
+
+```bash
+$ kill -HUP `pidof snapd`
+```
+
+Note that in this example, we are using the `pidof` command to retrieve the process ID of the `snapd` process. If the `pidof` command is not available on your system you might have to use a `ps aux` command and pipe the output of that command to a `grep snapd` command in order to obtain the process ID of the `snapd` process. Once the `snapd` process receives that signal it will restart and pick up any changes that have been made to the configuration file that was originally used to Âstart the `snapd` process.
+
+Do keep in mind that this signal will trigger a **restart** of the `snapd` process. This means that any running tasks will be shut down and any loaded plugins will be unloaded. In reality, this means that when the `snapd` process restarts any plugins not in the `auto_discover_path` will need to be loaded manually once the `snapd` process restarts (and any tasks not in that same `auto_discover_path` will need to be restarted). However, any plugins in the `auto_discover_path` will be automatically reloaded and any tasks in that same `auto_discover_path` will be automatically restarted when the when the `snapd` process restarts in response to a `SIGHUP` signal.
+
 ## More information
 * [SNAPD.md](SNAPD.md)
 * [REST_API.md](REST_API.md)
--- a/scheduler/distributed_task_test.go
+++ b/scheduler/distributed_task_test.go
@@ -29,8 +29,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/intelsdi-x/gomit"
 	"github.com/intelsdi-x/snap/control"
 	"github.com/intelsdi-x/snap/core"
+	"github.com/intelsdi-x/snap/core/scheduler_event"
 	"github.com/intelsdi-x/snap/core/serror"
 	"github.com/intelsdi-x/snap/grpc/controlproxy"
 	"github.com/intelsdi-x/snap/pkg/schedule"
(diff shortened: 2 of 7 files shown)


scheduler/distributed_task_test.go
```diff
@@ -112,24 +114,34 @@ func TestDistributedWorkflow(t *testing.T) {
...
-			tasks := sch.GetTasks()
-			var task core.Task
-			for _, v := range tasks {
-				task = v
-			}
-			So(task.State(), ShouldEqual, core.TaskDisabled)
+			// setup an event handler and wait for the 'TaskDisabledEvent' to arrive
+			respChan := make(chan bool)
+			sch.RegisterEventHandler("test", &failHandler{respChan})
+			ok := <-respChan
```
tjmcs: let's chat about this offline, @IRCody; @danielscottt and I were trying to get the test to pass using a select with a timeout and seemed to see the same sort of intermittent failures that we saw when the `time.Sleep(time.Second)` statement was used to try to give the task time to fail enough times to be disabled by the scheduler. Must be some sort of race condition in how the tasks are disabled, but didn't want to dig into the scheduler to try to find the root cause (and even if I knew where to look I assumed the fix wouldn't really make sense as part of this PR and should be submitted as a separate PR focused specifically on that issue).

That being said, I'm fully supportive of trying to describe the problem we uncovered in this test as part of a separate issue (tagged/labeled as a bug?) to make sure we don't lose track of it. Like I said, let's discuss it and see if we can describe what the problem is as a new issue that can be tracked moving forward.

andrzej-k:
This feature needs to be documented, possibly here (https://github.com/intelsdi-x/snap/blob/master/docs/SNAPD.md) and/or here (https://github.com/intelsdi-x/snap/blob/master/docs/SNAPD_CONFIGURATION.md)

andrzej-k:
Some license headers need update to include year 2016, all files changed this year should have: "Copyright 2015-2016 Intel Corporation"

tjmcs:
I'll defer to others on the team regarding updates to the copyright notice in those files. I haven't updated the copyright notice on files I've modified in the past, and I'm not sure I need to modify any of the dates on these files. Is this something new in the process (to update the dates on any files we modify)? If so, when was this introduced as part of the process of submitting a PR?

As to the documentation, I'll see what I can add to this PR.

tjmcs:
@andrzej-k, I added a section to the end of the [SNAP_CONFIGURATION.md](docs/SNAPD_CONFIGURATION.md) file; hopefully that's sufficient to address your concerns regarding documenting this new feature.

docs/SNAPD_CONFIGURATION.md
```diff
@@ -216,6 +216,15 @@ The same configuration settings above can also be provided in a JSON formatted c
...
 
+## Restarting snapd to pick up configuration changes
+If changes are made to the configuration file, `snapd` must be restarted to pick up those changes. Fortunately, this is a simple matter of sending a `SIGHUP` signal to the `snapd` process. For example, the following command will restart the `snapd` process on the local system:
+
+```bash
+$ kill -HUP `pidof snapd`
+```
+
+Note that in this example, we are using the `pidof` command to retrieve the process ID of the `snapd` process. If the `pidof` command is not available on your system you might have to use a `ps aux` command and pipe the output of that command to a `grep snapd` command in order to obtain the process ID of the `snapd` process. Once the `snapd` process receives that signal it will restart and pick up any changes that have been made to the configuration file that was originally used to Âstart the `snapd` process.
+
```
jcooklin: @tjmcs I know it's implied but do you think we should mention that plugins and tasks that were loaded/created/started after snapd started will not be preserved?

tjmcs:
Added some text around what happens to loaded plugins and running tasks when the `snapd` process restarts, @jcooklin; take a look and LMK what you think.

jcooklin:
LGTM

PR #1083, opened 2016-07-16
Fixed #1080 rules for int/floats crash plugins

### LINKED ISSUE #1080: Minimum and maximum values config policy rules for int/floats crash plugins when used
Adding either
```go
cpolicy.NewIntegerRule("crashes", false, 0, 1, 2) 
```
or
```go
cpolicy.NewFloatRule("crashes", false, 0, 1, 2)
``` 
to a plugins GetConfigPolicy method will cause the plugin to crash on startup. This is because the min/max values are not handled properly [here](https://github.com/intelsdi-x/snap/blob/master/control/plugin/cpolicy/float.go#L140-L156) and [here](https://github.com/intelsdi-x/snap/blob/master/control/plugin/cpolicy/integer.go#L48-L64). The failing logic for both is:

```go
	options := make([]*int, 1)
	for i, o := range opts {
		options[i] = &o
	}

	return &IntRule{
		key:      key,
		required: req,
		default_: options[0],
	}, nil
```

This causes a crash because, in the case where len(opts) is greater than 1 (i.e. if a min or max is set), it will cause out of bounds error on options since it's statically size 1. A second issue is that the minimum/maximum are never set in the returned rule.

candysmurf:
Fixes #1080 

Summary of changes:
- Checks "opts" length before using it
- New func only sets the default
- max and min have to be set using setters

Testing done:
- unit tests

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/plugin/cpolicy/float.go
+++ b/control/plugin/cpolicy/float.go
@@ -136,23 +136,22 @@ func (f *FloatRule) GobDecode(buf []byte) error {
 	return nil
 }
 
-// NewFloatRule Returns a new float-typed rule. Arguments are key(string), required(bool), default(float64), min(float64), max(float64)
+// NewFloatRule returns a new float-typed rule. Arguments are key(string), required(bool), default(float64)
 func NewFloatRule(key string, req bool, opts ...float64) (*FloatRule, error) {
 	// Return error if key is empty
 	if key == "" {
 		return nil, EmptyKeyError
 	}
 
-	options := make([]*float64, 1)
-	for i, o := range opts {
-		options[i] = &o
-	}
-
-	return &FloatRule{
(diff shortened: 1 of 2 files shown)


control/plugin/cpolicy/float.go
```diff
@@ -148,11 +148,15 @@ func NewFloatRule(key string, req bool, opts ...float64) (*FloatRule, error) {
 		options[i] = &o
```
IRCody: This will still cause panic if len(opts) > 1.

candysmurf: ah, good catch.this code is no longer needed.

IRCody:
LGTM

PR #1053, opened 2016-07-07
Renamed file publisher plugin and updated names/imports

thomastaylor312:
This PR came from discussion on #1046 

Summary of changes:
- Renamed the file publisher plugin to `snap-publisher-mock-file`
- Created [new file publisher repo](https://github.com/intelsdi-x/snap-plugin-publisher-file) for non-testing file publishing
- Updated paths and meta data for the plugin

Testing done:
- Small unit tests

@intelsdi-x/snap-maintainers

PR #1085, opened 2016-07-18
Rename snap plugin to match existing convention.

nanliu:
All plugins have been renamed from snap-* to snap-plugin-* to match
existing plugin conventions.


thomastaylor312:
@nanliu Renaming plugins can be quite a nightmare as I found in [this PR](https://github.com/intelsdi-x/snap/pull/1053). You'll probably want to check all of the same files I modified so the unit tests and example files use them correctly

nanliu:
Tracked down all the files.

thomastaylor312:
LGTM

PR #1090, opened 2016-07-21
Fix plugin catalog unknown bug

nanliu:
Fixes github pages where the plugin showed up in the wrong category, and missing description.

@intelsdi-x/snap-maintainers


sarahjhh:
lgtm

PR #1092, opened 2016-07-21
Fixes #1061: ensures that all instances of a running plugin are stopped

### LINKED ISSUE #1061: Unloading a plugin should kill instances of the plugin that are running
Given you start the example mock-file.yaml task and then unload the mock plugins the latest plugin, which was the one running, will still be running after unloading it.

tjmcs:
Fixes #1061

Summary of changes:
- Adds code to stop all instances of a running plugin when that plugin is unloaded (and no ohter versions of that plugin can be found)
- Adds code to avoid the creation of *zombie proceses* when these plugin instances are stopped (previously, the call that killed these processes did not wait for them to exit, now it does)

Testing done:
- started `snapd` and had it autoload a number of plugins (including two versions of the mock collector plugin)
- started a task that used the mock collector plugin
- unloaded both versions of the mock collector plugin
- ensured that the underlying processes were no longer appeared in the process table (even though `snapd` was still running)

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/plugin/execution.go
+++ b/control/plugin/execution.go
@@ -53,8 +53,20 @@ type commandWrapper struct {
 }
 
 func (cw *commandWrapper) Path() string { return cw.cmd.Path }
-func (cw *commandWrapper) Kill() error  { return cw.cmd.Process.Kill() }
-func (cw *commandWrapper) Start()       { cw.cmd.Start() }
+func (cw *commandWrapper) Kill() error {
+	// first, kill the process wrapped up in the commandWrapper
+	if err := cw.cmd.Process.Kill(); err != nil {
+		log.WithFields(log.Fields{
+			"_block": "Kill",
+		}).Error(err)
+		return err
+	}
+	// then wait for it to exit (so that we don't have any zombie processes kicking
+	// around the system)
+	_, err := cw.cmd.Process.Wait()
+	return err
+}
+func (cw *commandWrapper) Start() { cw.cmd.Start() }
 
 // Initialize a new ExecutablePlugin from path to executable and daemon mode (true or false)
 func NewExecutablePlugin(a Arg, path string) (*ExecutablePlugin, error) {
--- a/control/runner.go
+++ b/control/runner.go
@@ -286,7 +286,9 @@ func (r *runner) HandleGomitEvent(e gomit.Event) {
 			return
 		}
 	case *control_event.UnloadPluginEvent:
-		// On plugin unload,  find the key and pool info for the plugin being unloaded.
+		// On plugin unload, first find the key and pool info for the plugin being unloaded
+		// so that we can determine what, if any, subscriptions need to be moved to another
+		// version of that plugin (if one is available).
 		r.availablePlugins.RLock()
 		var pool strategy.Pool
 		var k string
@@ -301,14 +303,24 @@ func (r *runner) HandleGomitEvent(e gomit.Event) {
 
 		r.availablePlugins.RUnlock()
 		if pool == nil {
+			// if we didn't find the named plugin in one of our pools, then simply return
 			return
 		}
-		// Check for the highest lower version plugin and move subscriptions that
-		// are not bound to a plugin version to this pool.
+		// Otherwise, we need to roll any subscriptions that may have existed for the plugin being
+		// unloaded to a new version (if we can find one). First, check for the highest version of the
+		// plugin being unloaded that is still available (if there is one)
 		plugin, err := r.pluginManager.get(fmt.Sprintf("%s:%s:%d", core.PluginType(v.Type).String(), v.Name, -1))
 		if err != nil {
+			if err == ErrPluginNotFound {
+				// if we got to here via a "PluginNotFoundError", then we can't find
+				// another version of this plugin, so we need to kill all instances
+				// of the plugin from this pool
+				pool.KillAll("Last version of plugin unloaded; no additional versions to roll over to")
+			}
 			return
 		}
+		// if we found another version, then move the subscriptions that are not currently bound to a
+		// plugin version to our pool
 		newPool, err := r.availablePlugins.getOrCreatePool(plugin.Key())
 		if err != nil {
 			return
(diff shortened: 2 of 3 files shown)


control/runner.go
```diff
@@ -286,7 +286,7 @@ func (r *runner) HandleGomitEvent(e gomit.Event) {
 			return
 		}
 	case *control_event.UnloadPluginEvent:
-		// On plugin unload,  find the key and pool info for the plugin being unloaded.
+		// On plugin unload, first find the key and pool info for the plugin being unloaded.
```
IRCody: Maybe add something to this comment about the reason why we are looking for that info - In order to move any subscriptions that we are able to move for this plugin.

control/runner.go
```diff
@@ -301,14 +301,24 @@ func (r *runner) HandleGomitEvent(e gomit.Event) {
...
 			return
 		}
-		// Check for the highest lower version plugin and move subscriptions that
-		// are not bound to a plugin version to this pool.
+		// Otherwise, we need to roll any subscriptions that may have existed for the plugin being
+		// unloaded to a new version (if we can find one). First, check for the highest version of the
+		// plugin being unloaded that is still available (if there is one)
 		plugin, err := r.pluginManager.get(fmt.Sprintf("%s:%s:%d", core.PluginType(v.Type).String(), v.Name, -1))
 		if err != nil {
+			if r.pluginManager.(*pluginManager).isPluginNotFoundErr(err) {
```
IRCody: These are all in the same package so I think you can just do if err == ErrPluginNotFound and get rid of this assertion + the isPluginNotFound function.

control/strategy/pool.go
```diff
@@ -297,6 +298,20 @@ func (p *pool) Kill(id uint32, reason string) {
 	}
 }
 
+// Kill all instances of a plugin
+func (p *pool) KillAll(reason string) {
+	for _, rp := range p.plugins {
+		rp.Stop(reason)
```
IRCody: Looking at it, would it make sense to do this for the loop instead to avoid duplication?

```go
for id := range p.plugins {
    p.SelectAndKill(rp, reason)
}
```

tjmcs: I don't see how that would work, @IRCody; the arguments to the `SelectAndKill(...)` method are a string providing the task ID we're removing instances for and a string providing the reason; the `rp` value you're showing in your snippet, above (even if I did change the for back to `for _, rp := range p.plugins { ... }` is an AvailablePlugins interface, not a string.  Thoughts on how you'd correct that type mismatch?

(Note; the `rp` value in the `SelectAndKill(...)` method comes from the `Remove(...)` call, which depends on having a task ID to use to select which plugin instances we're removing, and we don't have one of those...)

IRCody: You're right. I was conflating the id in the MapAvailablePlugin to the taskid.

IRCody: Looking at it again can we just re-use the Kill() function? It takes the right inputs and I think it fits that KillAll will just look through and call kill on each plugin individually.

control/plugin/execution.go
```diff
@@ -53,8 +53,20 @@ type commandWrapper struct {
...
+	// first, kill the process wrapped up in the commandWrapper
+	if err := cw.cmd.Process.Kill(); err != nil {
+		log.WithFields(log.Fields{
+			"_block": "Kill",
+		}).Error(err)
+		return err
+	}
+	// then wait for it to exit (so that we don't have any zombie processes kicking
+	// around the system)
+	_, err := cw.cmd.Process.Wait()
```
candysmurf: Nice work! if an error occurred, do you like to log it?

control/runner.go
```diff
@@ -286,7 +286,9 @@ func (r *runner) HandleGomitEvent(e gomit.Event) {
 			return
 		}
 	case *control_event.UnloadPluginEvent:
-		// On plugin unload,  find the key and pool info for the plugin being unloaded.
+		// On plugin unload, first find the key and pool info for the plugin being unloaded
+		// so that we can determine what, if any, subscriptions need to be moved to another
```
candysmurf: love these added comments!

control/strategy/pool.go
```diff
@@ -297,20 +299,43 @@ func (p *pool) Kill(id uint32, reason string) {
...
+	}
+	if err := rp.Stop(reason); err != nil {
+		log.WithFields(log.Fields{
+			"_block": "SelectAndKill",
 			"taskID": id,
 			"reason": reason,
 		}).Error(err)
 	}
-	rp.Stop(reason)
 	if err := rp.Kill(reason); err != nil {
```
candysmurf: I'm wondering if a kill fails if we would like to retry once at least?

control/plugin/execution.go
jcooklin: I think this should be a call to Process.Release() instead of Wait().

jcooklin: An alternative would be to call Wait() above (instead of Kill).  Wait releases any resources associated with the process.

control/runner.go
```diff
@@ -301,14 +303,24 @@ func (r *runner) HandleGomitEvent(e gomit.Event) {
...
+		// Otherwise, we need to roll any subscriptions that may have existed for the plugin being
+		// unloaded to a new version (if we can find one). First, check for the highest version of the
+		// plugin being unloaded that is still available (if there is one)
 		plugin, err := r.pluginManager.get(fmt.Sprintf("%s:%s:%d", core.PluginType(v.Type).String(), v.Name, -1))
 		if err != nil {
+			if err == ErrPluginNotFound {
+				// if we got to here via a "PluginNotFoundError", then we can't find
+				// another version of this plugin, so we need to kill all instances
+				// of the plugin from this pool
+				pool.KillAll("Last verstion of plugin unloaded; no additional versions to roll over to")
```
jcooklin: s/verstion/version/

tjmcs:
@jcooklin; just pushed changes in response your suggestions. Take a look and LMK what you think of the revised version.

jcooklin:
LGTM

IRCody:
LGTM

Issue #1052, opened 2016-07-07
Unable to load upgraded collector with reduced set of metrics


marcintao:
When new version of collector is exposes a reduced set of metrics (doesn't export them from `GetMetricTypes()`) , it's impossible to restart a task involving that collector. Snap will complain about missing metric during task creation, even though:
- task was removed,
- plugin was unloaded and loaded in new version (with version number changed),
- list of metrics reflects tracks the current state of plugins.

**OS Version**: Ubuntu 16.04
**Snap version**: `snapd version master-cc4eda7` (tip of master)

**Steps to reproduce**:
* start snap
* load a mock collector (snap-collector-mock1) and file collector,
* start a task involving mock collector,
* verify list of metrics,
* stop 'n' remove the task,
* unload mock1 plugin,
* load 'upgraded' mock1 plugin (missing metric `/intel/mock/foo`),
* start a task involving mock1.

**Actual results:**
* task fails to load due to missing metric `/intel/mock/foo`

**Expected results:**
* task starts successfully

**Complete report of activity:**
[bug_upgrading_log.txt](https://github.com/intelsdi-x/snap/files/352155/bug_upgrading_log.txt)

geauxvirtual:
Can you add the task manifest you were using to this issue?

marcintao:
Added task manifest:
[bug_task.txt](https://github.com/intelsdi-x/snap/files/353678/bug_task.txt)

andrzej-k:
Closing - fixed by #1146

Issue #852, opened 2016-04-12
It seems that CollectMetrics method in PluginControl has an unnecessary argument


katarzyna-z:
It seems that CollectMetrics method in PluginControl has an unnecessary argument, deadline time.Time is not used in CollectMetrics method, see: https://github.com/intelsdi-x/snap/blob/master/control/control.go#L900. Am I right?

IRCody:
You are right that it is not used. I am not sure of the history behind it though. Any thoughts @intelsdi-x/snap-maintainers?

jcooklin:
The deadline is currently being used in the scheduler so that jobs aren't started if the deadline has already passed ([here](https://github.com/intelsdi-x/snap/blob/master/scheduler/worker.go#L51-L56)).  The intent was to also use the deadline to trigger cancellation of the RPC call which would need to be done in the proxies for collect, process and publish.  

Labeling as "bug".

Relates to #965

marcin-krolik:
Fixed with #1109

PR #1109, opened 2016-07-29
Simplified interaction between scheduler and control

### LINKED ISSUE #832: The task should immediately start collecting the new metrics after a plugin is loaded that shares the same namespace

marcin-krolik:
Fixes #832, #852, #1052 
Closes #997 

Simplified interaction between scheduler and control. Introduces new subscription mechanism with task view. Plugin loading/unloading triggers subscribtion/unsubscribtions from pools.

Link to [design](https://intelsdi.atlassian.net/wiki/display/~jrcookli/Simplifying+the+interaction+between+scheduler+and+control) document

Summary of changes:
- subscriptionGroups in `pluginControl` for storing (requests, expanded metrics, plugins and configuration)
- No more expanding metrics on each collect job
- `ValidateDeps` takes care of expanding metrics when task is created
- subscriptionGroups are added/removed when tasks are started/stopped (`SubscribeDeps`/`UnsubscribeDeps`) 
- `CollectMetrics` leverages subscriptionGroups
- subscriptionGroups are processed on plugin load/unload events
- No need for bound vs unbound subscriptions with respect to pools 

## scenarios
* [upgrade collector] load plugin mock v1, start task mock-file.yaml (not pinned), load mock v2
* Expected mock v1 is unsubscribed (not running), mock v2 is subscribed (running)
scenario1

* [downgrade collector] load plugin mock v2, start task mock-file.yaml (not pinned), load v1
* Expected mock v2 is subscribed and running, mock v1 is loaded (not running) 
scenario2

* [a dynamic metric request] load plugin mock v1, start task dynamic-mock-file.yaml, load another-mock
* Expected another-mock is subscribed to by our task
scenario3

Testing done:
- unit tests
- manual

@intelsdi-x/snap-maintainers


control/control.go
```diff
@@ -603,14 +696,36 @@ func (p *pluginControl) ExpandWildcards(ns core.Namespace) ([]core.Namespace, se
...
+	// where only processors/publishers loaded (empty metric catalog)
+	if len(p.metricCatalog.Keys()) > 0 {
+		for _, mt := range mts {
+			errs := p.validateMetricTypeSubscription(mt, mt.Config())
+			if len(errs) > 0 {
+				serrs = append(serrs, errs...)
+			}
+		}
+	}
+
```
IzabellaRaulin: What do you think about move this checkingof the content of metricCatalog (Line 720) into `for` range over `requested` metrics? What is more, I suggest change it to simply checking `if metricCatalog is empty`, and if yes, return an error - because that means that requested metrics cannot be found there.

control/control.go
```diff
@@ -906,23 +982,79 @@ func (p *pluginControl) sendPluginSubscriptionEvent(taskID string, pl core.Plugi
 
 func (p *pluginControl) UnsubscribeDeps(taskID string, mts []core.Metric, plugins []core.Plugin) []serror.SnapError {
```
IzabellaRaulin: Would you mind to add some comments how UnsubscribeDeps works? Based on the name of this function and its arguments I assumed that input mts and plugins will be unsubscribed, what is incorrectly. By the way, thank you @marcin-krolik that you explained me that this function will unsubscribe everything what does not meet with incoming mts and plugins.

scheduler/scheduler.go
```diff
@@ -575,36 +585,39 @@ func (s *scheduler) stopTask(id, source string) []serror.SnapError {
...
+		depGroupMap[""] = depGroup{Metrics: []core.Metric{}}
+		getWorkflowPlugins(t.workflow.processNodes, t.workflow.publishNodes, depGroupMap)
 		var errs []serror.SnapError
 		for k := range depGroupMap {
 			mgr, err := t.RemoteManagers.Get(k)
 			if err != nil {
 				errs = append(errs, serror.New(err))
 			} else {
-				uerrs := mgr.UnsubscribeDeps(t.ID(), depGroupMap[k].Metrics, returnCorePlugin(depGroupMap[k].Plugins))
+				uerrs := mgr.UnsubscribeDeps(t.ID(), []core.Metric{}, returnCorePlugin(depGroupMap[k].Plugins))
```
IzabellaRaulin: Would You mind adding some comment why `[]core.Metric{}` is used instead `depGroupMap[k].Metrics`? Thanks

scheduler/scheduler.go
```diff
@@ -575,36 +585,39 @@ func (s *scheduler) stopTask(id, source string) []serror.SnapError {
 	default:
 		// Group depndencies by the host they live on and
 		// unsubscirbe them since task is stopping.
```
IzabellaRaulin: Would You mind to fix this typos -> "unsubscribe" and "dependencies" - I will really appreciate it.

control/control.go
marcin-krolik: I don't think that it's good idea to return an error for `len(p.metricCatalog.Keys()) == 0`. In distributed workflow control may have no collectors (publishers and/or processors loaded only)

control/control.go
marcin-krolik: Done

scheduler/scheduler.go
marcin-krolik: `depGroupMap[k].Metrics` is not used at all. Metric expansion is no longer done on scheduler side, but in control inside `ValidateDeps` and on plugin load/unload events when metric catalog has a chance to change.

scheduler/scheduler.go
marcin-krolik: Done

control/control.go
IzabellaRaulin: ok, thanks for answer

control/control_test.go
```diff
@@ -1606,10 +1551,25 @@ func (l *listenToPluginSubscriptionEvent) HandleGomitEvent(e gomit.Event) {
...
+		l.done <- struct{}{}
+	default:
+		controlLogger.WithFields(log.Fields{
+			"event:": v.Namespace(),
+			"_block": "HandleGomit",
+		}).Info("Unhandled Event")
+	}
+}
+
+/*
```
lmroz: If it's not in wip stage commented code should be removed (or if this should be kept leave a comment why).

scheduler/scheduler.go
```diff
@@ -848,7 +838,7 @@ func (s *scheduler) walkWorkflow(prnodes []*processNode, pbnodes []*publishNode,
 		} else {
 			depGroupMap[pr.Target] = depGroup{Plugins: []core.SubscribedPlugin{pr}}
 		}
-		s.walkWorkflow(pr.ProcessNodes, pr.PublishNodes, depGroupMap)
+		getWorkflowPlugins(pr.ProcessNodes, pr.PublishNodes, depGroupMap)
```
lmroz: Seems like descendent nodes can be appended to dg.Plugins multiple times, can it cause any problems?

grpc/common/common.go
```diff
@@ -155,6 +155,18 @@ func ToCoreMetric(mt *Metric) core.Metric {
 	return ret
 }
 
+func MetricToRequested(mts []*Metric) []core.RequestedMetric {
+	ret := []core.RequestedMetric{}
```
lmroz: This slice should have length of `len(mts)` to avoid unnecessary reallocs. Similar issue in following function.

grpc/common/common.go
```diff
@@ -178,6 +190,53 @@ func ToCoreMetrics(mts []*Metric) []core.Metric {
...
+func ToRequestedMetrics(mts []*Metric) []core.RequestedMetric {
+	metrics := make([]core.RequestedMetric, len(mts))
+	for i, mt := range mts {
+		metrics[i] = ToCoreMetric(mt)
+	}
+	return metrics
+}
+
+func RequestedToMetric(requested []core.RequestedMetric) []*Metric {
+	reqMets := []*Metric{}
```
lmroz: Same issue as mentioned before, strange in function above it's done correctly :+1:

grpc/common/common.go
```diff
@@ -178,6 +190,53 @@ func ToCoreMetrics(mts []*Metric) []core.Metric {
...
+				Sec:  time.Now().Unix(),
+				Nsec: int64(time.Now().Nanosecond()),
+			},
+		}
+		reqMets = append(reqMets, rm)
+	}
+	return reqMets
+}
+
+// TODO (kromar): implement convertion
```
lmroz: Perhaps you want to resolve this one and following before merging.

lmroz:
I'm not even in the middle of checking the logic, but I wonder what snap behaves on such corner case before and after your PR (does it change anything?):
```
collector+
         |
         +---> moving average processor ---> file publisher
         +---> square processor --> moving average processor ---> file publisher
```
What can go wrong there is routing branches of the flow to same instance of moving avg. processor.
In such case Process can be called with interleaving squared and regular samples giving incorrect result.

jcooklin:
@lmroz I don't think this PR affects this corner case however using config based with the moving average might. 💭

mgmt/rest/client/client_func_test.go
```diff
@@ -160,9 +161,14 @@ func TestSnapClient(t *testing.T) {
 					So(t1.Err.Error(), ShouldEqual, fmt.Sprintf("Task not found: ID(%s)", uuid))
 				})
 				Convey("invalid task (missing metric)", func() {
+					//TODO: load plugin mock:1, create new wf sample with incorrect metric
```
lmroz: TODOs

scheduler/scheduler.go
marcin-krolik: I don't think it may cause any problems. Gathered plugins are used for pool subscriptions which is idempotent operation.

grpc/common/common.go
marcin-krolik: Good catch, will fix it.

grpc/common/common.go
marcin-krolik: Will fix it.

control/control.go
```diff
@@ -90,8 +93,34 @@ type pluginControl struct {
 	grpcServer  *grpc.Server
 	closingChan chan bool
 	wg          sync.WaitGroup
+
+	// taskView mutex
+	view *sync.Mutex
```
lmroz: Perhaps it would be more readable if you used `[variable being guarded]Mutex` as a name.

control/control.go
```diff
@@ -603,17 +696,33 @@ func (p *pluginControl) ExpandWildcards(ns core.Namespace) ([]core.Namespace, se
...
+			if err != nil {
+				// use metric directly from the workflow
+				nss = []core.Namespace{r.Namespace()}
+			}
+
+			for _, ns := range nss {
+				mt := &metric{
+					namespace: ns,
+					version:   r.Version(),
+					config:    configTree.Get(ns.Strings()),
```
lmroz: Is it ok to have null here? I expect such value when namespaces from workflow (because `Match...` returned error) are used.

scheduler/scheduler.go
```diff
@@ -495,19 +500,24 @@ func (s *scheduler) startTask(id, source string) []serror.SnapError {
 		if err != nil {
 			errs = append(errs, serror.New(err))
 		} else {
-			errs = mgr.SubscribeDeps(t.ID(), depGroupMap[k].Metrics, cps)
+			// it is the only place where requested metrics are non-empty (t.workflow.metrics)
+			// on LoadPluginEvent SubscribeDeps will be called with empty list of requested metrics
+			// it is the only place where resulted metrics is empty (on Load/Unload expansion will be done before
+			// SubscribeDeps being called
+			errs = mgr.SubscribeDeps(t.ID(), []core.Metric{}, t.workflow.metrics, cps, t.workflow.configTree)
```
jcooklin: We should consider making the exported `SubscribeDeps` take only requested metrics.  

I'm going to send you some more details on this.

scheduler/scheduler.go
```diff
@@ -573,38 +583,42 @@ func (s *scheduler) stopTask(id, source string) []serror.SnapError {
...
+		getWorkflowPlugins(t.workflow.processNodes, t.workflow.publishNodes, depGroupMap)
 		var errs []serror.SnapError
 		for k := range depGroupMap {
 			mgr, err := t.RemoteManagers.Get(k)
 			if err != nil {
 				errs = append(errs, serror.New(err))
 			} else {
-				uerrs := mgr.UnsubscribeDeps(t.ID(), depGroupMap[k].Metrics, returnCorePlugin(depGroupMap[k].Plugins))
+				// empty metric list is passed to inform UnsubscribeDeps is called from scheduler on stop task
+				uerrs := mgr.UnsubscribeDeps(t.ID(), []core.Metric{}, returnCorePlugin(depGroupMap[k].Plugins))
```
jcooklin: Same as above.   Consider refactoring the exported `UnsubscribeDeps` so it only takes taskid and  []plugins.

control/control.go
```diff
@@ -90,8 +93,34 @@ type pluginControl struct {
...
+	// requests are RequestedMetrics which are "raw" metrics from workflow
+	requested []core.RequestedMetric
+	// results are Metrics expanded from requested
+	results    []core.Metric
+	configTree *cdata.ConfigDataTree
+	// plugins are collectors currently tied to results (expanded metrics)
+	plugins []core.Plugin
+}
+
+type corePlugin struct {
```
IRCody: Having both `core.Plugin` and `corePlugin` be so similar seems like it could be a problem. Is there not a struct already defined in control that implements core.Plugin that could be used?

control/control.go
```diff
@@ -226,6 +262,62 @@ func New(cfg *Config) *pluginControl {
...
+		}
+	case *control_event.UnloadPluginEvent:
+		// while task is running we care about collectors being unloaded
+		// only collectors can affect current workflow by changing metric catalog
+		if core.PluginType(v.Type) == core.CollectorPluginType {
+			for taskID, td := range p.taskView {
+				newMetrics, newPlugins, _ := p.getMetricsAndCollectors(td.requested, td.configTree)
+
+				// pass core.Metrics which are not longer valid
+				// reserve empty slice for task stop discovery
```
IRCody: I don't understand this comment, what empty slice are we reserving?

control/control.go
```diff
@@ -226,6 +262,62 @@ func New(cfg *Config) *pluginControl {
...
+					continue
+				}
+			}
+		}
+	case *control_event.UnloadPluginEvent:
+		// while task is running we care about collectors being unloaded
+		// only collectors can affect current workflow by changing metric catalog
+		if core.PluginType(v.Type) == core.CollectorPluginType {
+			for taskID, td := range p.taskView {
+				newMetrics, newPlugins, _ := p.getMetricsAndCollectors(td.requested, td.configTree)
```
IRCody: Are we swallowing an error here?

control/control.go
```diff
@@ -603,17 +696,33 @@ func (p *pluginControl) ExpandWildcards(ns core.Namespace) ([]core.Namespace, se
...
-			serrs = append(serrs, errs...)
+	var mts []core.Metric
+	// to prevent errors for control in distributed workflow
+	// where only processors/publishers loaded (empty metric catalog)
+	if len(p.metricCatalog.Keys()) > 0 {
+		for _, r := range requested {
+			nss, err := p.MatchQueryToNamespaces(r.Namespace())
+			if err != nil {
+				// use metric directly from the workflow
+				nss = []core.Namespace{r.Namespace()}
```
IRCody: This err is signaling that we were unable to match this query to a namespace (trying to understand what's happening through here).

control/control.go
```diff
@@ -603,17 +696,33 @@ func (p *pluginControl) ExpandWildcards(ns core.Namespace) ([]core.Namespace, se
...
+				mt := &metric{
+					namespace: ns,
+					version:   r.Version(),
+					config:    configTree.Get(ns.Strings()),
+				}
+				errs := p.validateMetricTypeSubscription(mt, mt.Config())
+				if len(errs) > 0 {
+					serrs = append(serrs, errs...)
+				}
+				mts = append(mts, mt)
```
IRCody: Is mts used anywhere? I see where it is appended to but not where it is referenced after that.

control/control.go
```diff
@@ -723,94 +831,57 @@ func (p *pluginControl) validateMetricTypeSubscription(mt core.RequestedMetric,
...
-	}
-	if len(plugins) == 0 {
-		serrs = append(serrs, serror.New(errors.New("No plugins found")))
-		return nil, serrs
-	}
-	return plugins, nil
-}
-
-func (p *pluginControl) SubscribeDeps(taskID string, mts []core.Metric, plugins []core.Plugin) []serror.SnapError {
+func (p *pluginControl) SubscribeDeps(taskID string, mts []core.Metric, requested []core.RequestedMetric, plugins []core.Plugin, configTree *cdata.ConfigDataTree) []serror.SnapError {
```
IRCody: Why core.Metric and requested.Metric, when requested.Metric is a subset of core.Metric? From the call-ees point of view what is the difference when using one vs the other?

control/control.go
```diff
@@ -723,94 +831,57 @@ func (p *pluginControl) validateMetricTypeSubscription(mt core.RequestedMetric,
...
-			serr := p.sendPluginSubscriptionEvent(taskID, gc.plugin)
+			serr := p.sendPluginSubscriptionEvent(taskID, c)
 			if serr != nil {
 				serrs = append(serrs, serr)
 			}
 		}
+	} else {
+		// update task view with current plugins and metrics.
+		// on Load/Unload plugins will only contains collectors
+		p.updateViewOnSubs(taskID, mts, requested, plugins, configTree)
```
IRCody: We don't care about the errors that come from updateViewOnSubs here?

control/control.go
```diff
@@ -822,7 +893,7 @@ func (p *pluginControl) SubscribeDeps(taskID string, mts []core.Metric, plugins
 				serrs = append(serrs, serror.New(err))
 				return serrs
 			}
-			pool.Subscribe(taskID, strategy.UnboundSubscriptionType)
+			pool.Subscribe(taskID)
```
IRCody: Subscribe doesn't contain the subscription type? I noticed in the change log it mentions removing them for simplicity but what is the change in behavior with them removed and how is that accomplished?

control/control.go
```diff
@@ -889,40 +962,95 @@ func (p *pluginControl) sendPluginSubscriptionEvent(taskID string, pl core.Plugi
...
+			pool.Unsubscribe(taskID)
+		}
+		serr := p.sendPluginUnsubscriptionEvent(taskID, sub)
+		if serr != nil {
+			serrs = append(serrs, serr)
 		}
 	}
 
-	for _, sub := range plugins {
+	// / this is the case for task stop (unsubscribe called from scheduler.stopTask)
```
IRCody: Why does it matter where unsubscribe is called from? Is there different behavior based on where it is called from? If so does it make sense for it to be the same function?

control/control.go
```diff
@@ -933,9 +1061,162 @@ func (p *pluginControl) UnsubscribeDeps(taskID string, mts []core.Metric, plugin
 		}
 	}
 
+	td.results = mts
+	td.plugins = plugins
+
+	p.taskView[taskID] = td
+
+	// just debug logging
```
IRCody: Do we want to loop on debug logging even when debug is not set.  Seems like we will eat the cost of these loops even if nothing will happen in them.

control/control.go
```diff
@@ -226,6 +262,62 @@ func New(cfg *Config) *pluginControl {
...
 
+func (p *pluginControl) HandleGomitEvent(e gomit.Event) {
+	switch v := e.Body.(type) {
+	case *control_event.LoadPluginEvent:
+		// while task is running we care about collectors being loaded
+		// only collectors can affect current workflow by changing metric catalog
+		if core.PluginType(v.Type) == core.CollectorPluginType {
+			for taskID, td := range p.taskView {
+				// get all new metrics and collector plugins
+				newMetrics, newPlugins, _ := p.getMetricsAndCollectors(td.requested, td.configTree)
```
IRCody: Is swallowing this error a good idea?

control/control_test.go
IRCody: I agree except the test probably needs to be modified to work with the new changes. We want to test metric subscription rollover.

control/available_plugin.go
```diff
@@ -285,22 +285,15 @@ func (a *availablePlugin) healthCheckFailed() {
...
-		defer a.emitter.Emit(pde)
+	pde := &control_event.DeadAvailablePluginEvent{
+		Name:    a.name,
+		Version: a.version,
+		Type:    int(a.pluginType),
+		Key:     a.key,
+		Id:      a.ID(),
+		String:  a.String(),
 	}
+	defer a.emitter.Emit(pde)
```
IRCody: Do we want to send a DeadAvailablePluginEvent even though we may or may not have passed the health check failure limit?

control/control.go
```diff
@@ -90,8 +90,20 @@ type pluginControl struct {
 	grpcServer  *grpc.Server
 	closingChan chan bool
 	wg          sync.WaitGroup
+
+	subscriptionGroups ManagesSubscriptionGroups
+}
+
+type corePlugin struct {
```
IRCody: Not wild about this name since it conflicts with core.Plugin.

control/control.go
```diff
@@ -723,148 +781,47 @@ func (p *pluginControl) validateMetricTypeSubscription(mt core.RequestedMetric,
...
-			}
-		}
-		serr := p.sendPluginSubscriptionEvent(taskID, sub)
-		if serr != nil {
-			serrs = append(serrs, serr)
-		}
+// UnsubscribeDeps unsubscribes a group of dependencies provided the subscription group ID
+func (p *pluginControl) UnsubscribeDeps(id string) []serror.SnapError {
+	// get all new metrics and collector plugins
+	_, _, _, _, err := p.subscriptionGroups.Get(id)
```
IRCody: What's going on here?

control/control.go
```diff
@@ -883,74 +840,92 @@ func (p *pluginControl) verifyPlugin(lp *loadedPlugin) error {
...
+				"_block": "control",
+				"action": "gather",
+				"metric": fmt.Sprintf("%s:%d", m.Namespace().String(), m.Version()),
+			}).Debug("gathered metrics from workflow request")
+		}
 	}
-	if _, err := p.eventManager.Emit(e); err != nil {
-		return serror.New(err)
+
+	if controlLogger.Level >= log.DebugLevel {
```
IRCody: Can this if {} be rolled into the previous one?

control/control.go
```diff
@@ -1049,13 +1024,21 @@ func (p *pluginControl) MetricExists(mns core.Namespace, ver int) bool {
...
 // returned.
-func (p *pluginControl) CollectMetrics(metricTypes []core.Metric, deadline time.Time, taskID string, allTags map[string]map[string]string) (metrics []core.Metric, errs []error) {
+func (p *pluginControl) CollectMetrics(id string, allTags map[string]map[string]string) (metrics []core.Metric, errs []error) {
 	// If control is not started we don't want tasks to be able to
 	// go through a workflow.
 	if !p.Started {
 		return nil, []error{ErrControllerNotStarted}
 	}
 
+	_, results, _, _, err := p.subscriptionGroups.Get(id)
```
IRCody: The return for subscriptionGroups.Get seems overloaded. If we only ever want a subset of what is returned maybe it should be split out into multiple functions that return what is needed?

control/subscription_group.go
```diff
@@ -0,0 +1,339 @@
...
+type pools interface {
+	getOrCreatePool(key string) (strategy.Pool, error)
+	getPool(key string) (strategy.Pool, serror.SnapError)
+}
+
+type ManagesSubscriptionGroups interface {
+	Process() (errs []serror.SnapError)
+	Add(id string, requested []core.RequestedMetric, metrics []core.Metric,
+		configTree cdata.ConfigDataTree, plugins []core.Plugin) []serror.SnapError
+	Get(id string) ([]core.RequestedMetric, []core.Metric,
```
IRCody: This seems like a lot to return. Will the caller ever want all of it?

control/subscription_group.go
```diff
@@ -0,0 +1,339 @@
...
+	if _, ok := s.subscriptionMap[id]; !ok {
+		return nil, nil, nil, cdata.ConfigDataTree{}, ErrSubscriptionGroupDoesNotExist
+	}
+	sg := s.subscriptionMap[id]
+	return sg.requested, sg.metrics, sg.plugins, sg.configTree, nil
+}
+
+func (s subscriptionGroups) update(id string, metrics []core.Metric, plugins []core.Plugin) error {
+	// s.Lock()
+	// defer s.Unlock()
```
IRCody: Can probably remove this. Seems to be locked by the caller.

control/subscription_group.go
```diff
@@ -0,0 +1,339 @@
...
+	if _, err := p.Emitter.Emit(e); err != nil {
+		return serror.New(err)
+	}
+	return nil
+}
+
+// comparePlugins compares the new state of plugins with the previous state.
+// It returns an array of plugins that need to be subscribed and an array of
+// plugins that need to be unsubscribed.
+func comparePlugins(new, old []core.Plugin) (adds, removes []core.Plugin) {
```
IRCody: Should probably not use `new` since it's a keyword word for the language.

control/subscription_group.go
```diff
@@ -0,0 +1,339 @@
...
+func (s *subscriptionGroups) process() (errs []serror.SnapError) {
+	for id, group := range s.subscriptionMap {
+		metrics, plugins, serrs := s.getMetricsAndCollectors(group.requested, group.configTree)
+		if serrs != nil {
+			for _, err := range errs {
+				errs = append(errs, err)
+			}
+		}
+
+		//add processors and publishers to collectors just gathered
```
IRCody: Needs to have a way to move process/publish plugins to newer/lower versions also.

scheduler/scheduler.go
```diff
@@ -486,7 +488,9 @@ func (s *scheduler) startTask(id, source string) []serror.SnapError {
 	}
 	// Group dependencies by the node they live on
 	// and subscribe to them.
-	depGroupMap := s.gatherMetricsAndPlugins(t.workflow)
+	depGroupMap := make(map[string]depGroup)
+	depGroupMap[""] = depGroup{Metrics: []core.Metric{}}
```
IRCody: is this needed? I'm not sure the Metrics on depgroup is needed anymore.

control/control.go
marcin-krolik: This is wrong. Changing to 
```go
if err != nil {
    serrs = append(serrs, err)
}
```

control/control.go
marcin-krolik: It's leftover from previous implementation where `mts` was gathered to validate it in separate loop. Will remove it.

control/control.go
marcin-krolik: It implements core.Plugin interface, I think naming is OK in this case. What would you suggest instead?

control/control.go
marcin-krolik: Done

control/subscription_group.go
marcin-krolik: Changes to `newPlugins` and `oldPlugins`

scheduler/scheduler.go
marcin-krolik: Metrics are no longer needed. I changed it, so no there's new alias type `type depGroupMap map[string][]core.SubscribedPlugin`

scheduler/scheduler.go
```diff
@@ -73,6 +73,12 @@ const (
 	schedulerStarted
 )
 
+type depGroupMap map[string][]core.SubscribedPlugin
+
+func newDepGroup() depGroupMap {
+	return depGroupMap{"": []core.SubscribedPlugin{}}
```
IRCody: What is the significance of having an empty key with an empty slice?

control/available_plugin.go
IRCody: It still seems weird to emit a DeadAvailablePluginEvent just for missing 1 heartbeat. What is the reasoning behind removing the failedChecks >= healthcheckfailurelimit?

control/control_grpc_server_test.go
```diff
@@ -325,20 +291,24 @@ func TestGRPCServerScheduler(t *testing.T) {
...
+				corePlugin{typeName: "collector", name: "mock", version: 1},
+				corePlugin{typeName: "processor", name: "passthru", version: 1},
+				corePlugin{typeName: "publisher", name: "mock-file", version: 3},
+			},
+			),
+			TaskId:  "my-snowflake-id",
+			CfgTree: common.ToCfgTree(cdata.NewTree()),
+		}
+		_, err := client.SubscribeDeps(context.Background(), req)
+		So(err, ShouldBeNil)
```
IRCody: It looks like the PublishMetrics() call for this block was removed and replaced with subscribeDeps. What is the purpose of subscribe/unsubscribe used in this block that is just testing PublishMetrics.

marcin-krolik: @IRCody `SubscribeDeps()` is needed to subscribe plugins and create needed pool, otherwise `availablePlugins.processMetrics()` will return with `bad key` error. Same thing applies for publishing. 
Removed blocks dealt with invoking process/publish with invalid inputs, which in current implementation makes no sense, because in such case `SubscribeDeps()` will return with error earlier.

control/available_plugin.go
marcin-krolik: @IRCody I agree, while this implementation fixes `TestFailedPlugin` it is incorrect and breaks the logic. I'm going to restore previous implementation. Changes to event listener in test case should fix test behavior.

control/subscription_group.go
marcin-krolik: @IRCody Do we want to have process/publish plugins always in latest version, don't we? How we could tell that one wants to use processor/publisher plugin in lower version? Correct me if I'm wrong but I think that if one really wants to switch to lower version of processor/publisher, one may load it and unloads the latest version. `process()` will find the only available version (in this case newly loaded lower version) and subscribe it. In case when one loads processor/publisher in latest version, `process()` will also automatically discover it and subscribe it.
When there are multiple version of processor/publisher loaded, always the latest one would be used.

IRCody: My understanding is that Snap supports running tasks that potentially access multiple versions of a plugin. You can peg your task to use specific versions of a plugin by specifying the version in the task manifest. If it's not specified it should use the latest. If it is specified it will always use that specific version, regardless of what else is loaded.

jcooklin: @IRCody, correct and this is what is currently implemented in `func (s *subscriptionGroups) process() (errs []serror.SnapError)`.  

@marcin-krolik , I just pushed a small commit that adds comment for documentation and renames the variable `latest`, which is a misnomer.

control/control_grpc_server_test.go
IRCody: >@IRCody SubscribeDeps() is needed to subscribe plugins and create pools, otherwise availablePlugins.processMetrics() will return with bad key error. Same thing applies for publishing. 

Ok I see your point. This is because collect/process/publish is given less info now than it used to be correct?

>Removed blocks dealt with invoking process/publish with invalid inputs, which in current implementation makes no sense, because in such case SubscribeDeps() will return with error earlier.

It seems like there will still be invalid inputs for collect/process/publish in this implementation -- at the very least invalid id's.

marcin-krolik: > Ok I see your point. This is because collect/process/publish is given less info now than it used to be correct?

True

>It seems like there will still be invalid inputs for collect/process/publish in this implementation -- at the very least invalid id's.

I will add this scenario in test case

mgmt/tribe/delegate.go
```diff
@@ -117,7 +117,9 @@ func (t *delegate) NotifyMsg(buf []byte) {
 		if err := decodeMessage(buf[1:], msg); err != nil {
 			panic(err)
 		}
+		t.tribe.mutex.RLock()
 		queryResp, ok := t.tribe.taskStateResponses[msg.UUID]
+		t.tribe.mutex.RUnlock()
```
IRCody: 👍

control/control_grpc_server.go
```diff
@@ -94,14 +95,14 @@ func (pc *ControlGRPCServer) CollectMetrics(ctx context.Context, r *rpc.CollectM
...
-	configTree := common.ToConfigDataTree(r.CfgTree)
+	configTree := cdata.NewTree()
 	serrors := pc.control.ValidateDeps(metrics, plugins, configTree)
 	return &rpc.ValidateDepsReply{Errors: common.NewErrors(serrors)}, nil
 }
 
 func (pc *ControlGRPCServer) SubscribeDeps(ctx context.Context, r *rpc.SubscribeDepsRequest) (*rpc.SubscribeDepsReply, error) {
 	plugins := common.MsgToCorePlugins(r.Plugins)
-	configTree := common.ToConfigDataTree(r.CfgTree)
+	configTree := cdata.NewTree()
```
IRCody: Maybe leave a comment here about why it's ok to pass in empty tree (that the configTree is only used on collection and remote collection is not currently supported).

control/available_plugin.go
```diff
@@ -457,6 +458,10 @@ func (ap *availablePlugins) publishMetrics(contentType string, content []byte, p
 		return []error{serror.New(ErrPoolNotFound, map[string]interface{}{"pool-key": key})}
 	}
 
+	if !pool.HasSubscription(taskID) {
+		return []error{serror.New(ErrBadTaskId, map[string]interface{}{"task-id": taskID})}
+	}
+
```
IRCody: How was the case where a plugin is started and added to a pool but not subscribed to handled previously? I think the subscription is what kicks off adding a plugin to a pool. I'd be curious what sequence of events would have you end up with a pool for a plugin that doesn't have a valid subscription to it?

control/control.go
IRCody: I also wonder if Match returns an error is it still valid to loop over nss?

marcin-krolik: @IRCody it doesn't make sense, added proper logging and continue with loop

@lmroz added init in case of config tree does not have configuration for current namespace

control/available_plugin.go
marcin-krolik: @IRCody after thinking about I can't find any scenario which would lead apart from artificially pass incorrect task id in unit tests. I will remove the commit  which introduced it

control/control_grpc_server_test.go
```diff
@@ -101,13 +101,13 @@ func TestGRPCServerScheduler(t *testing.T) {
...
 	}
-	c.Load(passthru)
+	catalogedPassthru, _ := c.Load(passthru)
 	<-lpe.done
 	filepub, err := core.NewRequestedPlugin(path.Join(fixtures.SnapPath, "plugin", "snap-plugin-publisher-mock-file"))
 	if err != nil {
 		log.Fatal(err)
 	}
-	c.Load(filepub)
+	catalogedFile, _ := c.Load(filepub)
```
IRCody: Should we grab/process the errors here just in case the `Load()` fails for some reason?

control/control_test.go
```diff
@@ -1599,8 +1604,9 @@ func TestMetricSubscriptionToOlderVersion(t *testing.T) {
 
 func TestDynamicMetricSubscriptionLoad(t *testing.T) {
 	Convey("Given a dynamic metric that is being collected", t, func() {
+		log.SetLevel(log.DebugLevel)
```
IRCody: Is this meant to be left here?

control/subscription_group.go
```diff
@@ -197,129 +184,262 @@ func (s subscriptionGroups) update(id string, metrics []core.Metric,
...
+		m.config = s.Config.Plugins.getPluginConfigDataNode(typ,
+			m.Plugin.Name(), m.Plugin.Version())
+	}
+
+	// When a metric is added to the MetricCatalog, the policy of rules defined by the plugin is added to the metric's policy.
+	// If no rules are defined for a metric, we set the metric's policy to an empty ConfigPolicyNode.
+	// Checking m.policy for nil will not work, we need to check if rules are nil.
+	if m.policy.HasRules() {
+		if m.Config() == nil {
+			// TODO (JC) fixup error
```
IRCody: Is this meant to be done prior to merging?

scheduler/scheduler.go
```diff
@@ -502,12 +502,12 @@ func (s *scheduler) startTask(id, source string) []serror.SnapError {
 	var subbedDeps []string
 	for k := range depGroups {
 		var errs []serror.SnapError
-		cps := returnCorePlugin(depGroups[k].subscribedPlugins)
+		// cps := returnCorePlugin(depGroups[k].subscribedPlugins)
```
IRCody: ?

docs/PLUGIN_LIFECYCLE.md
```diff
@@ -0,0 +1,86 @@
...
+    * `GetMetricTypes` returns the metrics the plugin provides
+    * `GetConfigPolicy` returns the conf policy that the plugin needs
+3. The plugin is stopped
+
+It should be emphasized that when a plugin is loaded it is started but stopped 
+as soon as the metric catalog has been updated.  
+
+## What happens when a task is started
+
+When a task is started the plugins that it references are started and subscribed
```
bjray: I'd suggest replacing "it" with "the task".

docs/PLUGIN_LIFECYCLE.md
```diff
@@ -0,0 +1,86 @@
...
+    * The schedule is validated
+    * The config provided for the metrics (collectors), processors and 
+    publishers are validated
+2. On **task starting** the plugins are started (`snapctl task start <TASK_ID>`)
+3. Subscriptions for each plugin referenced by the task are incremented 
+
+## Diving deeper
+
+**Task started** - When a task is started the plugins which are referenced by 
+the workflow are subscribed to and a 'subscription group' is created.  A
```
bjray: first use of the term "workflow" might be confusing here.  I'd try to be more specific if possible.  Maybe the "task manifest" or "workflow manifest" or are you referring to "workflow" in a more general case?

control/subscription_group_medium_test.go
```diff
@@ -0,0 +1,671 @@
...
+				anotherMock := mockSubscribedPlugin{
+					typeName: core.CollectorPluginType,
+					name:     "anothermock",
+					version:  1,
+					config:   cdata.NewNode(),
+				}
+
+				_, err := loadPlg(c, path.Join(os.Getenv("SNAP_PATH"), "plugin", "snap-plugin-collector-anothermock1"))
+				<-lpe.load
+				So(err, ShouldBeNil)
```
IRCody: Should we Check the err here before waiting on the event? If it errors we won't ever receive the event (I think).

control/subscription_group_medium_test.go
```diff
@@ -0,0 +1,671 @@
...
+	c := New(getTestSGConfig())
+
+	lpe := newLstnToPluginEvents()
+	c.eventManager.RegisterHandler("TestSubscriptionGroups_Process", lpe)
+	c.Start()
+
+	Convey("Loading a mock collector plugn", t, func() {
+		_, err := loadPlg(c, path.Join(os.Getenv("SNAP_PATH"), "plugin", "snap-plugin-collector-mock1"))
+		<-lpe.load
+		So(err, ShouldBeNil)
```
IRCody: Same comment here.

IRCody:
LGTM

IRCody:
Rebase LGTM merging.

PR #1111, opened 2016-07-29
WIP: add scripts to create deb and rpm packages

### LINKED ISSUE #774: Packaging for Centos 7
It would be super handy to have rpms (with idomatic structure: docs in man pages or /usr/share, configuration in /etc, systemd units, daemon output ending in syslog, etc). instead of having to build or pull in the bare binary.

woodsaj:
Fixes #774

Summary of changes:
- Add script to build deb packages using fpm.  Builds for Ubuntu14.04(Trusty), and for Ubuntu16.04(Xenial) and Debian 8 (Jessie).  Packages are build by calling `make deb`
- Add script to build rpm packages using fpm. Builds for Centos6 and Centos7. packages are built by calling `make rpm` 

Currently the packages include the snapd config from the examples dir, which leads to snapd failing to start properly, so this should probably be changed to a minimal config instead.

Also, when testing this on Ubuntu16.04, there is already a package called snapd. So the packages need a different name.  I am open to suggestions.

@intelsdi-x/snap-maintainers


mjbrender:
Great to see your work on this @woodsaj. I'd like to connect you with @nanliu who's built tooling around packages. Also, have you seen this page? http://snap-telemetry.io/download.html

nanliu:
@woodsaj, I added access to https://github.com/intelsdi-x/snap_packaging. Let me know if you have any feedback. Also we are building a linux snap binary for every merge on master now at:

* http://snap.ci.snap-telemetry.io/snap/master/{git_sha1}/snapd
* http://snap.ci.snap-telemetry.io/snap/master/{git_sha1}/snapctl

The latest build on master is available at:

* http://snap.ci.snap-telemetry.io/snap/master/latest/snapd
* http://snap.ci.snap-telemetry.io/snap/master/latest/snapctl

If there are interest, I can push the release version here as well.

scripts/initscripts/snapd.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
```
nanliu: This service name conflicts with a package installed by default in Ubuntu 16.04.1, so we are going to change it. See: https://github.com/intelsdi-x/snap_packaging/issues/13

scripts/initscripts/snapd.service
```diff
@@ -0,0 +1,18 @@
...
+Description=Starts and stops a single snapd instance on this system
+Documentation=https://github.com/intelsdi-x/snap
+Wants=network-online.target
+After=network-online.target
```
nanliu: Currently my systemd script is simply After=network.target. Should I switch to After=network-online.target?

scripts/initscripts/snapd.service
```diff
@@ -0,0 +1,18 @@
...
+
+[Service]
+User=root
+Group=root
+Type=simple
+Restart=on-failure
+ExecStart=/usr/local/bin/snapd --config=/etc/snapd/snapd.yaml
+LimitNOFILE=102400
+TimeoutStopSec=60
```
nanliu: I will add the Limit and TimoutStop since these seem like good ideas.

scripts/initscripts/snapd.service
```diff
@@ -0,0 +1,18 @@
...
+Documentation=https://github.com/intelsdi-x/snap
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+User=root
+Group=root
+Type=simple
+Restart=on-failure
+ExecStart=/usr/local/bin/snapd --config=/etc/snapd/snapd.yaml
```
nanliu: The default is the .conf file, any preference for using yaml? I can see if we can make it an option in the service script.

nanliu:
I'm going to incorporate some of this PR to our package scripts:
https://github.com/intelsdi-x/snap_packaging/pull/14

Would love to get your feedback.

scripts/initscripts/snapd.service
woodsaj: network-online.target seems more appropriate. If users have snapd configured to listen on a specific IP, we cant start snapd until that IP is assigned to an interface. And in the case of DHCP, that wont be until the network is up.

scripts/initscripts/snapd.service
woodsaj: Default should be fine.  Though as the config file, and other settings can be passed via Environment variables it may be a good idea to use an "EnvironmentFile" in the service definition and have that set SNAP_CONFIG_PATH

EnvironmentFile should probably be /etc/default/snap-telemetry or /etc/sysconfig/snap-telemetry

nanliu:
I've incorporated all the feedback to snap_packaging including EnvironmentFile. Let us know if you have any other suggestions. These changes are expected in snap 0.16 package release.

nanliu:
Thanks for the PR. I'm going to close this since we merged all the suggested changes to our package repo. If you have any other request related to packages let me know.

PR #1094, opened 2016-07-25
make: Change make install to build snap

ssorsa:
Fixes #XXXX

Summary of changes:
- change `make install` to include building of snap before `cp` 

Testing done:
-  Tested manually

@intelsdi-x/snap-maintainers

Change `make install` also to build snap before `cp`, by adding the command for
executing `make all`. Tested manually.

Signed-off-by: Saija Sorsa <saija.sorsa@intel.com>

PR #1119, opened 2016-08-01
make: Change 'make install' NOT to build Snap. Revert of #1094

### LINKED ISSUE #1108: New make install doesn't work if you don't have GOPATH set up for root.
There is a possibility that following the steps from "Build and Test" tutorial (https://github.com/intelsdi-x/snap/blob/master/docs/BUILD_AND_TEST.md) a user could omit fact that they need to export GOPATH variable for root as well. And if they do omit this detail, snap will not compile with the command "sudo make install" - in the latest commit `make install` causes not only `cp` binaries to /user/local/bin, but also before that executes `make` for which GOPATH is needed.

**What do you think about adding an annotation in this tutorial that GOPATH should be set both for user and root?** This issue will probably reveal itself to the users that start their journey with Snap. Annotation will help us not to spoil the first impression with compilation error.

mkuculyma:
Fixes #1108: 

Summary of changes:
~~Small update for BUILD_AND_TEST.md tutorial.~~
~~After changing 'make install' to also compile Snap source, the GOPATH environment variable needs to set up for root as well. Otherwise one gets a compilation error.~~
- Changes in Makefile: remove `make all` from `make install`

@intelsdi-x/snap-maintainers

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,6 @@ check:
 all:
 	bash -c "./scripts/build_snap.sh"
 install:
-	$(MAKE) all
 	cp build/bin/snapd /usr/local/bin/
 	cp build/bin/snapctl /usr/local/bin/
 proto:


mjbrender:
Hey @mkuculyma, thank you for catching this missing piece of the configuration! Upon review, I'm left with a few questions: 

* Do I need to have $GOPATH set as root on certain OSes or all of them? 
* How do you verify that t's setup properly before running `make`? 

The clearer you can make this page, the more I will thank you with Snappy stickers 🐢 🐢 🐢 🐢 🐢

nanliu:
We currently [check go_path](https://github.com/intelsdi-x/snap/blob/master/scripts/common.sh#L69-L79) and raise an error if it's not set.  I think the one scenario where GOPATH might be stripped is `sudo make install` and the sudo config:

```
Defaults env_reset
```

This stems from merge #1094. In that PR, I mentioned make & make install are typically two steps, because as the user compiling snap you need $GOPATH, but should not be required for copying the binary which is typically the only task for `make install`.

I would even go as far to say we should revert #1094 to allow this to be different steps and we no longer have this requirement (and split the task to compile as user, install as root):

```
GOPATH=... 
make
sudo make install
```

With #1094 , make install as root require the GOPATH configured correctly for root, and probably won't be very happy with GVM (and all the bash path munging).

geauxvirtual:
I tend to agree with the discussion earlier today that `make` and `make install` should be two separate commands.

tjmcs:
I know I'm kind of going back on myself here, but after the merge I've found things confusing (to say the least) with `make install` also running `make`. I think this comes down to the fact that we aren't really using the `Makefile` to drive the `make` and `make install` commands on it's own, instead that `Makefile` simply runs some shell scripts that drive the `go build ...` commands that are used to actually build Snap.

The reason I find this so confusing is that the following command (now that the change was merged and the `make install` command runs `make`) is that the following command is a perfectly legitimate command for building packages under a `bash` login shell (and I use it all the time):
```bash
$ make && make install
...
```
but when I run that command now to build Snap it actually builds Snap, then builds it a second time, and then installs it. If this were driven purely out of the `Makefile` it probably wouldn't be an issue (since the `Makefile` would know when you need to rebuild something and when you don't), but our shell scripts aren't nearly that smart. So after playing with this for a week I'm voting with @geauxvirtual ; `make` and `make install` should be separate commands.

jcooklin:
@mkuculyma will you revert #1094 as part of this PR?

IRCody:
This PR doesn't seem to be updating BUILD_AND_TEST.md anymore. Is that intentional? Is there other documentation changes that are going into this PR or is it just the change for `make install` to no longer run `make all` first?

mkuculyma:
The subject of this PR has changed. It was supposed to be additional note in documentation but it has been decided to make a change in Makefile instead. So documentation does not need updates anymore. It is intentional to make no changes in BUILD_AND_TEST.md. The only change is `make install` to no longer run `make all`. I've edited the title so that it reflects the actual changes.

PR #1127, opened 2016-08-05
Fixes #1113 & #1131 and Resolves #1126; improves task creation process in `snapctl`

### LINKED ISSUE #285: Name in task manifest
The name of the task should be an option in the task manifest and supported form a snapctl load.

### LINKED ISSUE #1113: Task disabled when max-failures set to -1

### LINKED ISSUE #1126: Command-line parameters should be consistently applied when creating tasks
When creating tasks, the current version of `snapctl` only uses the command-line parameters associated with the `snapctl task create ...` command when a task is being created using a workflow manifest. In the case where a task is being created using a task manifest, any command-line parameters that are included in the `snapctl task create ...` command are silently ignored.

It would be much less confusing to the user if these command-line parameters were consistently applied for both of these use cases (creating a task using either a task manifest or a workflow manifest), with any command-line parameters that are set by the user when creating a task using a task manifest overriding the corresponding definitions for these parameters in the task manifest.

### LINKED ISSUE #1131: snapctl printing "If the number of maximum failures is not specified, use default value of x"
When starting a task without the maximum number of failures, snapctl will print out the following:
```
% ./snapctl task create -t task.json  
If the number of maximum failures is not specified, use default value of 10
```
this feels like an odd place to be printing this out and is phrased somewhat oddly. If we want to inform the user of what we are doing it could be phrased like:
```
No value provided for maximum failures, using default value of 10
```
To more clearly state what is happening, rather than putting it as a conditional (which is only printed in the case where the # of max failures is not specified).

Relevant code is [here](https://github.com/intelsdi-x/snap/blob/a9436084ae0c803b00e1452cf104db1b917692c1/cmd/snapctl/task.go#L134).

tjmcs:
Fixes #1113 
Fixes #1131 
Fixes #285 
Resolves #1126

**Summary of changes**

There are two sets of changes in this pull request. The first, which fix Issue #1113 are fairly simple and don't change the behavior of the system at all. They can be summarized as follows:
- Removed a line of code from the `core/task.go` file that used to set the `max-failure` value to `10` for all tasks, replacing it with a conditional that set the underlying option only in those cases where a `MaxFailures` value was included as part of the task creation request.
- Modified the `scheduler/task.go` file so that the value of the `DefaultStopOnFailure` constant defined in that file is set to `10` (making it consistent with the behavior that used to be defined elsewhere but was inconsistently set to a value of `3` in this file)
- Removed the definitions of a similar constant (the `DefaultMaxFailures` constant) from the `cmd/snapctl/flags.go` and `mgmt/rest/flags.go` files
- Modified the command-line flag used to set the `max-failures` parameter so that it is a string flag rather than an integer flag and added code to parse this string value as an integer (and throw an error if the value passed in through this command-line flag cannot be parsed as an integer). This removes the string `(default: 0)` from the help output for the `snapctl task create ...` command.

The second set of changes were made to resolve Issue #1126. This set of changes *does* modify the behavior of the system, changing the `snapctl` command so that it honors any command-line flags that the user might have set, regardless of whether the user is creating a task using a task manifest or a workflow manifest. In the case where the user is creating a task using a task manifest and has included (new) definitions for parameters in appear that task manifest as command-line arguments, those command-line argument values will take precedence over the values defined in the task manifest. The changes in this second commit can best be summarized as follows:
- Added methods to the `task` interface that can be used to parse any command-line parameters that might have been set by the user so that those command-line parameter values can be used when creating a new task. In the case where a workflow manifest is being used, those values will be used to set the necessary values for the underlying task. When a task manifest is being used those values will override the corresponding values from the task manifest file.
- Added code to check the boundaries of a `windowed` task schedule (if one is defined in the task manifest and/or on the command-line) to ensure that the time window defined is valid.
- Modified the functions used to create a task using either a task manifest or a workflow manifest so that they both call into our new methods to apply the command-line parameters defined by the user when creating the task.

**Note**; this pull request is a replacement for PR #1120, which fixed and resolved the same issues but in a single commit. This PR breaks out the fix into two separate commits, making it much clearer what is being changed and how.

**Testing done**

Tested the first set of changes manually by running two tests using two different task manifests:
- For the first test, created a task using a task manifest that had an undefined value for the `max-failures` parameter, then unloading the collector plugin(s) that the task depended on and ensuring that the task was disabled after `10` failures (the default)
- For the second test, created a task using a task manifest that had a value set for the `max-failures` parameter (the value we chose was `3`, but any value greater than zero would do for this test), then unloading the collector plugin(s) that the task depended on and ensuring that the task was disabled after the chosen number of failures (in our case, `3`).

The second set of changes were also tested manually by running a single test:
- A task was created using the second a task manifest from our previous tests (which assigned a value of `3` to the `max-failures` parameter and a value of `1s` for the `interval`), but passing in different values on the command-line for these two parameters (we chose values of `5` and `750ms`, respectively, for our test).
- For this test, we first ensured that our task was running at a shorter interval than the interval defined in the task manifest (it was, with the interval between task runs matching the `750ms` value passed in on the command line; four times every three seconds or 80 times a minute).
- We then unloaded the collector plugin(s) that the task depended on, and verified that the task was disabled after `5` failures rather than `3`.

Finally, the current set of legacy tests was run to ensure that we hadn't changed the behavior of the system from the previous behavior, we'd only added the ability to override some of the parameters defined in the task manifest using values defined on the command-line.

@intelsdi-x/snap-maintainers

### DIFF
--- a/cmd/snapctl/task.go
+++ b/cmd/snapctl/task.go
@@ -101,8 +101,8 @@ func createTask(ctx *cli.Context) error {
 }
 
 func stringValToInt(val string) (int, error) {
-	// parse the input (string) value as an integer (log a Fatal
-	// error if we cannot parse the value as an integer)
+	// parse the input (string) as an integer value (and return that integer value
+	// to the caller or an error if the input value cannot be parsed as an integer)
 	parsedField, err := strconv.Atoi(val)
 	if err != nil {
 		splitErr := strings.Split(err.Error(), ": ")
@@ -114,14 +114,219 @@ func stringValToInt(val string) (int, error) {
 	return parsedField, nil
 }
 
+// Parses the command-line parameters (if any) and uses them to override the underlying
+// schedule for this task or to set a schedule for that task (if one is not already defined,
+// as is the case when we're building a new task from a workflow manifest).
+//
+// Note: in this method, any of the following types of time windows can be specified:
+//
+//    +---------------------------...  (start with no stop and no duration; no end time for window)
+//    |
+//  start
+//
+//    ...---------------------------+  (stop with no start and no duration; no start time for window)
+//                                  |
+//                                 stop
+//
+//    +-----------------------------+  (start with a duration but no stop)
+//    |                             |
+//    |---------------------------->|
+//  start        duration
+//
+//    +-----------------------------+  (stop with a duration but no start)
+//    |                             |
+//    |<----------------------------|
+//               duration         stop
+//
+//    +-----------------------------+ (start and stop both specified)
+//    |                             |
+//    |<--------------------------->|
+//  start                         stop
+//
+//    +-----------------------------+ (only duration specified, implies start is the current time)
+//    |                             |
+//    |---------------------------->|
+//  Now()        duration
+//
+func (t *task) setWindowedSchedule(start *time.Time, stop *time.Time, duration *time.Duration) error {
+	// if there is an empty schedule already defined for this task, then set the
+	// type for that schedule to 'windowed'
+	if t.Schedule.Type == "" {
+		t.Schedule.Type = "windowed"
+	} else if t.Schedule.Type != "windowed" {
+		// else if the task's existing schedule is not a 'windowed' schedule,
+		// then return an error
+		return fmt.Errorf("Usage error (schedule type mismatch); cannot replace existing schedule of type '%v' with a new, 'windowed' schedule", t.Schedule.Type)
+	}
+	// if a duration was passed in, determine the start and stop times for our new
+	// 'windowed' schedule from the input parameters
+	if duration != nil {
+		// if start and stop were both defined, then return an error (since specifying the
+		// start, stop, *and* duration for a 'windowed' schedule is not allowed)
+		if start != nil && stop != nil {
+			return fmt.Errorf("Usage error (too many parameters); the window start, stop, and duration cannot all be specified for a 'windowed' schedule")
+		}
+		// if start is set and stop is not then use duration to create stop
+		if start != nil && stop == nil {
+			newStop := start.Add(*duration)
+			t.Schedule.StartTime = start
+			t.Schedule.StopTime = &newStop
+			return nil
+		}
+		// if stop is set and start is not then use duration to create start
+		if stop != nil && start == nil {
+			newStart := stop.Add(*duration * -1)
+			t.Schedule.StartTime = &newStart
+			t.Schedule.StopTime = stop
+			return nil
+		}
+		// otherwise, the start and stop are both undefined but a duration was passed in,
+		// so use the current date/time (plus the 'createTaskNowPad' value) as the
+		// start date/time and construct a stop date/time from that start date/time
+		// and the duration
+		newStart := time.Now().Add(createTaskNowPad)
+		newStop := newStart.Add(*duration)
+		t.Schedule.StartTime = &newStart
+		t.Schedule.StopTime = &newStop
+		return nil
+	}
+	// if a start date/time was specified, we will use it to replace
+	// the current schedule's start date/time
+	if start != nil {
+		t.Schedule.StartTime = start
+	}
+	// if a stop date/time was specified, we will use it to replace the
+	// current schedule's stop date/time
+	if stop != nil {
+		t.Schedule.StopTime = stop
+	}
+	// if we get this far, then just return a nil error (indicating success)
+	return nil
+}
+
+// parse the command-line options and use them to setup a new schedule for this task
+func (t *task) setScheduleFromCliOptions(ctx *cli.Context) error {
+	// check the start, stop, and duration values to see if we're looking at a windowed schedule (or not)
+	// first, get the parameters that define the windowed schedule
+	start := mergeDateTime(
+		strings.ToUpper(ctx.String("start-time")),
+		strings.ToUpper(ctx.String("start-date")),
+	)
+	stop := mergeDateTime(
+		strings.ToUpper(ctx.String("stop-time")),
+		strings.ToUpper(ctx.String("stop-date")),
+	)
+	// Grab the duration string (if one was passed in) and parse it
+	durationStr := ctx.String("duration")
+	var duration *time.Duration
+	if ctx.IsSet("duration") || durationStr != "" {
+		d, err := time.ParseDuration(durationStr)
+		if err != nil {
+			return fmt.Errorf("Usage error (bad duration format); %v", err)
+		}
+		duration = &d
+	}
+	// Grab the interval for the schedule (if one was provided). Note that if an
+	// interval value was not passed in and there is no interval defined for the
+	// schedule associated with this task, it's an error
+	interval := ctx.String("interval")
+	if !ctx.IsSet("interval") && interval == "" && t.Schedule.Interval == "" {
+		return fmt.Errorf("Usage error (missing interval value); when constructing a new task schedule an interval must be provided")
+	}
+	// if a start, stop, or duration value was provided, or if the existing schedule for this task
+	// is 'windowed', then it's a 'windowed' schedule
+	isWindowed := (start != nil || stop != nil || duration != nil || t.Schedule.Type == "windowed")
+	// if an interval was passed in, then attempt to parse it (first as a duration,
+	// then as the definition of a cron job)
+	isCron := false
+	if interval != "" {
+		// first try to parse it as a duration
+		_, err := time.ParseDuration(interval)
+		if err != nil {
+			// if that didn't work, then try parsing the interval as cron job entry
+			_, e := cron.Parse(interval)
+			if e != nil {
+				return fmt.Errorf("Usage error (bad interval value): cannot parse interval value '%v' either as a duration or a cron entry", interval)
+			}
+			// if it's a 'windowed' schedule, then return an error (we can't use a
+			// cron entry interval with a 'windowed' schedule)
+			if isWindowed {
+				return fmt.Errorf("Usage error; cannot use a cron entry ('%v') as the interval for a 'windowed' schedule", interval)
+			}
+			isCron = true
+		}
+		t.Schedule.Interval = interval
+	}
+	// if it's a 'windowed' schedule, then create a new 'windowed' schedule and add it to
+	// the current task; the existing schedule (if on exists) will be replaced by the new
+	// schedule in this method (note that it is an error to try to replace an existing
+	// schedule with a new schedule of a different type, so an error will be returned from
+	// this method call if that is the case)
+	if isWindowed {
+		return t.setWindowedSchedule(start, stop, duration)
+	}
+	// if it's not a 'windowed' schedule, then set the schedule type based on the 'isCron' flag,
+	// which was set above.
+	if isCron {
+		// make sure the current schedule type (if there is one) matches; if not it is an error
+		if t.Schedule.Type != "" && t.Schedule.Type != "cron" {
+			return fmt.Errorf("Usage error; cannot replace existing schedule of type '%v' with a new, 'cron' schedule", t.Schedule.Type)
+		}
+		t.Schedule.Type = "cron"
+		return nil
+	}
+	// if it wasn't a 'windowed' schedule and it's not a 'cron' schedule, then it must be a 'simple'
+	// schedule, so first make sure the current schedule type (if there is one) matches; if not
+	// then it's an error
+	if t.Schedule.Type != "" && t.Schedule.Type != "simple" {
+		return fmt.Errorf("Usage error; cannot replace existing schedule of type '%v' with a new, 'simple' schedule", t.Schedule.Type)
+	}
+	// if we get this far set the schedule type and return a nil error (indicating success)
+	t.Schedule.Type = "simple"
+	return nil
+}
+
+// merge the command-line options into the current task
+func (t *task) mergeCliOptions(ctx *cli.Context) error {
+	// set the name of the task (if a 'name' was provided in the CLI options)
+	name := ctx.String("name")
+	if ctx.IsSet("name") || name != "" {
+		t.Name = name
+	}
+	// set the deadline of the task (if a 'deadline' was provided in the CLI options)
+	deadline := ctx.String("deadline")
+	if ctx.IsSet("deadline") || deadline != "" {
+		t.Deadline = deadline
+	}
+	// set the MaxFailures for the task (if a 'max-failures' value was provided in the CLI options)
+	maxFailuresStrVal := ctx.String("max-failures")
+	if ctx.IsSet("max-failures") || maxFailuresStrVal != "" {
+		maxFailures, err := stringValToInt(maxFailuresStrVal)
+		if err != nil {
+			return err
+		}
+		t.MaxFailures = maxFailures
+	}
+	// shouldn't ever happen, but...
+	if t.Version != 1 {
+		return fmt.Errorf("Invalid version provided while creating task")
+	}
+	// set the schedule for the task from the CLI options (and return the results
+	// of that method call, indicating whether or not an error was encountered while
+	// setting up that schedule)
+	return t.setScheduleFromCliOptions(ctx)
+}
+
 func createTaskUsingTaskManifest(ctx *cli.Context) error {
+	// get the task manifest file to use
 	path := ctx.String("task-manifest")
 	ext := filepath.Ext(path)
 	file, e := ioutil.ReadFile(path)
 	if e != nil {
-		return fmt.Errorf("File error [%s]- %v\n", ext, e)
+		return fmt.Errorf("File error [%s] - %v\n", ext, e)
 	}
 
+	// create an empty task struct and unmarshal the contents of the file into that object
 	t := task{}
 	switch ext {
 	case ".yaml", ".yml":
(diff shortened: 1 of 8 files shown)


cmd/snapctl/task.go
```diff
@@ -114,14 +114,226 @@ func stringValToInt(val string) (int, error) {
 	return int(parsedField), nil
 }
 
+func (t *task) checkScheduleWindow(start time.Time, stop time.Time) error {
+	// if the stop time defined by the user is in the past, then throw
+	// an error (at least some of the defined window should be in the future)
+	if time.Now().Add(createTaskNowPad).After(stop) {
+		return fmt.Errorf("Usage error (poorly defined time window); when constructing a new 'windowed' schedule, part of that window must be in the future")
+	} else if start.After(stop) {
```
geauxvirtual: I would remove the `else if` and just `if` as each statement will return from the function with an error if the `if` statement is true.

cmd/snapctl/task.go
```diff
@@ -162,18 +376,21 @@ func createTaskUsingTaskManifest(ctx *cli.Context) error {
...
 func createTaskUsingWFManifest(ctx *cli.Context) error {
-	// Get the workflow
+	// Get the workflow manifest filename from the command-line
 	path := ctx.String("workflow-manifest")
 	ext := filepath.Ext(path)
 	file, e := ioutil.ReadFile(path)
 
-	if !ctx.IsSet("interval") && !ctx.IsSet("i") {
-		return fmt.Errorf("Workflow manifest requires interval to be set via flag.")
+	// check to make sure that an interval was specified using the appropriate command-line flag
```
geauxvirtual: This block of code (384-387) should be placed after the error check on opening the file.

cmd/snapctl/task.go
```diff
@@ -114,14 +114,226 @@ func stringValToInt(val string) (int, error) {
...
+	interval := ctx.String("interval")
+	if !ctx.IsSet("interval") && interval == "" && t.Schedule.Interval == "" {
+		return fmt.Errorf("Usage error (missing interval value); when constructing a new task schedule, an interval must be provided")
+	}
+	// if a start, stop, or duration value was provided, then the CLI options are being used to
+	// specify a new 'windowed' schedule for this task, so use them to replace the 'windowed' schedule
+	// in the current task (if one exists); note that it is an error to try to replace an existing
+	// sechedule with a new schedule of a different type, so throw an error if that is the case
+	if start != nil || stop != nil || duration != nil {
+		t.setWindowedSchedule(start, stop, duration)
```
geauxvirtual: `setWindowedSchedule` is returning an error, but that is not being captured here. This should return that error `return t.setWindowedSchedule(start, stop, duration), and remove the `else if`.

cmd/snapctl/task.go
```diff
@@ -114,14 +114,226 @@ func stringValToInt(val string) (int, error) {
...
+	// return a nil error (indicating success)
+	return nil
+}
+
+// Parses the command-line parameters and uses them to override the underlying schedule
+// in this task (if any) or set a schedule for that task (if one is not already defined)
+func (t *task) setWindowedSchedule(start *time.Time, stop *time.Time, duration *time.Duration) error {
+	// if this task does not have an existing schedule, create an empty
+	// windowed schedule
+	if t.Schedule == nil {
```
geauxvirtual: t.Schedule should never be nil here. A task manifest should always include a schedule, and if it doesn't, an error should be thrown long before the code gets down this path.

cmd/snapctl/task.go
```diff
@@ -114,14 +114,226 @@ func stringValToInt(val string) (int, error) {
...
+	t.Schedule.StartTime = start
+	t.Schedule.StopTime = stop
+	// and return a nil error (indicating success)
+	return nil
+}
+
+// parse the command-line options and use them to setup a new schedule for this task
+func (t *task) setScheduleFromCliOptions(ctx *cli.Context) error {
+	// if there is no schedule associated with this task, create an empty one
+	if t.Schedule == nil {
```
geauxvirtual: This block of code should be moved to `createTaskFromWFManifest`. A task manifest should always include a schedule. When creating a task from a workflow manifest, these changes now initialize a new task in `createTaskFromWFManfiest`, and should also initialize a schedule at that time.

cmd/snapctl/task.go
```diff
@@ -114,14 +114,226 @@ func stringValToInt(val string) (int, error) {
...
+		if err := t.checkScheduleWindow(*start, *stop); err != nil {
+			return err
+		}
+		// if it's a valid window, then use the (new) start/stop values to set the
+		// start/stop date/times for our new schedule
+		t.Schedule.StartTime = start
+		t.Schedule.StopTime = stop
+		// and return a nil error (indicating success)
+		return nil
+	} else if (start == nil && scheduleStart == nil) || (stop == nil && scheduleStop == nil) {
```
geauxvirtual: With the first `if` returning if duration is not nil, this does not need to be an `else if`.

cmd/snapctl/task.go
```diff
@@ -124,17 +335,13 @@ func createTaskUsingTaskManifest(ctx *cli.Context) error {
...
-	t.Name = ctx.String("name")
-	if t.Version != 1 {
-		return fmt.Errorf("Invalid version provided")
-	}
-
-	// If the number of failures does not specific, default value is 10
-	if t.MaxFailures == 0 {
-		fmt.Println("If the number of maximum failures is not specified, use default value of", DefaultMaxFailures)
-		t.MaxFailures = DefaultMaxFailures
+	// merge any CLI optiones specified by the user (if any) into the current task;
```
geauxvirtual: optiones -> options

cmd/snapctl/task.go
```diff
@@ -100,14 +100,225 @@ func createTask(ctx *cli.Context) error {
 	return err
 }
 
+func stringValToInt(val string) (int, error) {
+	// parse the input (string) value as an integer (log a Fatal
```
geauxvirtual: Maybe just word this comment as return an error instead of logging a fatal error message since we don't actually log a message in snapctl, but just return an error to the user.

cmd/snapctl/task.go
```diff
@@ -100,14 +100,225 @@ func createTask(ctx *cli.Context) error {
...
+//    |<--------------------------->|
+//  start                         stop
+//
+//    +-----------------------------+ (only duration specified, implies start is the current time)
+//    |                             |
+//    |---------------------------->|
+//  Now()        duration
+//
+func (t *task) setWindowedSchedule(start *time.Time, stop *time.Time, duration *time.Duration) error {
+	// if this task does not have an existing schedule, create an empty
```
geauxvirtual: First comment need to be removed or updated.

cmd/snapctl/task.go
```diff
@@ -100,14 +100,225 @@ func createTask(ctx *cli.Context) error {
...
+		errStr := splitErr[len(splitErr)-1]
+		// return a value of zero and the error encountered during string parsing
+		return 0, fmt.Errorf("Value '%v' cannot be parsed as an integer (%v)", val, errStr)
+	}
+	// return the integer equivalent of the input string and a nil error (indicating success)
+	return int(parsedField), nil
+}
+
+// Parses the command-line parameters and uses them to override the underlying schedule
+// in this task (if any) or set a schedule for that task (if one is not already defined).
```
geauxvirtual: I would remove `(if any)` as this makes it sound like a task manifest does not need to include a schedule.

cmd/snapctl/task.go
```diff
@@ -100,6 +100,20 @@ func createTask(ctx *cli.Context) error {
...
+	// error if we cannot parse the value as an integer)
+	parsedField, err := strconv.Atoi(val)
+	if err != nil {
+		splitErr := strings.Split(err.Error(), ": ")
+		errStr := splitErr[len(splitErr)-1]
+		// return a value of zero and the error encountered during string parsing
+		return 0, fmt.Errorf("Value '%v' cannot be parsed as an integer (%v)", val, errStr)
+	}
+	// return the integer equivalent of the input string and a nil error (indicating success)
+	return int(parsedField), nil
```
IRCody: Is this int() call needed? Atoi already returns an int. See [Atoi](https://golang.org/pkg/strconv/#Atoi).

cmd/snapctl/task.go
```diff
@@ -198,7 +206,11 @@ func createTaskUsingWFManifest(ctx *cli.Context) error {
 
 	// Deadline for a task
 	dl := ctx.String("deadline")
-	maxFailures := ctx.Int("max-failures")
+	maxFailuresStr := ctx.String("max-failures")
+	maxFailures, err := stringValToInt(vmaxFailuresStr)
```
IRCody: Should `vmaxFailuresStr` be `maxFailuresStr`?

cmd/snapctl/task.go
```diff
@@ -198,7 +206,11 @@ func createTaskUsingWFManifest(ctx *cli.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("Value for command-line option 'max-failures' (%v) cannot be parsed as an integer\n", maxFailuresStr)
+	}
```
IRCody: It seems like the whole point of having stringValToInt() instead of using strconv.Atoi directly is to provide a customized error message, but then here we discard that error message. Is there value in using stringValToInt over strconv.Atoi since this is the case?

core/task.go
```diff
@@ -197,7 +197,12 @@ func CreateTaskFromContent(body io.ReadCloser,
 	if tr.Name != "" {
 		opts = append(opts, SetTaskName(tr.Name))
 	}
-	opts = append(opts, OptionStopOnFailure(10))
+
+	// if a MaxFailures value is included as part of the task creation request
+	if tr.MaxFailures > 0 {
```
IRCody: What is the reasoning behind restricting this to be >0?

cmd/snapctl/task.go
```diff
@@ -100,14 +100,224 @@ func createTask(ctx *cli.Context) error {
...
+		errStr := splitErr[len(splitErr)-1]
+		// return a value of zero and the error encountered during string parsing
+		return 0, fmt.Errorf("Value '%v' cannot be parsed as an integer (%v)", val, errStr)
+	}
+	// return the integer equivalent of the input string and a nil error (indicating success)
+	return parsedField, nil
+}
+
+// Parses the command-line parameters (if any) and uses them to override the underlying
+// sehedule for  this task or to set a schedule for that task (if one is not already defined,
```
IRCody: schedule*

IRCody:
LGTM. @geauxvirtual do you have any outstanding concerns?

cmd/snapctl/task.go
```diff
@@ -100,14 +100,231 @@ func createTask(ctx *cli.Context) error {
...
+		// and the duration
+		newStart := time.Now().Add(createTaskNowPad)
+		newStop := newStart.Add(*duration)
+		t.Schedule.StartTime = &newStart
+		t.Schedule.StopTime = &newStop
+		return nil
+	}
+	// if only the start date/time was specified, we will use it to replace
+	// the current schedule's start date/time
+	if start != nil && stop == nil {
```
geauxvirtual: Missing a condition if both start and stop are not nil to set StartTime and StopTime to those values.

cmd/snapctl/task.go
```diff
@@ -100,14 +100,231 @@ func createTask(ctx *cli.Context) error {
...
+		t.Schedule.StartTime = start
+		return nil
+	}
+	// if only the stop date/time was specified, we will use it to replace the
+	// current schedule's stop date/time
+	if start == nil && stop != nil {
+		t.Schedule.StopTime = stop
+	}
```
geauxvirtual: Add a return to this if statement.

tjmcs: Do you want that return statement just for clarity, @geauxvirtual? The line directly below that if is actually a `return nil`, which accomplishes the same thing...

cmd/snapctl/task.go
tjmcs: @geauxvirtual; I just pushed up a change that I think addresses both this line comment and your next one (by ensuring that either value will be set if the corresponding input is non-nil)

cmd/snapctl/task.go
```diff
@@ -100,14 +100,230 @@ func createTask(ctx *cli.Context) error {
...
+	// Grab the interval for the schedule (if one was provided). Note that if an
+	// interval value was not passed in and there is no interval defined for the
+	// schedule associated with this task, it's an error
+	interval := ctx.String("interval")
+	if !ctx.IsSet("interval") && interval == "" && t.Schedule.Interval == "" {
+		return fmt.Errorf("Usage error (missing interval value); when constructing a new task schedule an interval must be provided")
+	}
+	// if a start, stop, or duration value was provided, then it's a 'windowed' schedule
+	isWindowed := (start != nil || stop != nil || duration != nil)
+	// if there was an interval was passed in, then attempt to parse it (first as
```
geauxvirtual: `if there was an interval passed in` or `if an interval was passed in`

cmd/snapctl/task.go
```diff
@@ -100,14 +100,230 @@ func createTask(ctx *cli.Context) error {
...
+	// current schedule's stop date/time
+	if stop != nil {
+		t.Schedule.StopTime = stop
+	}
+	// if we get this far, then just return a nil error (indicating success)
+	return nil
+}
+
+// parse the command-line options and use them to setup a new schedule for this task
+func (t *task) setScheduleFromCliOptions(ctx *cli.Context) error {
```
geauxvirtual: There is still a logic issue in this function. The function is trying to determine the schedule based on command line parameters passed in; however, a task manifest already defines the schedule type. For a windowed schedule, if just `--interval` is set on the command line, then `isWindowed` will be set to false, and this function will return an error saying `cannot replace existing schedule type with a simple schedule`.

cmd/snapctl/task.go
```diff
@@ -100,14 +100,230 @@ func createTask(ctx *cli.Context) error {
...
+	}
+	// return the integer equivalent of the input string and a nil error (indicating success)
+	return parsedField, nil
+}
+
+// Parses the command-line parameters (if any) and uses them to override the underlying
+// schedule for this task or to set a schedule for that task (if one is not already defined,
+// as is the case when we're building a new task from a workflow manifest).
+//
+// Note; in this method, any of the following types of time windows can be specified:
```
geauxvirtual: `Note:`

cmd/snapctl/task.go
tjmcs: I think the most recent push to this branch resolves this issue, @geauxvirtual; LMK if you see anything else

cmd/snapctl/task.go
```diff
@@ -100,14 +100,231 @@ func createTask(ctx *cli.Context) error {
...
+		// then return an error
+		return fmt.Errorf("Usage error (schedule type mismatch); cannot replace existing schedule of type '%v' with a new, 'windowed' schedule", t.Schedule.Type)
+	}
+	// if a duration was passed in, determine the start and stop times for our new
+	// 'windowed' schedule from the input parameters
+	if duration != nil {
+		// if start and stop were both defined, then return an error (since all three parameters cannot be used
+		// to define the 'windowed' schedule)
+		if start != nil && stop != nil {
+			return fmt.Errorf("Usage error (too many parameters); only two of the parameters that define the window (start time, stop time and duration) can be specified for a 'windowed' schedule")
```
geauxvirtual: This error message needs to be reworded. There are technically 5 parameters that can be specified for a windowed schedule: start-time, start-date, stop-time, stop-date, and duration.

cmd/snapctl/task.go
```diff
@@ -100,14 +100,231 @@ func createTask(ctx *cli.Context) error {
...
+	if duration != nil {
+		// if start and stop were both defined, then return an error (since specifying the
+		// start, stop, *and* duration for a 'windowed' schedule is not allowed)
+		if start != nil && stop != nil {
+			return fmt.Errorf("Usage error (too many parameters); the window start, stop, and duration cannot all be specified for a 'windowed' schedule")
+		}
+		// if start is set and stop is not then use duration to create stop
+		if start != nil && stop == nil {
+			newStop := start.Add(*duration)
+			t.Schedule.StopTime = &newStop
```
geauxvirtual: Looks like this needs `t.Schedule.StartTime = start`

cmd/snapctl/task.go
```diff
@@ -100,14 +100,231 @@ func createTask(ctx *cli.Context) error {
...
+			return nil
+		}
+		// if stop is set and start is not then use duration to create start
+		if stop != nil && start == nil {
+			newStart := stop.Add(*duration * -1)
+			t.Schedule.StartTime = &newStart
```
geauxvirtual: And need `t.Schedule.StopTime = stop` added here.

tjmcs: I think the latest commits on this branch address this line comment (and the previous one), but LMK if you see anything else that needs to be changed, @geauxvirtual

geauxvirtual:
LGTM

PR #1135, opened 2016-08-11
Added USE collector plugin to the plugin catalog

sandlbn:
Summary of changes:
- Added USE collector plugin to the plugin catalog

@intelsdi-x/snap-maintainers


mjbrender:
LGTM. We'll reorganize the Plugin Catalog next time we generate it.

PR #1143, opened 2016-08-16
Fix #1136: Log better error message when generating workflow fails

### LINKED ISSUE #1136: Scheduler not started error returned incorrectly from createTask
In [scheduler.createTask](https://github.com/intelsdi-x/snap/blob/master/scheduler/scheduler.go#L296-L319)

```go
	// Generate a workflow from the workflow map
	wf, err := wmapToWorkflow(wfMap)
	if err != nil {
		te.errs = append(te.errs, serror.New(err))
		f := buildErrorsLog(te.Errors(), logger)
		f.Error(ErrSchedulerNotStarted.Error())
		return nil, te
	}
```

ErrSchedulerNotStarted is returned incorrectly. I'm not sure exactly what the error should be but it should be related to parsing the wmap into a workflow since that is what is triggering the error.

geauxvirtual:
Fixes #1136

Summary of changes:
- Corrects log message when a workflow map is unable to be generated from a workflow.

Testing done:
- Verified tests pass.

@intelsdi-x/snap-maintainers

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -297,7 +297,7 @@ func (s *scheduler) createTask(sch schedule.Schedule, wfMap *wmap.WorkflowMap, s
 	if s.state != schedulerStarted {
 		te.errs = append(te.errs, serror.New(ErrSchedulerNotStarted))
 		f := buildErrorsLog(te.Errors(), logger)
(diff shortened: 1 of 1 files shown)


IRCody:
LGTM

PR #1149, opened 2016-08-17
Fix #1129 - Add schedule validation check for task manifests in snapctl

### LINKED ISSUE #1129: snapctl should return a better error if a task manifest is passed in without a schedule
Currently, if a task manifest does not include a schedule, snapctl will return `runtime error: invalid memory address or nil pointer dereference`. When creating a task, there is no validation that the task manifest passed in included a schedule. An error should be returned to the user saying the task manifest needs to include a schedule instead of a vague runtime error.

### LINKED ISSUE #1130: Passing a task manifest without a workflow throws a panic in snapd
Trying to create a task from a task manifest that does not include a workflow causes a PANIC in the goroutine trying to create the task. The output of the panic is displayed in the snapd log file and returned to the user as an error in creating the task.

geauxvirtual:
Fixes #1129 

Summary of changes:
- Validates a task from a task manifest includes a schedule and the schedule is not empty.
- Moves the version check for a task manifest into a better function that includes other validations.

Testing done:
- Verified passing a task manifest without a schedule returns an error.
- Verified passing a task manifest with an empty schedule returns an error.

@intelsdi-x/snap-maintainers

### DIFF
--- a/cmd/snapctl/task.go
+++ b/cmd/snapctl/task.go
@@ -343,6 +343,11 @@ func createTaskUsingTaskManifest(ctx *cli.Context) error {
 		return fmt.Errorf("Unsupported file type %s\n", ext)
 	}
 
+	// Validate task manifest includes schedule, workflow, and version
+	if err := validateTask(t); err != nil {
+		return err
+	}
+
 	// merge any CLI options specified by the user (if any) into the current task;
 	// if an error is encountered, return it
 	if err := t.mergeCliOptions(ctx); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/snapctl/task.go
```diff
@@ -343,6 +339,11 @@ func createTaskUsingTaskManifest(ctx *cli.Context) error {
 		return fmt.Errorf("Unsupported file type %s\n", ext)
 	}
 
+	// Validate task manifest includes schedule, workflow, and version
+	if err := validateTask(t); err != nil {
+		return err
+	}
+
```
jcooklin: Should `validateTask(t)` be called after `t.mergeCliOptions` below?

geauxvirtual: The idea is to validate that both a schedule and workflow were included in the task manifest file and throw an appropriate error before the code attempts to use certain aspects of the created task struct so that the user will receive a better error than `runtime error: invalid memory address or nil pointer dereference`.

jcooklin:
LGTM - however this PR doesn't fix #1130 as it is worded.  The issue exists if curl is used.

geauxvirtual:
Good catch. I can add to this PR some validation on the server side, or pull out the commits around #1130 and open a new PR with those commits plus the additional server side changes.

PR #1160, opened 2016-08-24
Makes plugins used in example tasks consistent

### LINKED ISSUE #1141: Plugin name in distributed-mock-file.json incorrect
In [distributed-mock-file.json](https://github.com/intelsdi-x/snap/blob/master/examples/tasks/distributed-mock-file.json) the file publishers name was not updated when it was changed from file => mock-file which makes the example not work without having an extra dependency outside of snap.

kjang96:
Fixes #1141 

Summary of changes:
- Changed all references to publisher plugins in examples/tasks to reference "file" publisher rather than "mock-file"
- Tasks are now consistent with the top-level README.md and won't run tasks that call plugins that haven't been loaded as per README instructions

Testing done:
- Verified that the examples in top-level README now work

@intelsdi-x/snap-maintainers

### DIFF
--- a/examples/tasks/mock-file.json
+++ b/examples/tasks/mock-file.json
@@ -24,7 +24,7 @@
                     "process": null,
                     "publish": [
                         {
-                            "plugin_name": "mock-file",                            
+                            "plugin_name": "file",                            
                             "config": {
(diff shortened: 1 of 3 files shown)


IRCody:
LGTM

Issue #1158, opened 2016-08-23
Clarify scope and purpose of examples/ folder


mjbrender:
The goal of files/folders in `examples/` is not clear. I'd like to discuss a couple points (and can check them off as we do).

Biggest question: 
- [ ] 1. What is the first example we want everyone to use with Snap? 

More on tactics: 
- [ ] 2. How should the `tasks/` folder differ from example tasks in each plugin repository?
- [ ] 3. How can we ensure all these files - tasks, configs, full demos - run? 
- [ ] 4. When is it best to have a folder in this repo v a separate one (example from [Velocity](https://github.com/nanliu/snap-demo-velocity))? 

A few opinions on where we go from here:

1 -> I'd like people to be able to deploy "an Application", Snap, Grafana as containers, not include any persistence layer, and have a dashboard of Tasks running in a Grafana Graph. We don't need persistence to do this when using the Snap App from raintank ([link](https://grafana.net/plugins/raintank-snap-app)). We then collect stats from the Docker API. 

The "Application" we use as example could be as simple as `dd if=/dev/urandom | bzip2 -9 > /dev/null` running. Just something to give it a graph worth seeing.

If we want a non-Docker version, it could be a Vagrantfile with limited dependencies. When I think of a reliable demo, I think of psutil. We use it extensively when we demo and it always has numbers to show. It should be a good baseline to get Grafana Graphs together worth showing if we don't go the Docker route. 

2 -> I'm unsure here. Only opinion is that we have examples that include different types of plugins (collector, processor, publisher). 

3 -> I could see requirements coming from our build process. I don't have an opinion on what they are, but it'd be awesome to write them down. 

4 -> Not sure again. I think it's scoping comes from build workflow. 

Relates to: 
#786 << Have valid examples, including the metadata (especially stop using mock)
#1007 << Adding further schedule types

IRCody:
>1 What is the first example we want everyone to use with Snap?

>1 -> I'd like people to be able to deploy "an Application", Snap, Grafana as containers, not include any persistence layer, and have a dashboard of Tasks running in a Grafana Graph. We don't need persistence to do this when using the Snap App from raintank (link). We then collect stats from the Docker API.

>The "Application" we use as example could be as simple as dd if=/dev/urandom | bzip2 -9 > /dev/null running. Just something to give it a graph worth seeing.

[...]

PR #1161, opened 2016-08-24
Multi-Plugin Example

nichelle-hall:
An example of using multiple plugins to gather specific measurements and publish them to Grafana through some persistence layer (Graphite or Influxdb).

Summary of changes:
- Added another folder in the examples directory complete with Dockerfiles, tasks, scripts, and Grafana dashboard that will be utilized when the users runs the example


mjbrender:
I really like this idea of an example @NichelleBot. Comments will be in-line. We may adjust this as we scope the examples folder further based on #1158, but that shouldn't stop this from merging.

examples/multi-plugins/README.md
```diff
@@ -1,6 +1,42 @@
...
+  - Gets, loads and starts snap plugins
+  - Creates and starts a task that collects from psutil, cpu, and cgroups then publishes to InfluxDB
+  
+Start the example by running `run-multi.sh`
+
+### Requirements
+- docker
+- docker-compose 
+- netcat 
+- SNAP_PATH env var set to the snap build directory
```
mjbrender: this demo you're using as a baseline is working on a revision. That shouldn't be an issue.

mjbrender:
@NichelleBot can you squash down these commits? Thanks!

jcooklin:
@NichelleBot and I have talked about this PR **not** requiring netcat, SNAP_PATH, or building anything to work.  

We also discussed how this example can be integrated into a large e2e test as part of another issue.

candysmurf:
@jcooklin, should this example be in this repo?

jcooklin:
@candysmurf, not as it is currently formed but in general yes.  We discussed making the following changes.

- [ ] Remove the need for compiling anything to run the example
 - A requirement to run the example(s) will be that docker-compose is available and configured
- [ ] Add a large (end-2-end) test that tests the example 
 - Configure CI to run the e2e test

PR #1165, opened 2016-08-29
Fixed Bug #1042 Snapctl error message should truncate

### LINKED ISSUE #1042: Snapctl error message on Task should truncate
```
build/bin/snapctl task list                                                                                                              ✭
ID 					 NAME 						 STATE 		 HIT 	 MISS 	 FAIL 	 CREATED 		 LAST FAILURE
0c9ad69d-f8a2-409c-b162-de6f3e3bd4d5 	 Task-0c9ad69d-f8a2-409c-b162-de6f3e3bd4d5 	 Disabled 	 10 	 0 	 10 	 2:16PM 6-30-2016 	 Publish call error: Post http://127.0.0.1:8086/write?consistency=&db=db_test&precision=s&rp=default: dial tcp 127.0.0.1:8086: getsockopt: connection refused
```

The FAIL message causes line wrapping and is either truncating really long or not truncated.

It would be better to have it truncate to keep the table consistent and have the user use a specific `task info` to see details like fail messages.

kjlyon:
Fixes #1042 

Summary of changes:
-  Long error messages will not wrap any more unless --verbose is specified
-  Error will display until edge of terminal window and end with '...'
-  If terminal is large enough for whole error message or --verbose is specified, whole error message will be shown

Testing done:
-  Tested with mock-file.yaml and mock.go to display error messages of varying length

@intelsdi-x/snap-maintainers

Error message size adjusts based on size of terminal so that it does not wrap.

### DIFF
--- a/cmd/snapctl/task.go
+++ b/cmd/snapctl/task.go
@@ -37,6 +37,7 @@ import (
 	"github.com/intelsdi-x/snap/mgmt/rest/client"
 	"github.com/intelsdi-x/snap/scheduler/wmap"
 	"github.com/robfig/cron"
+	"golang.org/x/crypto/ssh/terminal"
 
 	"github.com/ghodss/yaml"
 )
@@ -461,6 +462,8 @@ func mergeDateTime(tm, dt string) *time.Time {
 
 func listTask(ctx *cli.Context) error {
 	tasks := pClient.GetTasks()
+	termWidth, _, _ := terminal.GetSize(int(os.Stdout.Fd()))
+	verbose := ctx.Bool("verbose")
 	if tasks.Err != nil {
 		return fmt.Errorf("Error getting tasks:\n%v\n", tasks.Err)
 	}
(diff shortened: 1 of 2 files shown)


cmd/snapctl/task.go
```diff
@@ -477,6 +480,10 @@ func listTask(ctx *cli.Context) error {
 		"LAST FAILURE",
 	)
 	for _, task := range tasks.ScheduledTasks {
+		lastErrorMessage := task.LastFailureMessage
+		if !ctx.Bool("verbose") && len(lastErrorMessage)+153 > termWidth && termWidth > 164 {
```
IRCody: What is the significance of 153 and 164?

kjlyon: 154 is the number of characters in the output before the error:
61b55838-0605-430c-a209-2f3328880141   	 Task-61b55838-0605-430c-a209-2f3328880141     	 Running       	 13    	 0     	 6     	 4:17PM 8-29-2016  **error is printed here**

164 is 154 + "LAST FAILURE". So that if the title bar wraps then the message will wrap too. If the title bar doesn't wrap, then the error will truncate.

IRCody: I think adding a comment to say what they mean could be good though so that in the future if the format of the table changes we will know to update this also.

It also seems to work if the task name is shorter than the default one which was my other concern here.

cmd/snapctl/task.go
```diff
@@ -477,22 +481,39 @@ func listTask(ctx *cli.Context) error {
 		"LAST FAILURE",
 	)
 	for _, task := range tasks.ScheduledTasks {
+		lastErrorMessage := task.LastFailureMessage
```
IRCody: I don't think this is needed anymore. Probably makes sense to just pass task.LastFailureMessage in directly.

IRCody:
LGTM

PR #1177, opened 2016-09-01
Added option to So clause to fix Rest client error

### LINKED ISSUE #1176: Rest client test at line 430 is unreliable
* /home/travis/gopath/src/github.com/intelsdi-x/snap/mgmt/rest/client/client_func_test.go 
  Line 430:
  Expected: 'error 0: Task is already stopped. '
  Actual:   'error 0: Subscription does not exist '
  (Should be equal)

kjlyon:
Fixes #1176 

Summary of changes:
- Added option to So clause to fix following error 

```
So(t2.Err.Error(), ShouldBeIn, []string{"error 0: Task is already stopped. ", "error 0: Subscription does not exist "})
```

**Error**
```
home/travis/gopath/src/github.com/intelsdi-x/snap/mgmt/rest/client/client_func_test.go 
  Line 430:
  Expected: 'error 0: Task is already stopped. '
  Actual:   'error 0: Subscription does not exist '
  (Should be equal)
```

Testing done:
- Unit testing

@intelsdi-x/snap-maintainers

### DIFF
--- a/mgmt/rest/client/client_func_test.go
+++ b/mgmt/rest/client/client_func_test.go
@@ -427,7 +427,7 @@ func TestSnapClient(t *testing.T) {
 					//try stopping again to make sure channel is closed
 					t2 := c.StopTask(tt.ID)
 					So(t2.Err, ShouldNotBeNil)
-					So(t2.Err.Error(), ShouldEqual, "error 0: Task is already stopped. ")
+					So(t2.Err.Error(), ShouldBeIn, []string{"error 0: Task is already stopped. ", "error 0: Subscription does not exist "})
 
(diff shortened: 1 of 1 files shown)


candysmurf:
This can fix one kind of interim Travis issues. @jcooklin, @IRCody, Do you guys foresee any other issues such an error may indicate? Please share your thoughts.

jcooklin:
LGTM

@candysmurf In this case either of the two errors would be correct since there is a race at play.   Specific to this test which error is returned doesn't matter as much as the an error is returned and it's one of the two ('task already stopped' or 'subscription doesn't exist').

PR #1183, opened 2016-09-07
Fix #1130: Validate task workflow and schedule when creating task request.

### LINKED ISSUE #1130: Passing a task manifest without a workflow throws a panic in snapd

geauxvirtual:
Fixes #1130

Summary of changes:
- Validates workflow to not be nil or empty
- Validate schedule to not be nil or empty

Testing done:
- Validated proper error was returned when trying to create a task without a workflow or schedule.
- Verified tests pass.

@intelsdi-x/snap-maintainers

request.

### DIFF
--- a/core/task.go
+++ b/core/task.go
@@ -161,7 +161,7 @@ type TaskCreationRequest struct {
 	Version     int               `json:"version"`
 	Deadline    string            `json:"deadline"`
 	Workflow    *wmap.WorkflowMap `json:"workflow"`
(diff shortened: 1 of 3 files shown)


candysmurf:
👍 , LGTM

PR #1185, opened 2016-09-07
Added functionality for environment variables in SNAP_PATH

### LINKED ISSUE #1184: Errors occur when SNAP_PATH has variables in it.
Environment variables are not read correctly when SNAP_PATH has variables like $GVM_ROOT, $gvm_go_name, or $gvm_pkgset_name as seen in the example below: `"\$GVM_ROOT/pkgsets/\$gvm_go_name/\$gvm_pkgset_name/src/github.com/intelsdi-x/snap/build"`

This results in many errors and file not found messages, even when the file is there. 
Each call to `os.Getenv("SNAP_PATH")` should actually be `os.ExpandEnv(os.Getenv("SNAP_PATH"))`

kjlyon:
Fixes #1184 

Summary of changes:
- Wrapped each call to `os.Getenv("SNAP_PATH")` with `os.ExpandEnv(..)`
- Now Snap_Paths that have variables such as the one below are interpreted correctly.
`\$GVM_ROOT/pkgsets/\$gvm_go_name/\$gvm_pkgset_name/src/github.com/intelsdi-x/snap/build`

Testing done:
- Unit testing

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/fixtures/fixtures.go
+++ b/control/fixtures/fixtures.go
@@ -32,7 +32,7 @@ import (
 
 var (
 	PluginName = "snap-plugin-collector-mock2"
-	SnapPath   = os.Getenv("SNAP_PATH")
+	SnapPath   = os.ExpandEnv(os.Getenv("SNAP_PATH"))
 	PluginPath = path.Join(SnapPath, "plugin", PluginName)
 
 	JSONRPCPluginName = "snap-plugin-collector-mock1"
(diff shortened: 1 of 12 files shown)


control/fixtures/fixtures.go
```diff
@@ -32,7 +32,7 @@ import (
 
 var (
 	PluginName = "snap-plugin-collector-mock2"
-	SnapPath   = os.Getenv("SNAP_PATH")
+	SnapPath   = os.ExpandEnv(os.Getenv("SNAP_PATH"))
```
candysmurf: if you can put this as a "func" then calling it. It seems better than changing it all over the places. The advantage for simply change it in one place in the future.

candysmurf:
@kjlyon, chatted with Joel. As changes across multiple packages, it seems hard to generalize it in a central location plus the simplicity of the change. We'll just leave it as it's and merge the change.

Thanks!

PR #1187, opened 2016-09-07
Fix issue #1186

### LINKED ISSUE #1186: Plugins errors should not be displayed at Debug level but at Error level
If you look at [this code snippet](https://github.com/intelsdi-x/snap/blob/master/control/plugin/execution.go#L163) and now that stdout and stderr of plugins are remapped to snapd it seems reasonable to assume that errors occurring in plugins should be logged as errors in snapd so that user gets notified about them using proper logging level at snapd startup

obourdon:
Fixes #1186 

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/plugin/execution.go
+++ b/control/plugin/execution.go
@@ -160,8 +160,7 @@ func (e *ExecutablePlugin) captureStderr() {
 			execLogger.
 				WithField("io", "stderr").
 				WithField("plugin", path.Base(e.cmd.Path())).
-				Debug(stdErrScanner.Text())
-
+				Error(stdErrScanner.Text())
 		}
 	}()
 }


jcooklin:
The go log packages writes to stderr.  I'm not convinced that snapd should print plugin log statements at the `Error` log level just because it's being written to standard err.

candysmurf:
@jcooklin, what level of logging are you suggesting? 

Probably it's a good time to fix logs to separate them into different log levels so that debugging logs won't show up in the releases.

jcooklin:
@obourdon as we discussed there are PROs and CONs for capturing the plugins stderr as either DEBUG or ERROR.   I'm leaning towards staying conservative meaning that when someone wants to debug a plugin they will run snapd in debug mode to effectively surface whatever log messages the plugin might be logging (on stderr).   

In the future we can decide if we want to extend the contract snapd establishes with plugins.  For instance, I can imagine a plugin exposing to snapd, during the initial handshake, that it supports a strict implementation of structured logging such that snapd could interpret the plugins logs as it receives them and uses the same level when it prints it to snaps consolidated log (snaps stderr by default) however this will come at a cost (complexity and cycles).

candysmurf:
@obourdon, based on @jcooklin's feedback, how do you like to deal with this PR? thanks.

obourdon:
@candysmurf @jcooklin if you look at [this issue on the snap collector plugin for elasticsearch](https://github.com/intelsdi-x/snap-plugin-collector-elasticsearch/issues/16), you might want to reconsider integrating this change ;-).

If you still think that there is more work to do around it as mentioned in the [details of the comment above](https://github.com/intelsdi-x/snap/pull/1187#issuecomment-245456047), I have absolutely no strong objection in withdrawing this PR for now

Please advise, thanks

candysmurf:
@obourdon, thanks for getting back to us. I'll leave this to @jcooklin for his implication analysis and recommendations.

jcooklin:
>@candysmurf @jcooklin if you look at this issue on the snap collector plugin for elasticsearch, you might want to reconsider integrating this change ;-).

The plugin in question should have returned an error instead of panicking [here](https://github.com/intelsdi-x/snap-plugin-collector-elasticsearch/blob/10ec335a857b3bdedb36fdaddd077083465abd00/elasticsearch/elasticsearch.go#L195).  👈 (cc: @candysmurf )

All things being considered I'd like to leave the behavior as it currently is for the reasons I stated ☝️ .

PR #1191, opened 2016-09-09
Expose plugin load timeout as a configurable option

### LINKED ISSUE #1164: Need to block processing tasks until all plugins are loaded
On a virtualized and shared environment which is quite loaded (several users running heavy jobs) snapd is launched using auto discovery path (-a option) to load 10+ plugins (cpu df disk interface load meminfo processes swap smart heka) and a task which runs every 10 seconds and collect all metrics from the above collectors.

Since the load increased on the system, it seems like the task stops right away in error after startup of snap daemon because at the first iteration of metrics gathering, some plugins are not loaded yet.

obourdon:
Fixes #1164 

Summary of changes:
- Expose plugin load timeout in plugin manager so that the plugin can take more time to load on a busy system

Testing done:
- successfully tested on our environment where 12 plugins are autoloaded with an additional auto-loaded task which was failing due to this issue

@intelsdi-x/snap-maintainers


obourdon:
Could you please review this PR and tell me if I need to rework some parts ? Thanks

cc: [@intelsdi-x/snap-maintainers](https://github.com/orgs/intelsdi-x/teams/snap-maintainers)

jcooklin:
LGTM

PR #1193, opened 2016-09-09
Added a warning when plugin name is uppercase

### LINKED ISSUE #1178: case sensitive plugin name
Currently, it is possible to load 2 plugins with the same name, if we change the case.

So, "Use" and "use" collector could both be loaded with the same version.

This could open snap to a few subtle issues, such as removing the wrong plugin.

Not sure how to solve this, but talking on slack, @IRCody and @thomastaylor312 had some ideas. 

I'm opening this issue so can move the conversation here and other can contribute.

kindermoumoute:
Fixes #1178 

Summary of changes:
- Added a warning when a a plugin with uppercase characters is added.

Testing done:
- Renamed a plugin name (e.g. the Snap publisher) and load it.
```bash
$ snapctl plugin load snap-plugin-publisher-file
WARN[2016-09-09T10:20:25-07:00] DEPRECATED: the name of the plugin is in uppercase, the next version of Snap will not support it anymore.  _module=control plugin-name=filE plugin-type=publisher plugin-version=2
```

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/plugin_manager.go
+++ b/control/plugin_manager.go
@@ -85,6 +85,14 @@ func newLoadedPlugins() *loadedPlugins {
 
 // add adds a loadedPlugin pointer to the table
 func (l *loadedPlugins) add(lp *loadedPlugin) serror.SnapError {
+	lowerName := strings.ToLower(lp.Meta.Name)
+	if lowerName != lp.Meta.Name {
+		controlLogger.WithFields(log.Fields{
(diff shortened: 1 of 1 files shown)


kindermoumoute:
I'm closing because we will solve the problem by converting every plugin name in lowercase in the plugin catalog. Loading a task will also convert plugin names to lowercase.

PR #1194, opened 2016-09-09
Make plugin name not case sensitive

### LINKED ISSUE #1178: case sensitive plugin name

kindermoumoute:
Fixes #1178 

Summary of changes:
- Display a warning every time a plugin with uppercase characters is loaded, or used from a task.
- Convert the plugin name to lowercase when loaded in the plugin catalog.
- Convert the plugin names to lowercase when a task is created.

Testing done:
- Loaded a plugin with a correct name (e.g. "file"), and run a task with it (e.g. "file")==> no warning.
- Loaded a plugin with an incorrect name (e.g. "FILE"), and run a task with an incorrect name (e.g. "FILE")==> Warning when loading, `snapctl plugin list` shows a lowercase name, second warning when loading the task.
- Loaded a plugin with an incorrect name (e.g. "FILE"), and run a task with the correct name (e.g. "file")==> Warning when loading, `snapctl plugin list` shows a lowercase name, still display the second warning when loading the task.
- Loaded a plugin with a correct name (e.g. "file"), and run a task with an incorrect name (e.g. "FILE")==> no warning.

@intelsdi-x/snap-maintainers

All plugins names will load as lower case, even if they have an upper case in them.

### DIFF
--- a/control/plugin/execution.go
+++ b/control/plugin/execution.go
@@ -26,6 +26,7 @@ import (
 	"io"
 	"os/exec"
 	"path"
+	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


IRCody:
LGTM

jcooklin:
LGTM

PR #1197, opened 2016-09-09
Updating Wish List design

mjbrender:
Designing a different layout for the wishlist as a prototype for us to auto-generate it after this.

* Linking wish list items to open issues that will explore it
* Updating recommendations on how to use the wish list so people keep opening issues


mjbrender:
@nanliu this is something we should review together as part of the auto-generation of the catalog. Talk to you more next week 👌

jcooklin:
LGTM

Issue #1214, opened 2016-09-16
Large numbers of calls to os.Hostname() when reapplying tags


jtlisi:
After running a stack trace there was a large number of reads to `/proc/sys/kernel/hostname`

```
strace -f -p `pidof snapd
41504 openat(AT_FDCWD, "/proc/sys/kernel/hostname", O_RDONLY|O_CLOEXEC) = 35
.
. for each collected metric
.
41504 openat(AT_FDCWD, "/proc/sys/kernel/hostname", O_RDONLY|O_CLOEXEC) = 35
```

After running a pstree it was confirmed that the origin was snapd
`{snapd}(41504)`

The source of all of these calls to os.Hostname() seem to come from this little snippet in [control/control.go](https://github.com/intelsdi-x/snap/blob/master/control/control.go)

```
go func() {
	for m := range cMetrics {
		// Reapply standard tags after collection as a precaution.  It is common for
		// plugin authors to inadvertently overwrite or not pass along the data
		// passed to CollectMetrics so we will help them out here.
		for i := range m {
			m[i] = addStandardAndWorkflowTags(m[i], allTags)
		}
		metrics = append(metrics, m...)
		wg.Done()
	}
}()
```

[control/metrics.go](https://github.com/intelsdi-x/snap/blob/master/control/metrics.go)
```
func addStandardAndWorkflowTags(m core.Metric, allTags map[string]map[string]string) core.Metric {
	hostname, err := hostnameReader.Hostname()
```
```
// hostnameReader, hostnamer created for mocking
func init() {
	hostnameReader = &hostnameReaderType{}
}

type hostnamer interface {
	Hostname() (name string, err error)
}

type hostnameReaderType struct{}

func (h *hostnameReaderType) Hostname() (name string, err error) {
	return os.Hostname()
}
```

While the performance impact isn't that serious due to the nature of `/proc/`, I do think its expensive enough to store in a var. Let me know what you think and I put in a PR.

Thanks,

Jacob

IRCody:
I think it makes sense to reduce the number of calls since the hostname shouldn't be changing super often. Were you thinking about having it called once per collect or just once in per snapd startup?

jtlisi:
Once per collect seems easiest to me, and the performance impact of one call is basically negligible. What do you think?

IRCody:
I think that makes sense. You are going to make a pr to do this?

jtlisi:
Yes, I'll try and have it done by Monday.

PR #1215, opened 2016-09-19
Make hostname a stored variable so it is not retrieved for each metri…

jtlisi:
Fixes #

Summary of changes:
- Moved the retrieval of the hostname out the addStandardAndWorkflowTags so that it is not iterated for each metric at every collection.

Testing done:
- Checked the stack trace of a running agent to ensure calls to os.hostname() are not present

@intelsdi-x/snap-maintainers

…c when tags are applied

### DIFF
--- a/control/metrics.go
+++ b/control/metrics.go
@@ -46,17 +46,46 @@ var (
 
 // hostnameReader, hostnamer created for mocking
 func init() {
-	hostnameReader = &hostnameReaderType{}
+	host, err := os.Hostname()
+	if err != nil {
+		log.WithFields(log.Fields{
+			"_module": "control",
+			"_file":   "metrics.go,",
+			"_block":  "addStandardAndWorkflowTags",
+			"error":   err.Error(),
+		}).Error("Unable to determine hostname")
+		host = "not_found"
+	}
+	hostnameReader = &hostnameReaderType{hostname: host, hostnameRefreshTTL: time.Hour, lastRefresh: time.Now()}
 }
 
 type hostnamer interface {
-	Hostname() (name string, err error)
+	Hostname() (name string)
 }
 
-type hostnameReaderType struct{}
+type hostnameReaderType struct {
+	hostname           string
+	hostnameRefreshTTL time.Duration
+	lastRefresh        time.Time
+}
 
-func (h *hostnameReaderType) Hostname() (name string, err error) {
-	return os.Hostname()
+func (h *hostnameReaderType) Hostname() (name string) {
+	if time.Now().After(h.lastRefresh.Add(h.hostnameRefreshTTL)) {
+		host, err := os.Hostname()
+		if err != nil {
+			log.WithFields(log.Fields{
+				"_module": "control",
+				"_file":   "metrics.go,",
+				"_block":  "addStandardAndWorkflowTags",
+				"error":   err.Error(),
+			}).Error("Unable to determine hostname")
+			host = "not_found"
+		}
+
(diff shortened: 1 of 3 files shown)


control/metrics.go
```diff
@@ -55,17 +55,28 @@ var (
...
+	h, err := os.Hostname()
+	if err != nil {
+		log.WithFields(log.Fields{
+			"_module": "control",
+			"_file":   "metrics.go,",
+			"_block":  "addStandardAndWorkflowTags",
+			"error":   err.Error(),
+		}).Error("Unable to determine hostname")
+	}
+	hostnameReader = &hostnameReaderType{hostname: h}
```
IRCody: I think this is going to result in only calling once per run of snapd since init will only be run at that time, instead of the intended once per collect call.

jcooklin: What is the argument against using a simple hostname global set once when snapd starts?

IRCody: Just that it was the intention of the PR. Is there a case where the hostname would change more regularly than snapd would be restarting?

jtlisi: Gotcha, I'll change it accordingly

control/metrics.go
```diff
@@ -50,22 +50,33 @@ var (
...
+			"_block":  "addStandardAndWorkflowTags",
+			"error":   err.Error(),
+		}).Error("Unable to determine hostname")
+	}
+	hostname = h
 }
 
 type hostnamer interface {
-	Hostname() (name string, err error)
+	Hostname() (name string)
```
IRCody: Do we still need the hostnamer interface?

control/metrics.go
lynxbat: Does this means a hostname change *requires* a restart of snapd?

control/metrics.go
jtlisi: I believe not, I'll take that out as well.

control/metrics.go
jtlisi: @lynxbat I believe so, I can move the hostname out into the metrics.go file and have it run every collection. This would also have to be done at [control/plugin_manager.go](https://github.com/intelsdi-x/snap/blob/master/control/plugin_manager.go)

lynxbat: I think requiring a restart is nasty and we should avoid that one.

Maybe having a TTL that is pretty long (60min)? And have it invalidate and refresh past that. You can just have the accessor function hide this behavior. This way we prevent too much churn (imagine a 10ms interval) but prevent hard locked restart.

We would just need to add comments to the README to call out the 60min refresh. We could postpone exposing that as config until someone asks for it.

jtlisi:
Hey does anyone know why the legacy build is failing on 1.5. I didn't touch the segment of code that seems to be failing.

IRCody:
@jtlisi: It was failing because there is a flakey tests that hasn't been tracked down yet. Was definitely unrelated to your change. I restarted the test when I saw your comment. Sorry for not commenting at the time.

IRCody:
@jtlisi: I think this looks good. Can you rebase it down to 1 commit that starts the message off as "Fixes #1214:...." so that github will close that issue when this is merged?

The only thing I think could be changed is what happens when there is an error with the hostname. In addition to logging it, would it make sense to change the returned hostname to something like "Unknown"?

IRCody:
Thanks for updating this @jtlisi.

@jcooklin: Do you have any other concerns before merging?

jcooklin:
LGTM

PR #1220, opened 2016-09-19
Support specifying an instance of dynamic metric

### LINKED ISSUE #654: Task creation fails when specifying a specific instance of a dynamic metric.

IzabellaRaulin:
Fixes #654

### Specifying an instance of dynamic metric

_snap-plugin-collector-mock is used as example_

#### Before:
 - user cannot define which of dynamic metric is interesting to be collected, so there was only the possibility of collecting metrics for all instance:
```
 "collect": {
            "metrics": {
                "/intel/mock/*/baz": {}
            },
```

#### After:
- user can define which specified instance of dynamic metric should be collected (what helps to avoid collecting all of them even user does not need them)

```
 "collect": {
            "metrics": {
                "/intel/mock/host0/baz": {}
            },
```
That is very useful for fully aware users who know which available instances - however that information can be also gained by running task with `"/intel/*` at first and check available metrics.

- tuple can be used to declared more than one specified instances to collect
```
 "collect": {
            "metrics": {
                "/intel/mock/(host0;host1)/baz": {}
            },
```

Notice, that tuple items are between brackets and they are separated by semicolon

### How to prepare collector plugin to use that
Notice, that most of our collector plugins are ready to use that feature. In plugin.CollectMetrics plugin should be prepared for incoming metric with "*", but also should has `else` which allows to collect specified dynamic metrics. See example:
- snap-plugin-collector-cpu ([L199](https://github.com/intelsdi-x/snap-plugin-collector-cpu/blob/master/cpu/cpu.go#L199) - [L216](https://github.com/intelsdi-x/snap-plugin-collector-cpu/blob/master/cpu/cpu.go#L216))
- snap-plugin-collector-interface([L158](https://github.com/intelsdi-x/snap-plugin-collector-interface/blob/master/iface/interface.go#L158) - [L172](https://github.com/intelsdi-x/snap-plugin-collector-interface/blob/master/iface/interface.go#L172))
- snap-plugin-collector-disk ([L196](https://github.com/intelsdi-x/snap-plugin-collector-disk/blob/master/disk/disk.go#L196) - [L229](https://github.com/intelsdi-x/snap-plugin-collector-disk/blob/master/disk/disk.go#L229))
- snap-plugin-collector-docker ([L206](https://github.com/intelsdi-x/snap-plugin-collector-docker/blob/master/docker/docker.go#L206) - [L211](https://github.com/intelsdi-x/snap-plugin-collector-docker/blob/master/docker/docker.go#L211))  
_(good example how to handle dynamic elements in dynamic metric)_

### Summary of changes:
**a) support defining specific instance of dynamic metric in task manifest**
User might define in task manifest: 
- `/intel/mock/*/bar` to collect metric for all available instance
- or specific dynamic metric:
 - `/intel/mock/host0/bar` (in that case only one metric will be collected)
 - `/intel/mock/(host0;host1)/bar` (that returns two metrics, works like logic AND)
 - `/intel/mock/host0/*` (that returns all available metrics for instance "host0" from mock collector)

In situation when there is no such instance of dynamic metric (e.g. for snap-plugin-collector-mock `host10` is not available), an appropriate error should be returned (the collector is responsible for that, see [snap-plugin-collector-mock1/mock/mock.go#L73](https://github.com/intelsdi-x/snap/blob/master/plugin/collector/snap-plugin-collector-mock1/mock/mock.go#L73))

 **b) changing the way of matching dynamic query due to allowing the specifying of dynamic metrics, but there is no change from user perspective**
   - removing regexp!!!
   - removing restriction of not allowed chars in metric's namespace
   - now metric namespace cannot only contain wildcard at the end and tuple (low probability) 

**c) adding error message to trying Fetch() on empty metric catalog**
   _Example:_
   running `snapctl metric list` when NO PLUGIN has been already loaded 
  image

### Testing done:
- small, medium, legacy tests
- running [examples/tasks/mock-file_specific_instance.json](https://github.com/IzabellaRaulin/snap/blob/fixed_654/examples/tasks/mock-file_specific_instance.json) 
- running [examples/tasks/mock-file_tuple.json](https://github.com/IzabellaRaulin/snap/blob/fixed_654/examples/tasks/mock-file_tuple.json) 
- checking plugin swapping (ver1, ver2) for both of above tasks

### Additional tests:
**a) specified disk instance for snap-plugin-collector-disk**

    Before                                             | Now
    --------------                                   | ----------------
    "/intel/procfs/disk/*/ops_read": {}, | "/intel/procfs/disk/sda/ops_read": {},

See [task manifest](https://github.com/IzabellaRaulin/snap-plugin-collector-disk/blob/specified_instance/examples/tasks/disk_sda-file.json)
image
Related with https://github.com/intelsdi-x/snap-plugin-collector-disk/issues/12

**b) specified interface for snap-plugin-collector-interface**

    Before                                                 | Now
    --------------                                       | ----------------
    "/intel/procfs/iface/*/bytes_recv": {},  | "/intel/procfs/iface/eno1/bytes_recv": {}, 

See [task manifest for iface eno1](https://github.com/IzabellaRaulin/snap-plugin-collector-interface/blob/specific_interface/examples/task/iface_eno1-file.json)
image
Related with https://github.com/intelsdi-x/snap-plugin-collector-interface/issues/13

See [task manifest for iface enp2s0](https://github.com/IzabellaRaulin/snap-plugin-collector-interface/blob/specific_interface/examples/task/iface_enp2s0-file.json)
image

**c) specified docker container ID for snap-plugin-collector-docker**

Before                                                                 | Now
 --------------                                                       | ----------------
 "/intel/docker/\*/stats/network/\*/rx_bytes": {}  | "/intel/docker/5e06711bddc3/stats/network/eth0/rx_bytes": {} 

See [task manifest for container 5e06711bddc3](https://github.com/IzabellaRaulin/snap-plugin-collector-docker/blob/specified_instance/examples/tasks/docker_5e06711bddc3-file.json)
image


control/mttrie_medium_test.go
```diff
@@ -0,0 +1,530 @@
+// +build medium
```
jcooklin: legacy -> medium 🎉 🐢

docs/TASKS.md
```diff
@@ -114,21 +114,52 @@ An example json task that uses remote targets can be found under [examples](http
...
 #### collect
 
-The collect section describes which metrics to collect. Metrics can be enumerated explicitly via:
- - a concrete _namespace_
- - a wildcard, `*`
- - a tuple, `(m1|m2|m3)`
+The collect section describes which metrics are requested to be collected.
 
-The tuple begins and ends with brackets and items inside are separeted by vertical bar. It works like logical `or`, so it gives an error only if none of these metrics can be collected.
+Metrics can be enumerated explicitly via using:
```
jcooklin: s/explicitly via using/using/

docs/TASKS.md
```diff
@@ -114,21 +114,52 @@ An example json task that uses remote targets can be found under [examples](http
...
 
-| Metrics declared in task manifest | Collected metrics                                              |
-|:----------------------------------|:---------------------------------------------------------------|
-| /intel/mock/\*                    | /intel/mock/foo <br/> /intel/mock/bar <br/> /intel/mock/\*/baz |
-| /intel/mock/(foo\|bar)            | /intel/mock/foo <br/> /intel/mock/bar <br/>                    |
-| /intel/mock/\*/baz                | /intel/mock/\*/baz                                             |
+ a) **concrete _namespace_**
+
+Declaring the same metric's name in task manifest as it is presented on metric list (see `snapctl metric list`)
```
jcooklin: Declaring a metric's name exactly as it appears in the metric catalog.

docs/TASKS.md
```diff
@@ -114,21 +114,52 @@ An example json task that uses remote targets can be found under [examples](http
...
+	|
+    /intel/mock/bar                             |  /intel/mock/bar
+	|
+    /intel/mock/\*/baz <br/> _(dynamic metric)_ |  /intel/mock/host0/baz <br/> /intel/mock/host1/baz <br/> /intel/mock/host2/baz  <br/> /intel/mock/host3/baz  <br/> /intel/mock/host4/baz <br/> /intel/mock/host5/baz <br/> /intel/mock/host6/baz <br/> /intel/mock/host7/baz  <br/> /intel/mock/host8/baz <br/> /intel/mock/host9/baz <br/><br/> _(collect metrics for all instances of the dynamic metric)_
+
+ 
+ 
+ b) **_specified_ _instance_ of dynamic metrics**
+
+ This refers to dynamic metric which namespace contains dynamic element represented by an asterisk within the meaning of representation of dynamic value (e.g. hostname, cgroup id, etc.)
```
jcooklin: Specifying a dynamic metric refers to providing a `value` in place of the dynamic element in the namespace (e.g. hostname, cgroup id, etc.).  It's important to remember that dynamic elements are represented by an asterisk when presented in the metric catalog.

docs/TASKS.md
```diff
@@ -114,21 +114,52 @@ An example json task that uses remote targets can be found under [examples](http
...
+
+ In task manifest can be declared a specific instance of dynamic metric to collect and only that specific one will be collected (if exist)
```
jcooklin: When a  task manifest contains a specific instance of a dynamic metric only that instance will be collected.  If the value does not exist the task will error.

docs/TASKS.md
```diff
@@ -114,21 +114,52 @@ An example json task that uses remote targets can be found under [examples](http
...
+
+ Metrics requested in task manifest | Collected metrics
+------------------------------------|------------------------
+/intel/mock/host0/baz <br/><br/> _(specific instance of "/intel/mock/*/baz")_ |  /intel/mock/host0/baz <br/><br/> _(only this one metric will be collected)_
+
+
+ c) **dynamic _query_**
+
+ This allows to express metrics to collect by using
```
jcooklin: Dynamic queries are those that contain **wildcards** (`*`) and/or tuples of values `(x;y;z)`.

control/metrics.go
```diff
@@ -447,39 +290,134 @@ func (mc *metricCatalog) RmUnloadedPluginMetrics(lp *loadedPlugin) {
...
+				version:            catalogedmt.Version(),
+				lastAdvertisedTime: catalogedmt.LastAdvertisedTime(),
+				tags:               catalogedmt.Tags(),
+				policy:             catalogedmt.Plugin.ConfigPolicy.Get(catalogedmt.Namespace().Strings()),
+				config:             catalogedmt.Config(),
+				unit:               catalogedmt.Unit(),
+				description:        catalogedmt.Description(),
+				subscriptions:      catalogedmt.SubscriptionCount(),
+			}
+			returnedmts = appendIfUnique(returnedmts, returnedmt)
```
IzabellaRaulin: This never happens, **so definitely `appendIfUnique` can be removed.** Probably I suggested on appendIfMissing which is use for metrics' keys. Thanks for catching that.

The only possible when it could have sense is requesting the same items in tuple (as "/intel/mock/(host0;host0)/baz"), but I think it's ok for that returning two metrics - user should notice that duplication and make correction on his own.

control/metrics.go
```diff
@@ -706,3 +563,123 @@ func addStandardAndWorkflowTags(m core.Metric, allTags map[string]map[string]str
...
+func containsTuple(nsElement string) (bool, []string) {
+	tupleItems := []string{}
+	if isTuple(nsElement) {
+		if strings.ContainsAny(nsElement, "*") {
+			// an asterisk covers all tuples cases (eg. /intel/mock/(host0;host1;*)/baz)
+			// so to avoid retrieving the same metric more than once, return only '*' as a tuple's items
+			tupleItems = []string{"*"}
+		} else {
+			tuple := strings.TrimSuffix(strings.TrimPrefix(nsElement, core.TuplePrefix), core.TupleSuffix)
+			tuple = strings.Replace(tuple, core.TupleSeparator, " ", -1)
```
IzabellaRaulin: The reason why I replace `core.TupleSeparator` with space is allowing to use "strings.Fields" which removing white spaces, so all of the following cases are possible:
- no spaces: `/intel/mock/(host0;host1)/baz`
- spaces before and after semicolon: `/intel/mock/(host0 ; host1)/baz`
- spaces next to brackets: `/intel/mock/( host0 ; host1 )/baz`

Now I see that will not work for tuple's item consisting of two words. So, there is no sense to use "strings.Fields". Removing white spaces by using "string.TrimSpace" for each of tuple's items will give the same result and do not impact on two-word-item.

#### Conclusion:
This fragment:
```
tuple := strings.TrimSuffix(strings.TrimPrefix(nsElement, core.TuplePrefix), core.TupleSuffix)
tuple = strings.Replace(tuple, core.TupleSeparator, " ", -1)
tupleItems = strings.Fields(tuple)
```
will be changed to:
```
items := strings.Split(tuple, core.TupleSeparator)
 // removing all leading and trailing white space
 for _, item := range items {
  tupleItems = append(tupleItems, strings.TrimSpace(item))
}
```

control/metrics.go
```diff
@@ -706,3 +563,123 @@ func addStandardAndWorkflowTags(m core.Metric, allTags map[string]map[string]str
...
+			tupleItems = strings.Fields(tuple)
+		}
+		return true, tupleItems
+	}
+	return false, nil
+}
+
+// findTuplesMatches returns all matched combination of queried tuples in incoming namespace, where by a tuple there is a mean of `(host0;host1;host3)`
```
IzabellaRaulin: thanks, done

docs/TASKS.md
IzabellaRaulin: ok, done. I leave the fragment "(see `snapctl metric list`)" - let's me know is it ok if this stays.

docs/TASKS.md
IzabellaRaulin: ok, thanks

docs/TASKS.md
IzabellaRaulin: ok, thanks

docs/TASKS.md
IzabellaRaulin: Please, confirm that's what you mean:
### Before:

This allows to express metrics to collect by using
 - **a wildcard `*`** - that matches with any value in the metric namespace or, if the wildcard is in the end, with all metrics with the given prefix
 - **a tuple** `(metric1;metric2;metric3)` - that matches with all items separated by semicolon and works like logical _and_, so it gives an error if even one of these metrics cannot be collected

### After:
Dynamic queries are those that contain **wildcards** (`*`) and/or **tuples of values** `(x;y;z)`

control/mttrie_medium_test.go
IzabellaRaulin: Yes :) The first of 121 lines have not been changed, but they fulfill medium tests requirements.
The rest of new test cases I preferred added as a medium tests - so, there was not sense to keep the file metrics_test.go (legacy) longer.

docs/TASKS.md
jcooklin: I was suggesting s/This allows to express metrics to collect by using/Dynamic queries are those that contain **wildcards** (`*`) and/or tuples of values `(x;y;z)`/

I would leave the list.

IzabellaRaulin: I agree that the list should stay. To keep logical flow, I used "Dynamic queries are those that contain" as an introduction to this list. Please let me know if this **after2** is acceptable.
### After2:
Dynamic queries are those that contain:
 - **wildcards** `*` - that matches with any value in the metric namespace or, if the wildcard is in the end, with all metrics with the given prefix
 - and/or **tuples of values** `(x;y;z)` - that matches with all items separated by semicolon and works like logical _and_, so it gives an error if even one of these items cannot be collected

control/metrics.go
IzabellaRaulin: I added also small tests which cover that (see [control/metrics_small_test.go](https://github.com/intelsdi-x/snap/pull/1220/files#diff-d03b6fc79a4cf76dee147f92b0cd76ef))

docs/TASKS.md
jcooklin: 👍

control/mttrie.go
```diff
@@ -145,81 +141,161 @@ func (mtt *mttNode) Add(mt *metricType) {
...
+		default:
+			children = parent.gatherChildren(ns[0])
+		}
+		nodes = append(nodes, children...)
+		return nodes
+	}
+	children = parent.gatherChildren(ns[0])
+
+	for _, child := range children {
+		nodes = child.search(nodes, ns[1:])
```
IRCody: Does this not just re-assign over the return value, which would end up returning the results from the last child looked at? I feel like maybe I am missing something.

control/subscription_group.go
```diff
@@ -51,13 +51,12 @@ type ManagesSubscriptionGroups interface {
 	Add(id string, requested []core.RequestedMetric,
 		configTree *cdata.ConfigDataTree,
 		plugins []core.SubscribedPlugin) []serror.SnapError
-	Get(id string) ([]core.Metric, []serror.SnapError, error)
+	Get(id string) (map[string]metricTypes, []serror.SnapError, error)
```
IRCody: It seems out of place for this to not take a type from core and instead uses `metricTypes` which is different than `metricType`.

docs/TASKS.md
```diff
@@ -114,21 +114,52 @@ An example json task that uses remote targets can be found under [examples](http
...
 
-The tuple begins and ends with brackets and items inside are separeted by vertical bar. It works like logical `or`, so it gives an error only if none of these metrics can be collected.
+Metrics can be enumerated using:
 
-| Metrics declared in task manifest | Collected metrics                                              |
-|:----------------------------------|:---------------------------------------------------------------|
-| /intel/mock/\*                    | /intel/mock/foo <br/> /intel/mock/bar <br/> /intel/mock/\*/baz |
-| /intel/mock/(foo\|bar)            | /intel/mock/foo <br/> /intel/mock/bar <br/>                    |
-| /intel/mock/\*/baz                | /intel/mock/\*/baz                                             |
+ a) **concrete _namespace_**
```
IRCody: I think these a/b/c sections would read better with headings (i.e. `# Concrete namespace` etc).

control/metrics.go
```diff
@@ -75,8 +66,15 @@ func errorMetricNotFound(ns string, ver ...int) error {
 	return fmt.Errorf("Metric not found: %s", ns)
 }
 
-func errorMetricContainsNotAllowedChars(ns string) error {
-	return fmt.Errorf("Metric namespace %s contains not allowed characters. Avoid using %s", ns, listNotAllowedChars())
+func errorMetricsNotFound(ns string, ver ...int) error {
```
IzabellaRaulin: Do you mean  "errorMetricNotFound" and "errorMetricsNotFound" and relationship between them?

- `errorMetricsNotFoud` was added to highlight that this error coming from Fetch() or GetMetric**s**() and that give a possibility to:
  - add handling an error for ns = "/" - that means that try to get all cataloged metrics returns empty mts, so metric catalog is empty
  - more descriptive error message for requested ns which contains query, e.g. "No metric below the given namespace: /intel/mock/"

- whereas `errorMetricNotFound` is returned by validation that this concrete ns (without query) exists in metric catalog

That is the only difference between them, so**there is a possibility to combined them in one** - I will do that.

control/metrics.go
```diff
@@ -91,16 +89,8 @@ func errorMetricDynamicElementHasNoName(value, ns string) error {
...
-// exemplary output: "brackets [( ) [ ] { }], spaces [ ], punctuations [. , ; ? !], slashes [| \ /], carets [^], quotations [" ` ']"
-func listNotAllowedChars() string {
-	var result string
-	for groupName, chars := range notAllowedChars {
-		result += fmt.Sprintf(" %s %s,", groupName, chars)
-	}
-	// trim the comma in the end
-	return strings.TrimSuffix(result, ",")
+func errorMetricElementHasTuple(value, ns string) error {
+	return fmt.Errorf("A element %s should not define tuple for namespace %s.", value, ns)
```
IzabellaRaulin: This is to avoid misinterpretation of tuple - plugin should NOT expose metric with words in brackets separated by semicolon (e.g. '(sth;sth)`) which will be interpreted as a tuple in the GetMetrics(). 

So, basically restriction of not allowed characters are not longer needed, but the only case which needs to be checked in `validateNamespace` is that related with tuple. But, as you can see, probability of fulfillment [this condition in `containsTuple`](https://github.com/intelsdi-x/snap/pull/1220/files#diff-a1ff046a48fbfbca760e04470fc7ac73R561) by namespace's element is rather low.  See [tests](https://github.com/intelsdi-x/snap/pull/1220/files#diff-d03b6fc79a4cf76dee147f92b0cd76efR100) to find more how it work.

control/metrics.go
```diff
@@ -706,3 +546,126 @@ func addStandardAndWorkflowTags(m core.Metric, allTags map[string]map[string]str
...
+// and not contain elements which might be erroneously recognized as a tuple
+func validateMetricNamespace(ns core.Namespace) error {
+	value := ""
+	for _, i := range ns {
+		// A dynamic element requires the name while a static element does not.
+		if i.Name != "" && i.Value != "*" {
+			return errorMetricStaticElementHasName(i.Value, i.Name, ns.String())
+		}
+		if i.Name == "" && i.Value == "*" {
+			return errorMetricDynamicElementHasNoName(i.Value, ns.String())
```
IzabellaRaulin: Yes, definitely - I've just removed this `validateMetricNamespace` to the end of file, so that is not my changes (it's existing code, see [snap/blob/master/control/metrics.go#L376](https://github.com/intelsdi-x/snap/blob/master/control/metrics.go#L376)), but it should be changed to `isDynamic()`. Thanks for notice that!

control/control.go
```diff
@@ -688,85 +674,80 @@ func (p *pluginControl) verifyPlugin(lp *loadedPlugin) error {
...
+			}
+			// set config to metric
+			mt.config = cfg
+
+			// loaded plugin which exposes the metric
+			lp := mt.Plugin
+			key := lp.Key()
+
+			// groups metricTypes by a plugin.Key()
+			pmt, _ := newMetricsGroupedByPlugin[key]
```
IzabellaRaulin: Please, look on snap/master [control/control.go#L1092](https://github.com/intelsdi-x/snap/blob/master/control/control.go#L1092) - this is exactly the same code. So, it definitely works and do not panic, because where there is no item in map under the key (for example map is empty), the empty `pmt` will be returned (but not nil). 

I take this fragment because there is no need to do `groupMetricByPlugin()` every time at each collection. It's enough to do that when metrics and collector are taken for subscription_group. I limited making changes in the rest of existing code if these are not really necessary, so I kept that fragment in form as it is on current master. 

But I agree that this could be changed in the future together with adding `newMetricTypes()` which will return empty metricTypes or even change `map[string]metricTypes` to `map[string]*metricTypes` could be also taken into consideration.

@IRCody, what do you thing about keeping that as it is? Is it acceptable for you?

control/control.go
```diff
@@ -848,15 +829,26 @@ func (p *pluginControl) FetchMetrics(ns core.Namespace, version int) ([]core.Cat
...
-	return p.metricCatalog.Get(ns, ver)
+	return p.metricCatalog.GetMetric(ns, ver)
+}
+
+func (p *pluginControl) GetMetrics(ns core.Namespace, ver int) ([]core.CatalogedMetric, error) {
+	mts, err := p.metricCatalog.GetMetrics(ns, ver)
+	if err != nil {
+		return nil, err
+	}
+	rmts := make([]core.CatalogedMetric, len(mts))
```
IzabellaRaulin: It has to stay here. I cannot move that to p.metriccatalog.GetMetrics() becuse that will remove information about plugin which is keep in metricType and available to retrieve form it.
Why:
- metricType includes core.CatalogedMetricType
- core.CatalogedMetricType does NOT include metricType

This "for" loop is only needed for a slice. For a single metricType there is no problem to return that directly.

See how it looks on snap master [snap/blob/master/control/control.go#L860](https://github.com/intelsdi-x/snap/blob/master/control/control.go#L860).

control/control_test.go
```diff
@@ -1280,6 +1284,167 @@ func TestCollectMetrics(t *testing.T) {
 	})
 }
 
+func TestCollectMetrics_DynamicMetrics(t *testing.T) {
```
IzabellaRaulin: yes, I agree. Changed

control/control_test.go
```diff
@@ -1280,6 +1284,167 @@ func TestCollectMetrics(t *testing.T) {
...
+				// cache duration
+				So(pool.Plugins()[1].HitCount(), ShouldEqual, 1)
+				So(pool.Plugins()[2].HitCount(), ShouldEqual, 0)
+				c.Stop()
+			})
+		})
+	})
+}
+
+func TestCollectMetrics_SpecifiedDynamicMetrics(t *testing.T) {
```
IzabellaRaulin: yes, I agree. Changed

control/mttrie.go
```diff
@@ -110,15 +107,14 @@ func (m *MTTrie) RemoveMetric(mt metricType) {
 	if a != nil {
 		for v, x := range a.mts {
 			if mt.Version() == x.Version() {
-				// Delete the metric from the node
+				// delete this metric from the node
```
IzabellaRaulin: > Why change from children->descendants

I changed children to descendants because it was confusing - it's gathering all descendants, not only children (direct descendants). Now it becomes consistent through this file and children mean direct descendants. Before it was not clear.

> change the capitalization of a comment
Please take a look on another comments in this file and how they look:
- https://github.com/intelsdi-x/snap/blob/master/control/mttrie.go#L132
- https://github.com/intelsdi-x/snap/blob/master/control/mttrie.go#L178

So, I decided to change  "Delete" > "delete" to keep consistence with the rest of comments.

control/mttrie.go
```diff
@@ -145,81 +141,161 @@ func (mtt *mttNode) Add(mt *metricType) {
...
 	}
-	if node.mts == nil {
-		return nil, errorMetricNotFound("/" + strings.Join(ns, "/"))
+	// there is an expectation that only one metric should be fitted
+	if len(mts) > 1 {
+		return nil, fmt.Errorf("Incoming namespace `%s` is too ambiguous (version: %d)", "/"+strings.Join(ns, "/"), ver)
 	}
-	var mts []*metricType
-	for _, mt := range node.mts {
+	return mts[0], nil
```
IzabellaRaulin: Because it has to have one element at least after `GetMetrics()`, so there is no need to check it once again in `GetMetric()`. If there is no element, an error will be returned in [line 179](https://github.com/intelsdi-x/snap/pull/1220/files#diff-b40077b597d83dcd21ab10660b31e40aR179).

Please, see [mttrie.go/GetMetrics line 207](https://github.com/intelsdi-x/snap/pull/1220/files#diff-b40077b597d83dcd21ab10660b31e40aR190)

control/mttrie.go
IzabellaRaulin: It searches which children of `mtt` node fulfill incoming the first element of `ns` and append the results to `nodes` which are the results of searching. 
The next phase is do such searching for these children, but for the rest of ns elements (I mean `ns[1:]`).
In the end, returned `nodes` contains all nodes retrieved from metric tree which fufill requested `ns`. Notice, that it is not said that they will have the same parents.
```
func (mtt *mttNode) search(nodes []*mttNode, ns []string) []*mttNode {
     [...]
     return nodes
}
```

docs/TASKS.md
IzabellaRaulin: These are the points under "collect" section, so I prefer to leave it as it is now. But I do not exclude change that in the future.

control/metrics.go
```diff
@@ -706,3 +546,126 @@ func addStandardAndWorkflowTags(m core.Metric, allTags map[string]map[string]str
...
+			}
+		}
+		return true, tupleItems
+	}
+	return false, nil
+}
+
+// findTuplesMatches returns all matched combination of queried tuples in incoming namespace, where a tuple is in the form of `(host0;host1;host3)`
+// if the incoming namespace does not contain any tuple, return the incoming namespace as the only item in output slice
+// if the incoming namespace contains a tuple, return the copies of incoming namespace with appropriate values set to namespaces' elements
```
IzabellaRaulin: ok, done

control/subscription_group.go
IzabellaRaulin: I tried to use metricType as a first, but it has metricType.Policy(), metricTypes does not have to, so only metricTypes can be used as input for plugin.CollectMetrics() - is that make sense?

control/metrics.go
IzabellaRaulin: ok, done

docs/TASKS.md
IzabellaRaulin: I think it's not so easy - these points are in "collect section", so it needs to be `##### Concrete namespace`, so it will be really small (as you can see below)
##### Concrete namespace

Based on that, I prefer to keep that in this form as it is. Do You agree?

control/mttrie.go
IRCody: I think I was just sleepy last night. Thanks for clearing it up.

control/control.go
IRCody: My comment was mostly about how weird it looks (which is true whether it was in snap previously or not) but I don't think it has to be addressed as part of this PR.

control/control.go
IRCody: I see. Thanks for clearing it up.

control/metrics.go
IRCody: Thanks, I wasn't sure what cases this would come up in.

control/metrics.go
```diff
@@ -447,39 +289,134 @@ func (mc *metricCatalog) RmUnloadedPluginMetrics(lp *loadedPlugin) {
...
+			"_block":  "get-metric",
+			"error":   err,
+		}).Error("error getting metric")
+		return nil, err
+	}
+
+	ns = catalogedmt.Namespace()
+
+	if isDynamic, _ := ns.IsDynamic(); isDynamic {
+		// specify instance of dynamicMetric
```
IRCody: I think that makes it more clear (at least for me).

control/metrics.go
IzabellaRaulin: I talked with @IRCody and we agree that this is not a subject of this PR, so it can be done as separate pull request.

IRCody: I think it's fine to leave this work out of the PR to avoid making additional unrelated changes.

docs/TASKS.md
IRCody: It's a subjective style thing and I'm not sure if we have a definitive guideline for what to do. I don't have strong feelings about it.

jcooklin:
LGTM

PR #1224, opened 2016-09-21
Fix #1210: Do not start plugins unless all plugins are available for workflow

### LINKED ISSUE #1210: Plugins started on unsuccessful task start are not stopped
After creating a task, if the task is stopped and a plugin is removed that the task needs, attempting to start the task will result in starting the plugins required for the task, but not stopping those plugins when the task fails to start due to a missing required plugin. There is no way to stop these plugins without stopping snapd.

geauxvirtual:
Fixes #1210 

Summary of changes:
- Verify all plugins are available for a workflow before subscribing to pools.
- If there are any errors in the subscription process, return the errors. The subscription group will only be added to the subscriptionMap if there are no errors. This allows for a task to be started after loading a needed plugin. Previously, the task would not start reporting a subscription already existed.

Testing done:
- Verified small, medium, and legacy tests pass.
- Verified that plugins that are available are not started on failure to start a task if a plugin is not available.
- Verified that a task can be started after loading the missing plugin

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/subscription_group.go
+++ b/control/subscription_group.go
@@ -129,8 +129,11 @@ func (s subscriptionGroups) add(id string, requested []core.RequestedMetric,
 	}
 
 	errs := subscriptionGroup.process(id)
+	if errs != nil {
+		return errs
+	}
 	s.subscriptionMap[id] = subscriptionGroup
-	return errs
+	return nil
 }
 
 // Remove removes a subscription group given a subscription group ID.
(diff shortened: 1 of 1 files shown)


control/subscription_group.go
```diff
@@ -386,72 +389,53 @@ func (s *subscriptionGroup) process(id string) (serrs []serror.SnapError) {
...
 func (s *subscriptionGroup) subscribePlugins(id string,
 	plugins []core.SubscribedPlugin) (serrs []serror.SnapError) {
-	for _, sub := range plugins {
+	plgs := make([]*loadedPlugin, len(plugins))
+	// First range through plugins to verify if all required plugins
+	// are available
+	for i, sub := range plugins {
+		plg, err := s.pluginManager.get(key(sub))
+		if err != nil {
+			fmt.Println(err)
```
IRCody: Log?

control/subscription_group.go
```diff
@@ -386,72 +389,53 @@ func (s *subscriptionGroup) process(id string) (serrs []serror.SnapError) {
...
-				if err != nil {
-					serrs = append(serrs, serror.New(err))
-					return serrs
-				}
-				err = s.pluginRunner.runPlugin(latest.Details)
-				if err != nil {
-					serrs = append(serrs, serror.New(err))
-					return serrs
-				}
+				return
```
IRCody: I think this function is to long for naked returns to make sense.

control/subscription_group.go
```diff
@@ -386,72 +389,53 @@ func (s *subscriptionGroup) process(id string) (serrs []serror.SnapError) {
...
 		controlLogger.WithFields(log.Fields{
-			"name":    sub.Name(),
-			"type":    sub.TypeName(),
-			"version": sub.Version(),
+			"name":    plg.Name(),
+			"type":    plg.TypeName(),
+			"version": plg.Version(),
 			"_block":  "subscriptionGroup.subscribePlugins",
 		}).Debug("plugin subscription")
-		if sub.Version() < 1 {
```
IRCody: Are we losing functionality in the case where the veresion < 1 or was this not needed?

geauxvirtual: No functionality lost. Currently, the workflow for a plugin with no version specified (-1) is to get the plugin, then get or create the pool associated with that plugin. If a version was specified, the code would only get the plugin if the pool was eligible to be grown. The change is here to get all the plugins first, then if we all the plugins are available, subscribe to the pools, etc.

control/subscription_group.go
geauxvirtual: Extra snippet of code during some debugging.

control/subscription_group.go
IRCody: That makes sense. Thanks for the clarification.

PR #1225, opened 2016-09-21
Fixes #1222 and #1228 : Applies config defaults - Metric too ambiguous

### LINKED ISSUE #1222: Tasks don't reflect changes to plugin config
To reproduce add a plugin configuration item that will be used by a given task, start the given task and then update the configuration item.  

**Expected result**: When a task, depending on a configuration item that is not provided by the task, is changed the task should start using this new configuration item.
**Actual result**: The new configuration is ignored.

In the example below:
1. snapd is started with the following config. 
2. Plugins snap-plugin-collector-mock1 and snap-plugin-publisher-mock-file are loaded.
3. The task is left to run for 5 seconds.
4. The plugin-config for mock-file is updated 
 * `curl -X PUT --data '{"file": "/tmp/snap/snap-file2.out"}' http://localhost:8181/v1/plugins/publisher/mock-file/3/config`

```yaml
---
control:
  plugins:
    publisher:
      mock-file:
        all:
          file: /tmp/snap/snap-file.out
          test: true
```

ex

jcooklin:
Fixes #1222 and #1228 

Summary of changes for #1222 (applies config defaults):
- Renames `ReverseMerge` on ConfigDataNode to `ReverseMergeInPlace` 
- Adds `ReverseMerge` on ConfigDataNode which returns a new ConfigDataNode without modifying the source
- Adds `ApplyDefaults` on ConfigDataNode

Testing done:
- unit
- manual

Example: After a task is started the configuration for which file will be published to is updated through the API.  The next time the task fires this configuration is applied. 

after1222

-----

Summary of changes for #1228
- Updates validateMetric on subscriptionGroups to use GetMetrics instead of GetMetric since some namespace requests can expand to include multiple metrics.

Testing done:
- unit
- manual

Below is a screen capture that successful runs the tasks described in #1228.
after1228

Note: These two issues were included in the same PR since #1228 depended on #1222.  #1228 adds test coverage across `ValidateDeps` and as a result needed to apply defaults from the mocks config policy. 

@intelsdi-x/snap-maintainers


marcin-krolik:
@jcooklin Retested this PR with  #1228. Confirmed it fixes reported issue. :+1:

IRCody:
The second commit starts `Fixes #12228:` instead of #1228.

jcooklin:
Updated commit message.

marcin-krolik:
LGTM

control/control_test.go
```diff
@@ -815,9 +815,6 @@ func TestMetricConfig(t *testing.T) {
 			errs := c.subscriptionGroups.validateMetric(m1)
 			So(errs, ShouldBeNil)
 		})
-		Convey("So mock should have name: bob config from defaults", func() {
-			So(c.Config.Plugins.pluginCache["0"+core.Separator+"mock"+core.Separator+"1"].Table()["name"], ShouldResemble, ctypes.ConfigValueStr{Value: "bob"})
-		})
```
jcooklin: It wasn't valid here since the subscription group would need to be processed before the default would be applied to the plugin config cache.  I added the test back as part of the convey that runs `subscriptionGroups.ValidateMetric`.

control/subscription_group.go
```diff
@@ -339,10 +339,20 @@ func (s *subscriptionGroup) process(id string) (serrs []serror.SnapError) {
 		"metrics":    fmt.Sprintf("%+v", s.requestedMetrics),
 	}).Debug("gathered collectors")
 
-	//add processors and publishers to collectors just gathered
 	for _, plugin := range s.requestedPlugins {
+		//add processors and publishers to collectors just gathered
 		if plugin.TypeName() != core.CollectorPluginType.String() {
 			plugins = append(plugins, plugin)
+			// add defaults to plugins (exposed in a plugins ConfigPolicy)
+			if lp, err := s.pluginManager.get(
```
jcooklin: I don't think so however if you feel strongly I'll change it.

control/control_test.go
```diff
@@ -1391,7 +1394,7 @@ func TestCollectSpecifiedDynamicMetrics(t *testing.T) {
 		So(err, ShouldBeNil)
 		// metric catalog should contain the 3 following metrics:
 		// /intel/mock/foo; /intel/mock/bar; /intel/mock/*/baz
-		So(len(mts), ShouldEqual, 3)
+		So(len(mts), ShouldEqual, 4)
```
IRCody: Should we add the `/intel/mock/all/baz` to the  comment so it's accurate?

control/subscription_group.go
IRCody: I don't feel super strongly.

PR #1235, opened 2016-09-23
Use positional arguments for unloading plugins

### LINKED ISSUE #1207: Simplify `snapctl plugin unload` to use positional arguments
Proposal: Instead of accepting flags or a formatted string to unload plugins use positional arguments

Before: `snapctl plugin unload collector:mock:1` or `snapctl plugin unload -t collector -n mock -v 1`

After: `snapctl plugin unload collector mock 1`

kindermoumoute:
Fixes #1207 

Summary of changes:
- Change the syntax to unload plugin with snapctl to ` snapctl plugin unload <plugin_type> <plugin_name> <plugin_version>`.

Testing done:
- `snapctl plugin unload collector mock 1`
- `snapctl plugin unload collector "mock" 1`
- `snapctl plugin unload "" mock 1`
- `snapctl plugin unload collector mock abc`

@intelsdi-x/snap-maintainers

This change is to simplify and normalize the unload command for Snap 1.0. Please give your feedback as I Snap user, if you think it should be retro-compatible, if you agree with this solution, or if you see another way to do it.

### DIFF
--- a/cmd/snapctl/plugin.go
+++ b/cmd/snapctl/plugin.go
@@ -64,28 +64,18 @@ func loadPlugin(ctx *cli.Context) error {
 }
 
 func unloadPlugin(ctx *cli.Context) error {
-	pDetails := filepath.SplitList(ctx.Args().First())
-	var pType, pName string
-	var pVer int
-	var err error
+	pType := ctx.Args().Get(0)
+	pName := ctx.Args().Get(1)
+	pVer, err := strconv.Atoi(ctx.Args().Get(2))
 
-	if len(pDetails) == 3 {
-		pType = pDetails[0]
-		pName = pDetails[1]
-		pVer, err = strconv.Atoi(pDetails[2])
-		if err != nil {
-			return newUsageError("Can't convert version string to integer", ctx)
-		}
-	} else {
-		pType = ctx.String("plugin-type")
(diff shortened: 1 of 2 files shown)


cmd/snapctl/plugin.go
```diff
@@ -64,29 +64,19 @@ func loadPlugin(ctx *cli.Context) error {
 }
 
 func unloadPlugin(ctx *cli.Context) error {
-	pDetails := filepath.SplitList(ctx.Args().First())
-	var pType, pName string
-	var pVer int
-	var err error
+	pType := ctx.Args().Get(0)
+	pName := ctx.Args().Get(1)
+	pVer, err := strconv.Atoi(ctx.Args().Get(2))
```
IRCody: Would we need to check to verify that the number of args is correct here or does the library keep track of that for us?

kindermoumoute: If no argument `ctx.Args().Get(X)` return an empty string. So it goes to the condition below.

IRCody: 👍

marcin-krolik:
LGTM

candysmurf:
This is fine as there are only 3 fixed variables. I keep my opinion that such positional arguments won't work if the argument list is longer.

PR #1239, opened 2016-09-26
V1 Rest API Tests

kjlyon:
This is the first of a series of steps in eventually updating the Rest API. Once this PR lands we will introduce a new API version. The changes in this PR will verify that we don't inadvertently break the V1 API with the future changes.

Summary of changes:
- Wrote tests in rest_v1_tests.go that test the various routes found in the addRoutes function of server.go. Each test calls mock methods from the fixtures package and compares its response to a constant, expected response found in the mock files.
- Added mock manager classes in the fixtures class for managing config, metric, task, and tribe routes individually. Each of these classes has a mock manager, mock objects, necessary methods for implementing those, and constants for comparing against the v1 tests.

Testing done:
- Unit tests

@intelsdi-x/snap-maintainers

### DIFF
--- a/mgmt/rest/fixtures/mock_config_manager.go
+++ b/mgmt/rest/fixtures/mock_config_manager.go
@@ -1,0 +1,109 @@
+// +build legacy small medium large
+
+/*
+http://www.apache.org/licenses/LICENSE-2.0.txt
+
+Copyright 2016 Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fixtures
+
+import (
+	"github.com/intelsdi-x/snap/core"
+	"github.com/intelsdi-x/snap/core/cdata"
+	"github.com/intelsdi-x/snap/core/ctypes"
+)
+
+var mockConfig *cdata.ConfigDataNode
+
+func init() {
+	mockConfig = cdata.NewNode()
+	mockConfig.AddItem("User", ctypes.ConfigValueStr{Value: "KELLY"})
+	mockConfig.AddItem("Port", ctypes.ConfigValueInt{Value: 2})
+}
+
+type MockConfigManager struct{}
+
+func (MockConfigManager) GetPluginConfigDataNode(core.PluginType, string, int) cdata.ConfigDataNode {
+	return *mockConfig
+}
+func (MockConfigManager) GetPluginConfigDataNodeAll() cdata.ConfigDataNode {
+	return *mockConfig
+}
+func (MockConfigManager) MergePluginConfigDataNode(
+	pluginType core.PluginType, name string, ver int, cdn *cdata.ConfigDataNode) cdata.ConfigDataNode {
+	return *cdn
+}
+func (MockConfigManager) MergePluginConfigDataNodeAll(cdn *cdata.ConfigDataNode) cdata.ConfigDataNode {
+	return cdata.ConfigDataNode{}
+}
+func (MockConfigManager) DeletePluginConfigDataNodeField(
+	pluginType core.PluginType, name string, ver int, fields ...string) cdata.ConfigDataNode {
+	for _, field := range fields {
+		mockConfig.DeleteItem(field)
+
+	}
+	return *mockConfig
+}
+
+func (MockConfigManager) DeletePluginConfigDataNodeFieldAll(fields ...string) cdata.ConfigDataNode {
+	for _, field := range fields {
+		mockConfig.DeleteItem(field)
+
+	}
+	return *mockConfig
+}
+
+// These constants are the expected plugin config responses from running
+// rest_v1_test.go on the plugin config routes found in mgmt/rest/server.go
+const (
+	SET_PLUGIN_CONFIG_ITEM = `{
+  "meta": {
+    "code": 200,
+    "message": "Plugin config item(s) set",
+    "type": "config_plugin_item_created",
+    "version": 1
+  },
+  "body": {
+    "user": "Jane"
+  }
+}`
+
+	GET_PLUGIN_CONFIG_ITEM = `{
+  "meta": {
+    "code": 200,
+    "message": "Plugin config item retrieved",
+    "type": "config_plugin_item_returned",
+    "version": 1
+  },
+  "body": {
+    "Port": 2,
+    "User": "KELLY"
+  }
+}`
+
+	DELETE_PLUGIN_CONFIG_ITEM = `{
+  "meta": {
+    "code": 200,
+    "message": "Plugin config item field(s) deleted",
+    "type": "config_plugin_item_deleted",
+    "version": 1
+  },
+  "body": {
+    "Port": 2,
+    "User": "KELLY"
+  }
+}`
+)
--- a/mgmt/rest/fixtures/mock_metric_manager.go
+++ b/mgmt/rest/fixtures/mock_metric_manager.go
@@ -1,0 +1,302 @@
+// +build legacy small medium large
+
+/*
+http://www.apache.org/licenses/LICENSE-2.0.txt
+
+Copyright 2016 Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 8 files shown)


mgmt/rest/fixtures/mock_config_manager.go
```diff
@@ -0,0 +1,107 @@
...
+
+package fixtures
+
+import (
+	"github.com/intelsdi-x/snap/core"
+	"github.com/intelsdi-x/snap/core/cdata"
+	"github.com/intelsdi-x/snap/core/ctypes"
+)
+
+var mockConfig *cdata.ConfigDataNode
```
IRCody: Is it weird that mockConfig is a struct type in mgmt/rest?

mgmt/rest/fixtures/mock_metric_manager.go
```diff
@@ -0,0 +1,302 @@
...
+func (m MockCatalogedMetric) Namespace() core.Namespace {
+	return core.NewNamespace("one", "two", "three")
+}
+func (m MockCatalogedMetric) Version() int                      { return 5 }
+func (m MockCatalogedMetric) LastAdvertisedTime() time.Time     { return time.Time{} }
+func (m MockCatalogedMetric) Policy() *cpolicy.ConfigPolicyNode { return cpolicy.NewPolicyNode() }
+func (m MockCatalogedMetric) Description() string               { return "This Is A Description" }
+func (m MockCatalogedMetric) Unit() string                      { return "" }
+
+var metricCatalog []core.CatalogedMetric = []core.CatalogedMetric{
```
IRCody: Would it make sense to define pluginCatalog and this in close proximity so that they are easier to find while reading through it?

mgmt/rest/fixtures/mock_tribe_manager.go
```diff
@@ -0,0 +1,406 @@
...
+	"github.com/hashicorp/memberlist"
+	"github.com/intelsdi-x/snap/core"
+	"github.com/intelsdi-x/snap/core/serror"
+	"github.com/intelsdi-x/snap/mgmt/tribe/agreement"
+)
+
+//type mockTribe struct{}
+
+var mockTribeAgreement *agreement.Agreement
+var mockTribeMember *agreement.Member
```
IRCody: These should probably be inside a var block like:
```go
var (
        mockTribeAgreement *agreement.Agreement
        mockTribeMember *agreement.Member
)
```

mgmt/rest/rest_v1_test.go
```diff
@@ -0,0 +1,584 @@
...
+		Convey("Get tasks - v1/tasks", func() {
+			resp, err := http.Get(
+				fmt.Sprintf("http://localhost:%d/v1/tasks", r.port))
+			So(err, ShouldBeNil)
+			So(resp.StatusCode, ShouldEqual, 200)
+			body, err := ioutil.ReadAll(resp.Body)
+			So(err, ShouldBeNil)
+			responses := []string{
+				fmt.Sprintf(fixtures.GET_TASKS_RESPONSE, r.port, r.port),
+				fmt.Sprintf(fixtures.GET_TASKS_RESPONSE2, r.port, r.port)}
```
IRCody: Put this `}` on the next line.  Got to avoid hanging chads.

mgmt/rest/rest_v1_test.go
```diff
@@ -0,0 +1,584 @@
...
+			resp, err := http.Get(
+				fmt.Sprintf("http://localhost:%d/v1/tasks/:%s/watch", r.port, taskID))
+			So(err, ShouldBeNil)
+			So(resp.StatusCode, ShouldEqual, 200)
+		})
+
+		Convey("Add tasks - v1/tasks", func() {
+			reader := strings.NewReader(fixtures.TASK)
+			resp, err := http.Post(fmt.Sprintf("http://localhost:%d/v1/tasks", r.port),
+				http.DetectContentType([]byte(fixtures.TASK)), reader)
```
IRCody: This is lining up weird?

mgmt/rest/rest_v1_test.go
```diff
@@ -0,0 +1,584 @@
...
+			body, err = ioutil.ReadAll(resp.Body)
+			So(err, ShouldBeNil)
+			So(
+				fmt.Sprintf(fixtures.JOIN_TRIBE_AGREEMENT_RESPONSE_NAME_JOIN),
+				ShouldResemble,
+				string(body))
+
+		})
+	})
+}
```
IRCody: Would it make sense to break this file up so it's a little shorter and the tribe/task/metric route sections are separate?

IRCody:
Is the first commit in this PR necessary? It seems like the second commit unwinds the stuff in the first (at least somewhat).

mgmt/rest/fixtures/mock_tribe_manager.go
```diff
@@ -0,0 +1,406 @@
...
+import (
+	"net"
+
+	"github.com/hashicorp/memberlist"
+	"github.com/intelsdi-x/snap/core"
+	"github.com/intelsdi-x/snap/core/serror"
+	"github.com/intelsdi-x/snap/mgmt/tribe/agreement"
+)
+
+//type mockTribe struct{}
```
kjlyon: Whoops, I forgot to delete this. I'll fix that.

mgmt/rest/fixtures/fixtures.go
```diff
@@ -1,166 +0,0 @@
-// +build legacy small medium large
```
kjlyon: Probably. I was working with code from Joel. The code that was in fixtures.go was put into mock_metric_manager.go and other classes in the fixtures package. Should I overwrite the first commit since it isn't really relevant to how the code turned out in the end?

IRCody: Yes I think it makes sense to have them as one commit.

mgmt/rest/rest_v1_test.go
kjlyon: I don't think that would make sense because right now it takes 1 line of code to run the tests, if they were in separate files it would take 3 lines. In addition there will be a separate file for V2 tests. If we break up the tests into tribe/task/metric right now then the v2 tests should be broken up too. I don't think that is necessary.

IRCody: >I don't think that would make sense because right now it takes 1 line of code to run the tests,

Can you explain this? I'm not following.

mgmt/rest/fixtures/mock_config_manager.go
```diff
@@ -0,0 +1,107 @@
...
+func (MockConfigManager) GetPluginConfigDataNodeAll() cdata.ConfigDataNode {
+	return *mockConfig
+}
+func (MockConfigManager) MergePluginConfigDataNode(pluginType core.PluginType, name string, ver int, cdn *cdata.ConfigDataNode) cdata.ConfigDataNode {
+	return *cdn
+}
+func (MockConfigManager) MergePluginConfigDataNodeAll(cdn *cdata.ConfigDataNode) cdata.ConfigDataNode {
+	return cdata.ConfigDataNode{}
+}
+func (MockConfigManager) DeletePluginConfigDataNodeField(pluginType core.PluginType, name string, ver int, fields ...string) cdata.ConfigDataNode {
```
kjlyon: That is correct. These are mocks.

mgmt/rest/fixtures/mock_config_manager.go
kjlyon: No I think that is fine since it's in a different package.

mgmt/rest/fixtures/mock_metric_manager.go
```diff
@@ -0,0 +1,302 @@
...
+import (
+	"errors"
+	"time"
+
+	"github.com/intelsdi-x/snap/control/plugin/cpolicy"
+	"github.com/intelsdi-x/snap/core"
+	"github.com/intelsdi-x/snap/core/serror"
+)
+
+//////MockLoadedPlugin/////
```
kjlyon: These comments are just for readability of the file.

mgmt/rest/fixtures/mock_config_manager.go
```diff
@@ -0,0 +1,109 @@
...
+type MockConfigManager struct{}
+
+func (MockConfigManager) GetPluginConfigDataNode(core.PluginType, string, int) cdata.ConfigDataNode {
+	return *mockConfig
+}
+func (MockConfigManager) GetPluginConfigDataNodeAll() cdata.ConfigDataNode {
+	return *mockConfig
+}
+func (MockConfigManager) MergePluginConfigDataNode(
+	pluginType core.PluginType, name string, ver int, cdn *cdata.ConfigDataNode) cdata.ConfigDataNode {
```
IRCody: Does anyone else have opinions on this method of function definition?

jcooklin: @IRCody what specifically do you find an objection to?

mgmt/rest/rest_v1_test.go
kjlyon: It is really easy to run all the tests right now; it only takes one command. If we break them up into separate files it will take 3 commands to run all the tests - one command for each section: tribe, task, and metric.

mgmt/rest/fixtures/mock_config_manager.go
IRCody: I didn't say I had an objection.

PR #1242, opened 2016-09-28
Fix #1211: Return better error for plugin not found when starting a task

### LINKED ISSUE #1211: Return better error for when trying to start a task with a missing plugin
Related to #1210, when trying to start a task after a plugin has been removed, the only error returned is `plugin not found`, not which plugin was not found.

Here, https://github.com/intelsdi-x/snap/blob/master/control/subscription_group.go#L400, we are just returning the error from trying to get the loaded plugin, which is just a simple `plugin not found`. As opposed to here, https://github.com/intelsdi-x/snap/blob/master/control/subscription_group.go#L258, where the error is rewritten to include the plugin that could not be found.

geauxvirtual:
Fixes #1211

Summary of changes:
- Refactor and return same plugin not found error for both validating a task and starting a task.
- Refactor code creating plugin key to use key function.

Testing done:
- Verified tests pass for small, medium, and legacy.
- Verified error returns details on a plugin when a plugin is not found while trying to start a task.

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/subscription_group.go
+++ b/control/subscription_group.go
@@ -247,13 +247,7 @@ func (p *subscriptionGroups) validatePluginSubscription(pl core.SubscribedPlugin
 	}).Info(fmt.Sprintf("validating dependencies for plugin %s:%d", pl.Name(), pl.Version()))
 	lp, err := p.pluginManager.get(fmt.Sprintf("%s"+core.Separator+"%s"+core.Separator+"%d", pl.TypeName(), pl.Name(), pl.Version()))
 	if err != nil {
-		se := serror.New(fmt.Errorf("Plugin not found: type(%s) name(%s) version(%d)", pl.TypeName(), pl.Name(), pl.Version()))
-		se.SetFields(map[string]interface{}{
-			"name":    pl.Name(),
-			"version": pl.Version(),
-			"type":    pl.TypeName(),
-		})
-		serrs = append(serrs, se)
+		serrs = append(serrs, pluginNotFoundError(pl))
 		return serrs
 	}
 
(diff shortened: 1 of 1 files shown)


control/subscription_group.go
```diff
@@ -383,7 +377,7 @@ func (s *subscriptionGroup) subscribePlugins(id string,
 	for i, sub := range plugins {
 		plg, err := s.pluginManager.get(key(sub))
 		if err != nil {
-			serrs = append(serrs, serror.New(err))
+			serrs = append(serrs, pluginNotFoundError(sub))
```
IRCody: PluginNotFound is not the only error that can be returned from the `pluginManager.get` call. Do we need to worry about badkey errors also?

geauxvirtual: I'm looking through the code to see if there any chance the code could hit that error.

geauxvirtual: Looking at the code, it's seems the check for a bad key is only if the plugin is not found and breaking the key apart to see if the latest plugin needs to be grabbed. The same error `bad key` is returned for multiple scenarios, so it's never clear what part of the key is bad. 

`pluginManager.get` could return a clearer error on what plugin was not found, so then creating this error in subscription_group.go would then not be needed.

geauxvirtual: Another example in the code base of treating any error returned from `pluginManager.get` as plugin not found. https://github.com/intelsdi-x/snap/blob/master/control/plugin_manager.go#L509-L518

IRCody: Do you think `pluginManager.get` should be in charge of returning better errors?

geauxvirtual: After looking at the code, I would think the error handling would be better suited in `pluginManager.get` instead of `loadedPlugins.get`. At that point, `loadedPlugins.get` could possibly be simplified to return the loadedPlugin or nil, and let `pluginManager.get` decide what error to return: pluginNotFound or invalid key when no plugin is returned.

IRCody: Is that something you were going to do as part of this PR?

geauxvirtual: I feel refactoring that code would be better in another PR.

kindermoumoute:
LGTM

PR #1252, opened 2016-10-04
(SDI-1725) Fix #1234 Add arbitrary namespace separator

### LINKED ISSUE #1234: RFC: Arbitrary namespace separator
### Problem description
There are at least two collector plugins (df, swap) identified where dynamic namespace element value may contain "/" character, which breaks metrics namespace.

Problem was reported by Mirantis and is tracked in [github](https://github.com/intelsdi-x/snap-plugin-collector-df/issues/10).

Example:
```bash
Filesystem                Type     1024-blocks      Used Available Capacity Mounted on
udev                      devtmpfs     8113640        12   8113628       1% /dev
tmpfs                     tmpfs        1624944      2524   1622420       1% /run
/dev/dm-3                 ext4       103077152  84200980  13617116      87% /
none                      tmpfs         102400        32    102368       1% /run/user
/dev/mapper/data--vg-data ext4       722343000 319310640 371560244      47% /data
/dev/sda1                 ext4         1998672    393128   1484304      21% /boot
/dev/mapper/vms--vg-vms   ext4       309501952 151700240 142056896      52% /vms
```
In above example "Filesystem" is dynamic namespace element.

### Proposed solution
User is able to define custom namespace speparator in task manifest for each metric independently.
```yml
---
  workflow: 
    collect: 
      metrics: 
        /intel/mock/foo: {}
        %intel%procfs%filesystem%*%space_free: {}
```
Metrics requested in collection section would have its own separator defined as in above example. While parsing task manifest, each requested metrics would be split by first encountered character. For `/intel/mock/foo` it would be `/` and for `%intel%procfs%filesystem%*%space_free` it would be `%` 

This would require changes in:
1. `wmap.CollectWorkflowMapNode` function `GetMetrics()` to recognize metric separator and split requested metric accordingly
2. `core.Namespace` function `String()` to detect if any of its `NamespaceElement`s contains `/` character, and if so, use another character from the list of allowed separators.
3. Introduction of priority list of allowed separators in `core` package. This list would be used to provide substitutes for `/` character to allow proper display of metric. 
Example list = `{"/", "%", "#", "_", ":"}. Priority would be resolved by character position on the list. Lower index meaning higher priority.

User defined separator from task manifest would not be preserved though, as there is no need for that.

This solution should be transparent for current plugins (collectors, processors, publishers)

@intelsdi-x/snap-maintainers @obourdon @elemoine

candysmurf:
Fixes #1234 

Summary of changes:
- Added the prioritySeparatorList in core.metric.go
- Uses the first available namespace separator from the priority list when verifying the namespace
- Peeks the first character from the task manifest as the namespace separator.

Testing done:
- small, medium, and large tests
- Integration tests

@intelsdi-x/snap-maintainers

We can add more separators when cases arise.

### DIFF
--- a/core/metric.go
+++ b/core/metric.go
@@ -20,6 +20,7 @@ limitations under the License.
 package core
 
 import (
+	"fmt"
 	"strings"
 	"time"
 
@@ -31,6 +32,7 @@ var (
 	// Standard Tags are in added to the metric by the framework on plugin load.
 	// STD_TAG_PLUGIN_RUNNING_ON describes where the plugin is running (hostname).
 	STD_TAG_PLUGIN_RUNNING_ON = "plugin_running_on"
+	nsPriorityList            = []string{"/", "|", "%", ":", "-", ";", "_", "^", ">", "<", "+", "=", "&", "㊽", "Ä", "大", "小", "ᵹ", "☍", "ヒ"}
 )
 
 // Metric represents a snap metric collected or to be collected
@@ -51,7 +53,8 @@ type Namespace []NamespaceElement
 // the elements of the namespace.  A leading "/" is added.
 func (n Namespace) String() string {
 	ns := n.Strings()
-	return "/" + strings.Join(ns, "/")
+	s := n.getSeparator()
+	return s + strings.Join(ns, s)
 }
 
 // Strings returns an array of strings that represent the elements of the
@@ -62,6 +65,40 @@ func (n Namespace) Strings() []string {
 		ns = append(ns, namespaceElement.Value)
 	}
 	return ns
+}
+
+// getSeparator returns the highest suitable separator from the nsPriorityList.
+// Otherwise the core separator is returned.
+func (n Namespace) getSeparator() string {
+	smap := initSeparatorMap()
+
+	for _, e := range n {
+		// look at each char
+		for _, r := range e.Value {
+			ch := fmt.Sprintf("%c", r)
+			if v, ok := smap[ch]; ok && !v {
+				smap[ch] = true
+			}
+		}
+	}
+
+	// Go through our separator list
+	for _, s := range nsPriorityList {
+		if v, ok := smap[s]; ok && !v {
+			return s
+		}
+	}
+	return Separator
+}
+
+// initSeparatorMap populates the local map of nsPriorityList.
+func initSeparatorMap() map[string]bool {
+	m := map[string]bool{}
+
+	for _, s := range nsPriorityList {
+		m[s] = false
+	}
+	return m
 }
 
 // IsDynamic returns true if there is any element of the namespace which is
--- a/core/metric_test.go
+++ b/core/metric_test.go
@@ -1,0 +1,64 @@
+// +build small
+
+package core
+
+import (
+	"fmt"
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestMetricSeparator(t *testing.T) {
+	tc := getTestCases()
+	Convey("Test namespace separator", t, func() {
+		for _, c := range tc {
+			Convey("namespace "+c.input.String(), func() {
+				firstChar := getFirstChar(c.input.String())
+				So(firstChar, ShouldEqual, c.expected)
+			})
+		}
+	})
+}
+
+// GetFirstChar returns the first character from the input string.
+func getFirstChar(s string) string {
+	firstChar := ""
+	for _, r := range s {
+		firstChar = fmt.Sprintf("%c", r)
+		break
+	}
+	return firstChar
+}
+
+type testCase struct {
+	input    Namespace
+	expected string
+}
+
+// getTestCases tests the namespace and nsPriorityList.
+func getTestCases() []testCase {
+	tcs := []testCase{
+		testCase{
+			input:    NewNamespace("/hello", "/world"),
+			expected: "|",
+		},
+		testCase{
+			input:    NewNamespace("/hello", "/world", "corporate-service|"),
+			expected: "%",
+		},
+		testCase{
+			input:    NewNamespace("/hello", "/world", "|corporate-service%", "monday_to_friday"),
+			expected: ":",
+		},
+		testCase{
+			input:    NewNamespace("/hello", "/world", "corporate-service/%|-_^><+=:;&", "monday_friday", "㊽ÄA小ヒ☍"),
+			expected: "大",
+		},
+		testCase{
+			input:    NewNamespace("A小ヒ☍小㊽%:;", "/hello", "/world大|", "monday_friday", "corporate-service"),
+			expected: "^",
+		},
+	}
+	return tcs
+}
--- a/scheduler/wmap/sample/2.json
+++ b/scheduler/wmap/sample/2.json
@@ -1,0 +1,35 @@
+{
+    "collect": {
+        "metrics": {
+            "㊽foo㊽bar": {
+                "version": 1
+            },
+            "大foo大baz": {},
+            "/a0/b0": {},
+            "-a1-b1": {},
+            "_a2_b2": {},
+            "㊽a3㊽b3": {},
+            "Äa4Äb4": {},
+            "大a5大b5": {},
+            "小a6小b6": {},
+            "ᵹa7ᵹb7": {},
+            "☍a8☍b8": {},
+            "ヒa9ヒb9": {}
+        },
+        "config": {
+            "%foo%bar": {
+                "password": "drowssap",
+                "user": "root"
+            }
+        },
+        "tags": {
+            "㊽foo㊽bar": {
+                "tag1": "val1",
+                "tag2": "val2"
+            },
+            "大foo大baz": {
+                "tag3": "val3"
+            }
+        }
+    }
+}
--- a/scheduler/wmap/wmap.go
+++ b/scheduler/wmap/wmap.go
@@ -215,14 +215,27 @@ func (c *CollectWorkflowMapNode) GetMetrics() []Metric {
 	metrics := make([]Metric, len(c.Metrics))
 	i := 0
 	for k, v := range c.Metrics {
-		ns := strings.Trim(k, `/`)
+		// Identify the character to split on by peaking
+		// at the first character of each metric.
+		firstChar := getFirstChar(k)
+		ns := strings.Trim(k, firstChar)
 		metrics[i] = Metric{
-			namespace: strings.Split(ns, "/"),
+			namespace: strings.Split(ns, firstChar),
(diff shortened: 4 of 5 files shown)


core/metric.go
```diff
@@ -64,6 +67,31 @@ func (n Namespace) Strings() []string {
 	return ns
 }
 
+// getSeparator returns an valid separator from nsPriorityList.
+// Otherwise the core separator is returned.
+func (n Namespace) getSeparator() string {
```
kindermoumoute: Can we make it public?

core/metric.go
```diff
@@ -64,6 +67,31 @@ func (n Namespace) Strings() []string {
...
+	return Separator
+}
+
+// toMap converts namespace values into a character map.
+func (n Namespace) toMap() map[string]bool {
+	m := map[string]bool{}
+	for _, elem := range n {
+		rs := elem.Value
+		for _, r := range rs {
+			s := fmt.Sprintf("%c", r)
```
kindermoumoute: Maybe change `nsPriorityList` to a slice of rune so you don't deal with that line.

core/metric.go
```diff
@@ -64,6 +67,31 @@ func (n Namespace) Strings() []string {
 	return ns
 }
 
+// getSeparator returns an valid separator from nsPriorityList.
+// Otherwise the core separator is returned.
+func (n Namespace) getSeparator() string {
+	mp := n.toMap()
+	for _, r := range nsPriorityList {
+		if ok := mp[r]; !ok {
```
kindermoumoute: Same as`if !mp[r] {`?

core/metric.go
```diff
@@ -64,6 +67,31 @@ func (n Namespace) Strings() []string {
...
+			return r
+		}
+	}
+	return Separator
+}
+
+// toMap converts namespace values into a character map.
+func (n Namespace) toMap() map[string]bool {
```
kindermoumoute: Same, do we want to make this public?

scheduler/wmap/wmap.go
```diff
@@ -215,16 +215,29 @@ func (c *CollectWorkflowMapNode) GetMetrics() []Metric {
...
 		}
 		i++
 	}
 	return metrics
 }
 
+// getFirstChar returns the first character from the input string.
+func getFirstChar(s string) string {
+	firstChar := ""
+	for _, r := range s {
```
kindermoumoute: Why you don't use `s[0]`? Is it because you want to deal with empty strings?

scheduler/wmap/sample/2.json
```diff
@@ -0,0 +1,35 @@
...
+            "/a0/b0": {},
+            "-a1-b1": {},
+            "_a2_b2": {},
+            "㊽a3㊽b3": {},
+            "Äa4Äb4": {},
+            "大a5大b5": {},
+            "小a6小b6": {},
+            "ᵹa7ᵹb7": {},
+            "☍a8☍b8": {},
+            "ヒa9ヒb9": {}
```
kindermoumoute: Maybe add separators that are not in the list?

core/metric_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+			firstChar := ""
+			for _, r := range s {
+				firstChar = fmt.Sprintf("%c", r)
+				So(firstChar, ShouldEqual, "㊽")
+				break
+			}
+		})
+		Convey("namespace has many chars in our separator list", func() {
+			ns := NewNamespace("/hello", "/world", "corporate-service", "monday_friday", "㊽ÄA小ヒ☍")
```
kindermoumoute: Maybe add another case where priority characters are not in the same order than the list:
`"小hello", "/world", "corporate-service", "monday_friday", "㊽ÄAヒ☍"`

core/metric_test.go
```diff
@@ -0,0 +1,44 @@
+package core
+
+import (
+	"fmt"
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestMetricSeparator(t *testing.T) {
```
kindermoumoute: Could this have `input` and `output` test cases instead of hardcoding it?

core/metric.go
candysmurf: I don't see the needs. What's your use case?

core/metric.go
candysmurf: No, why?

core/metric.go
candysmurf: this way is better as we return string for the namespace. I tried rune earlier.

scheduler/wmap/wmap.go
candysmurf: No, we cannot use s[0] here as separators have unicode characeters.

scheduler/wmap/sample/2.json
candysmurf: that's support already. Here is to make sure all chars in the list won't cause issue.

core/metric.go
kindermoumoute: Just thought someone could want to use that for generating a task manifest dynamically. (not a relevant use case)

core/metric.go
```diff
@@ -64,6 +67,31 @@ func (n Namespace) Strings() []string {
 	return ns
 }
 
+// getSeparator returns an valid separator from nsPriorityList.
+// Otherwise the core separator is returned.
+func (n Namespace) getSeparator() string {
+	mp := n.toMap()
```
kindermoumoute: The idea of using a map for through characters is great.

Number of characters from namespaces are variable, and characters change from a namespace to another. What if instead of generating this map we use a static map of the priority characters?

This is globaly the same algorithm, here is an implementation I quickly wrote in a `test.go` file:
```
package main

import (
	"fmt"

)

var (
	nsPriorityList = map[string]bool{"/": false, "-": false, "_": false, "㊽": false, "Ä": false, "大": false, "小": false, "ᵹ": false, "☍": false, "ヒ": false}
)

type Namespace core.Namespace

func main() {
	//ns := core.NewNamespace("p大Älo-p", "ヒwo㊽rld/", "te☍_st小ᵹ")
	//ns := core.NewNamespace("p大Älo-p", "wo㊽rld/", "te_st")
	ns := core.NewNamespace("plop", "world", "test")

	fmt.Println(ns.GetSeparator())
}

// getSeparator returns an valid separator from nsPriorityList.
// Otherwise the core separator is returned.
func (n Namespace) GetSeparator() string {
	for _, elem := range n {
		for _, r := range elem.Value {
			s := fmt.Sprintf("%c", r)
			if _, ok := nsPriorityList[s]; ok && !nsPriorityList[s] {
				nsPriorityList[s] = true
			}
		}
	}
	for sep, used := range nsPriorityList {
		if !used {
			return sep
		}
	}
	return core.Separator
}
```

`go run test.go`

IRCody: This implementation loses order since ranging over a map is not ordered. In the example it's also modifying the global map which means it would only work once.

kindermoumoute: That's a really good point. For solving that I use a slice of `Separator`.
```Go
package main

import (
	"fmt"

)

var (
	sepSortedList = []Separator{
		Separator{Sep: "/"},
		Separator{Sep: "-"},
		Separator{Sep: "_"},
		Separator{Sep: "㊽"},
		Separator{Sep: "Ä"},
		Separator{Sep: "大"},
		Separator{Sep: "小"},
		Separator{Sep: "ᵹ"},
		Separator{Sep: "☍"},
		Separator{Sep: "ヒ"}}

	sepMap map[string]*Separator
)

type Separator struct {
	Sep  string
	Used bool
}

func init() {
	sepMap = make(map[string]*Separator)
	for i, sep := range sepSortedList {
		sepMap[sep.Sep] = &sepSortedList[i]
	}
}

type Namespace core.Namespace

func main() {
	ns := Namespace(core.NewNamespace("p大Älo-p", "ヒwo㊽rld/", "te☍_st小ᵹ"))
	// ns := core.NewNamespace("p大Älo-p", "wo㊽rld/", "te_st")
	// ns := core.NewNamespace("plop", "world", "test")

	fmt.Println(ns.GetSeparator())
}

func (n Namespace) GetSeparator() string {
	for _, elem := range n {
		for _, r := range elem.Value {
			s := fmt.Sprintf("%c", r)
			if sep, ok := sepMap[s]; ok {
				sep.Used = true
			}
		}
	}
	for _, sep := range sepSortedList {
		if !sep.Used {
			return sep.Sep
		}
	}
	return core.Separator
}
```
At the end, it's using a Map of pointer + a slice of Separator (string+bool).

candysmurf:
@jcooklin, @IRCody & @kindermoumoute, I can change to separator character map which may slightly save a bit memory. Algorithm will be the same so the computation complexity will be the same. Note that I prefer to create a separator map each time instead of reusing it. If Cody has any better thoughts, please share. It's important to us the performance and memory usage of Snap! thank you.

utils/misc.go
```diff
@@ -0,0 +1,13 @@
+package utils
```
geauxvirtual: This file should be under `pkg`, not at the root of the repo.

core/metric.go
```diff
@@ -31,6 +32,7 @@ var (
 	// Standard Tags are in added to the metric by the framework on plugin load.
 	// STD_TAG_PLUGIN_RUNNING_ON describes where the plugin is running (hostname).
 	STD_TAG_PLUGIN_RUNNING_ON = "plugin_running_on"
+	nsPriorityList            = []string{"/", "-", "_", "㊽", "Ä", "大", "小", "ᵹ", "☍", "ヒ"}
```
kindermoumoute: We could keep sticks first so it's visually the same:
`/ | \ % # ~ - _ ^ > < + = & $`

candysmurf:
Hi All, please take another look and let me know any suggestions. The backslash didn't use as it constantly thinks as the escaping for the next character.

pkg/utils/misc.go
```diff
@@ -0,0 +1,13 @@
+package utils
```
jcooklin: If we add this a more appropriate 'pkg' name would be "stringutil".  Code under pkg shouldn't have knowledge of snap and as such would be available for general reuse therefore the package name should be as self describing as possible.  

Outside of a test this function is only called once in wmap and with only a single function, in the pkg, it's not very useful to anyone else IMO.  I would prefer that we not break it out into a pkg at this time and we move the functionality back into wmap even if we have to duplicate the code in metric_test.go (since it's the other place it's being called).

Lastly, here is an alternate implementation that avoids the range and fmt.Sprintf.
```
func GetFirstChar(s string) string {
	if len(s) > 0 {
		return s[:1]
	}
	return ""
}
```

candysmurf: changed.

PR #1253, opened 2016-10-04
Fix for issue #1167, added handling of errors which are returned by scaner

### LINKED ISSUE #1167: Errors during capturing of stderr from plugins
It seems that there is an issue with capturing logs from plugin which are sent to stderr. The issue was observed using the cpu plugin. 

I've prepared the special versions of cpu plugin to observe what if plugin sends logs frequently:
- cpu plugin with frequent short log, the version is available [here](https://github.com/katarzyna-z/snap-plugin-collector-cpu/commit/a79c0a040d6308a9276a54d2e6f8cd27581f8b95), 
- cpu plugin with long log, the version is is available [here](https://github.com/katarzyna-z/snap-plugin-collector-cpu/commit/9738a2b2145067d19cb5e3c776ec9d233428b204).

I've tested this two special versions of cpu plugin with two version of snap:
- snap with capturing of logs from plugins, it is this [commit](https://github.com/intelsdi-x/snap/commit/5d91b04fa8b7d398fea04f5cbbe970f21cf32bd9),
- snap without capturing of logs from plugins, it is this [commit]( https://github.com/intelsdi-x/snap/commit/84aa75e8074ecbf633b43d07a96a12aa43f6861a).

I've observed that errors appear only in snap with capturing of logs from plugins. It seems that there is a problem in capturing stderr from plugins when log line is long. In the snap version without capturing of logs from plugins there was a special case for errors returned by the scanner, please see [this](https://github.com/intelsdi-x/snap/commit/5d91b04fa8b7d398fea04f5cbbe970f21cf32bd9#diff-9263d577284097ec1bec3e3fce13eb32L328) .

It could be fixed on plugin side (by adding a newline at the end of the log) but it will be good to have a protection also on snap side.

katarzyna-z:
Fixes #1167 

Hanging of plugins during capturing of stderr and stdout were observed when log line had a significant length. It was caused by errors occured during scanning of stderr or stdout. Errors returned by Scan()  were not handled by snapd. 

To fix the issue used the method which was used before adding of log capturing from plugin [commit]( https://github.com/intelsdi-x/snap/commit/5d91b04fa8b7d398fea04f5cbbe970f21cf32bd9) so now goto procedure is in use to deal with long line of logs.

Summary of changes:
- Added handling of errors returned by Scan method in the part of code which is used to capture stdout and stderr from plugins,
- Modified mock plugin to test log lines with significant length.

Testing done:
- Verified tests pass for small, medium, and legacy.
- Manual tests with mock plugin.

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/plugin/execution.go
+++ b/control/plugin/execution.go
@@ -115,21 +115,41 @@ func (e *ExecutablePlugin) Run(timeout time.Duration) (Response, error) {
 	e.cmd.Start()
 	e.captureStderr()
 	go func() {
-		for stdOutScanner.Scan() {
-			// The first chunk from the scanner is the plugin's response to the
-			// handshake.  Once we've received that, we can begin to forward
-			// logs on to snapd's log.
-			if !respReceived {
-				respBytes := stdOutScanner.Bytes()
-				err = json.Unmarshal(respBytes, &resp)
-				respReceived = true
-				close(doneChan)
-			} else {
-				execLogger.WithFields(log.Fields{
-					"plugin": path.Base(e.cmd.Path()),
-					"io":     "stdout",
-				}).Debug(stdOutScanner.Text())
+		for {
+			for stdOutScanner.Scan() {
+				// The first chunk from the scanner is the plugin's response to the
+				// handshake.  Once we've received that, we can begin to forward
+				// logs on to snapd's log.
+				if !respReceived {
+					respBytes := stdOutScanner.Bytes()
+					err = json.Unmarshal(respBytes, &resp)
+					respReceived = true
+					close(doneChan)
+				} else {
+					execLogger.WithFields(log.Fields{
+						"plugin": path.Base(e.cmd.Path()),
+						"io":     "stdout",
+					}).Debug(stdOutScanner.Text())
+				}
 			}
+
+			if errScanner := stdOutScanner.Err(); errScanner != nil {
+				reader := bufio.NewReader(e.stdout)
+				log, errRead := reader.ReadString('\n')
+				if errRead == io.EOF {
+					break
+				}
+
+				execLogger.
+					WithField("plugin", path.Base(e.cmd.Path())).
+					WithField("io", "stdout").
+					WithField("scanner_err", errScanner).
+					WithField("read_string_err", errRead).
+					Warn(log)
+
+				continue //scanner finished with errors so try to scan once again
+			}
+			break //scanner finished scanning without errors so break the loop
 		}
 	}()
 
@@ -167,12 +187,31 @@ func (e *ExecutablePlugin) Kill() error {
 func (e *ExecutablePlugin) captureStderr() {
 	stdErrScanner := bufio.NewScanner(e.stderr)
 	go func() {
-		for stdErrScanner.Scan() {
-			execLogger.
-				WithField("io", "stderr").
-				WithField("plugin", path.Base(e.cmd.Path())).
-				Debug(stdErrScanner.Text())
+		for {
+			for stdErrScanner.Scan() {
+				execLogger.
+					WithField("plugin", path.Base(e.cmd.Path())).
+					WithField("io", "stderr").
+					Debug(stdErrScanner.Text())
+			}
 
+			if errScanner := stdErrScanner.Err(); errScanner != nil {
+				reader := bufio.NewReader(e.stderr)
+				log, errRead := reader.ReadString('\n')
+				if errRead == io.EOF {
+					break
+				}
+
+				execLogger.
+					WithField("plugin", path.Base(e.cmd.Path())).
+					WithField("io", "stderr").
+					WithField("scanner_err", errScanner).
+					WithField("read_string_err", errRead).
+					Warn(log)
+
+				continue //scanner finished with errors so try to scan once again
+			}
+			break //scanner finished scanning without errors so break the loop
 		}
 	}()
 }
--- a/plugin/collector/snap-plugin-collector-mock2/mock/mock.go
+++ b/plugin/collector/snap-plugin-collector-mock2/mock/mock.go
@@ -20,9 +20,11 @@ limitations under the License.
 package mock
 
 import (
+	"bufio"
 	"fmt"
 	"log"
 	"math/rand"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


control/plugin/execution.go
```diff
@@ -131,6 +132,25 @@ func (e *ExecutablePlugin) Run(timeout time.Duration) (Response, error) {
 				}).Debug(stdOutScanner.Text())
 			}
 		}
+
+		if err := stdOutScanner.Err(); err != nil {
+			if err == bufio.ErrTooLong {
+				reader := bufio.NewReader(e.stdout)
+				log, _, _ := reader.ReadLine()
```
IRCody: Why `Readline` vs `ReadString('\n')`?

control/plugin/execution.go
```diff
@@ -115,6 +115,7 @@ func (e *ExecutablePlugin) Run(timeout time.Duration) (Response, error) {
 	e.cmd.Start()
 	e.captureStderr()
 	go func() {
+	OK:
```
IRCody: I think this would be clearer as a loop vs a goto.

plugin/collector/snap-plugin-collector-mock2/mock/mock.go
```diff
@@ -56,14 +58,24 @@ func (f *Mock) CollectMetrics(mts []plugin.MetricType) ([]plugin.MetricType, err
...
 				longLine = append(longLine, letterBytes[rand.Intn(len(letterBytes))])
 			}
-			fmt.Println(string(longLine))
+			fmt.Fprintln(os.Stdout, string(longLine))
 		}
+
+		if c, ok := mts[i].Config().Table()["long_stderr_log"]; ok && c.(ctypes.ConfigValueBool).Value {
+			letterBytes := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+			longLine := []byte{}
+			for i := 0; i < bufio.MaxScanTokenSize; i++ {
```
IRCody: Is the goal here to append to longline until it is larger than the MaxScanTokenSize?

control/plugin/execution.go
katarzyna-z: I don't see any special reasons to use `Readline()` instead of `ReadString('\n')`. I changed `Readline()`to `ReadString('\n')` to avoid parsing to string.

control/plugin/execution.go
katarzyna-z: OK, I changed it, now loops are in use.

plugin/collector/snap-plugin-collector-mock2/mock/mock.go
katarzyna-z: Here the goal is to have `longline` with length equal to `MaxScanTokenSize`. The condition in `Scan()` method has following form:
```
if len(s.buf) >= s.maxTokenSize || len(s.buf) > maxInt/2 {
				s.setErr(ErrTooLong)
				return false
}
```

control/plugin/execution.go
```diff
@@ -115,20 +115,35 @@ func (e *ExecutablePlugin) Run(timeout time.Duration) (Response, error) {
...
+				reader := bufio.NewReader(e.stdout)
+				log, err := reader.ReadString('\n')
+
+				execLogger.
+					WithField("plugin", path.Base(e.cmd.Path())).
+					WithField("io", "stdout").
+					WithField("scanner_err", stdOutScanner.Err()).
+					WithField("read_string_err", err).
+					Warn(log)
+				continue
```
IRCody: I don't think this continue is needed.

Also there is still a case where we will want to break out of the loop right? Otherwise this goroutine will never exit?

katarzyna-z: I made a small correction in code to finish the loop after scanning. Currently, I don't see solution without `continue`. Please take a look at this. If you have another idea, I'm open for your suggestions.

IRCody: In the iteration I was commenting on the continue served no purpose as there wasn't code executed after it. The update to add break below it changes that behavior.

control/plugin/execution.go
```diff
@@ -167,12 +183,27 @@ func (e *ExecutablePlugin) Kill() error {
...
+			for stdErrScanner.Scan() {
+				execLogger.
+					WithField("plugin", path.Base(e.cmd.Path())).
+					WithField("io", "stderr").
+					Debug(stdErrScanner.Text())
+			}
 
+			if err := stdErrScanner.Err(); err != nil {
+				reader := bufio.NewReader(e.stderr)
+				log, err := reader.ReadString('\n')
```
jcooklin: we need to handle the err from `ReadString`

```
log, err2 := reader.ReadString('\n')
				if err2 == io.EOF {
					break
				}
```
Without the above block when the plugin stops the go routine never exits.  You can see this by loading the modified mock2 plugin and starting the following task and then finally stop (ctrl-c) snapd.
```
---
  version: 1
  schedule: 
    type: "simple"
    interval: "1s"
  max-failures: 10
  workflow: 
    collect: 
      metrics: 
        /intel/mock/foo: {}
        /intel/mock/bar: {}        
      config: 
        /intel/mock: 
          name: "root"
          password: "secret"          
          long_stderr_log: true
```
img

control/plugin/execution.go
```diff
@@ -167,12 +183,27 @@ func (e *ExecutablePlugin) Kill() error {
...
+
+				execLogger.
+					WithField("plugin", path.Base(e.cmd.Path())).
+					WithField("io", "stderr").
+					WithField("scanner_err", stdErrScanner.Err()).
+					WithField("read_string_err", err).
```
jcooklin: No need to called stdErrScanner.Err() twice (use err).

control/plugin/execution.go
```diff
@@ -115,21 +115,37 @@ func (e *ExecutablePlugin) Run(timeout time.Duration) (Response, error) {
...
+					execLogger.WithFields(log.Fields{
+						"plugin": path.Base(e.cmd.Path()),
+						"io":     "stdout",
+					}).Debug(stdOutScanner.Text())
+				}
 			}
+
+			if err := stdOutScanner.Err(); err != nil {
+				reader := bufio.NewReader(e.stdout)
+				log, err := reader.ReadString('\n')
```
jcooklin: Same comments made below on the stderr routine apply here (re: handle ReadString error).

katarzyna-z:
Thanks for review. I made a correction. Could you check it?

IRCody:
LGTM. @jcooklin?

PR #1255, opened 2016-10-04
Created V2 REST API Routes

### LINKED ISSUE #634: In the route `"/v1/tribe/member/:name"` members should be plural
The route `"/v1/tribe/member/:name"` should be `"/v1/tribe/members/:name"`

Initially reported by @mingchan [here](https://github.com/intelsdi-x/snap/issues/633).

### LINKED ISSUE #730: Metric API to return a named list (metrics_returned) in the body?
Just going through the return for “GET /v1/metrics” call.

Should it return a named list like below instead?

```
“body”: {
    “metrics_returned”: [...]
}
```

So that it is consistent with other API calls like

```
GET /v1/plugins
GET /v1/tasks
```

They return “loaded_plugins” and “ScheduledTasks” lists respectively.

kjlyon:
Fixes #634 and #730 as well as other V2 API changes

Summary of changes:
- Fixed named list returned by metric API
- Fixed text case in V2/tests from "ScheduledTasks" to "scheduled_tasks"
- Made all V2 tribe routes to be /V2/tribes (made tribe plural)
- Changed V2/tribes/member/:name to  be V2/tribes/members/:name (made member plural)
- Added V2 folders in rest and in rbody to hold V2 methods

Note: 
- I will add additional issues for updating the documentation and writing V2 tests. 

Testing done:
- Unit tests

@intelsdi-x/snap-maintainers

### DIFF
--- a/mgmt/rest/fixtures/mock_tribe_manager.go
+++ b/mgmt/rest/fixtures/mock_tribe_manager.go
@@ -35,6 +35,19 @@ var (
 )
 
 func init() {
+	mockTribeMember = agreement.NewMember(&memberlist.Node{
+		Name: "Imma_Mock",
+		Addr: net.ParseIP("193.11.22.11"),
+		Port: uint16(0),
+		Meta: []byte("meta"), // Metadata from the delegate for this node.
+		PMin: uint8(0),       // Minimum protocol version this understands
+		PMax: uint8(0),       // Maximum protocol version this understands
+		PCur: uint8(0),       // Current version node is speaking
+		DMin: uint8(0),       // Min protocol version for the delegate to understand
+		DMax: uint8(0),       // Max protocol version for the delegate to understand
+		DCur: uint8(0),       // Current version delegate is speaking
+	})
+
 	mockTribeAgreement = agreement.New("Agree1")
 	mockTribeAgreement.PluginAgreement.Add(
 		agreement.Plugin{Name_: "mockVersion", Version_: 1, Type_: core.CollectorPluginType})
@@ -80,7 +93,7 @@ func (m *MockTribeManager) GetMembers() []string {
 	return []string{"one", "two", "three"}
 }
 func (m *MockTribeManager) GetMember(name string) *agreement.Member {
-	return &agreement.Member{}
+	return mockTribeMember
 }
 
 // These constants are the expected tribe responses from running
@@ -209,7 +222,7 @@ const (
     "version": 1
   },
   "body": {
-    "name": "",
+    "name": "Imma_Mock",
     "plugin_agreement": "",
     "tags": null,
     "task_agreements": null
--- a/mgmt/rest/metric.go
+++ b/mgmt/rest/metric.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/intelsdi-x/snap/core"
 	"github.com/intelsdi-x/snap/mgmt/rest/rbody"
+	apiV2 "github.com/intelsdi-x/snap/mgmt/rest/v2"
 )
 
 func (s *Server) getMetrics(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
@@ -40,6 +41,7 @@ func (s *Server) getMetrics(w http.ResponseWriter, r *http.Request, _ httprouter
 	// perform a query
 	q := r.URL.Query()
 	v := q.Get("ver")
+	apiVersion := r.URL.Path[1:3]
 	ns_query := q.Get("ns")
 	if ns_query != "" {
 		ver = 0 // 0: get all versions
@@ -62,7 +64,7 @@ func (s *Server) getMetrics(w http.ResponseWriter, r *http.Request, _ httprouter
 			respond(404, rbody.FromError(err), w)
 			return
 		}
-		respondWithMetrics(r.Host, mets, w)
+		respondWithMetrics(r.Host, mets, w, apiVersion)
 		return
 	}
 
@@ -71,7 +73,7 @@ func (s *Server) getMetrics(w http.ResponseWriter, r *http.Request, _ httprouter
 		respond(500, rbody.FromError(err), w)
 		return
 	}
-	respondWithMetrics(r.Host, mets, w)
+	respondWithMetrics(r.Host, mets, w, apiVersion)
 }
 
 func (s *Server) getMetricsFromTree(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
@@ -94,6 +96,7 @@ func (s *Server) getMetricsFromTree(w http.ResponseWriter, r *http.Request, para
 	)
 	q := r.URL.Query()
 	v := q.Get("ver")
+	apiVersion := r.URL.Path[1:3]
 
 	if ns[len(ns)-1] == "*" {
 		if v == "" {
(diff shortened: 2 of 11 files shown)


mgmt/rest/metric.go
```diff
@@ -40,6 +41,7 @@ func (s *Server) getMetrics(w http.ResponseWriter, r *http.Request, _ httprouter
 	// perform a query
 	q := r.URL.Query()
 	v := q.Get("ver")
+	apiVersion := r.URL.Path[1:3]
```
kindermoumoute: Maybe add a comment to explain it'll catch "v1", "v2". 

Do we want to handle an error if something is different than "v1" or "v2?

mgmt/rest/rbody/v2/metric.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type MetricsReturned MetricList
+
+func (m MetricsReturned) Len() int {
+	return len(m.Metrics)
+}
+
+func (m MetricsReturned) Less(i, j int) bool {
+	return (fmt.Sprintf("%s:%d", m.Metrics[i].Namespace, m.Metrics[i].Version)) < (fmt.Sprintf("%s:%d", m.Metrics[j].Namespace, m.Metrics[j].Version))
```
kindermoumoute: Do you need the column `:`?

mgmt/rest/v2/metric.go
```diff
@@ -0,0 +1,92 @@
...
+	"net/url"
+	"sort"
+
+	"github.com/codegangsta/negroni"
+	"github.com/intelsdi-x/snap/core"
+	"github.com/intelsdi-x/snap/mgmt/rest/rbody"
+	"github.com/intelsdi-x/snap/mgmt/rest/rbody/v2"
+)
+
+func RespondWithMetrics(host string, mets []core.CatalogedMetric, w http.ResponseWriter) {
```
kindermoumoute: Just thought it would make sense to use same variable name. I didn't see "mets" before, AFAIK "metrics" or "mts" is more common, with the singular "metric" or "m".

mgmt/rest/v2/metric.go
```diff
@@ -0,0 +1,92 @@
...
+	b := v2.NewMetricsReturned()
+
+	for _, met := range mets {
+		rt := met.Policy().RulesAsTable()
+		policies := make([]rbody.PolicyTable, 0, len(rt))
+		for _, r := range rt {
```
kindermoumoute: I feel there is a duplicate of code with https://github.com/intelsdi-x/snap/blob/67417ef0fca51def9cf0eaf002d0dc42c0aee842/control/plugin/cpolicy/node.go#L131-L147

It's better to:
- Create a method `RulesAsPolicyTable`.
- Or use the type `cpolicy.RuleTable` instead of `rbody.PolicyTable`, those type are exactly the same at 2 different places.

mgmt/rest/metric.go
kjlyon: I'll add a comment to address that. If it gets something back other than "v2" it will default to v1. I think that will be okay.

mgmt/rest/rbody/v2/metric.go
kjlyon: No, we don't need column in rbody/metric.go, so we wouldn't need it here either. Even though it is part of the MetricsReturned object, we never actually call this method in our mock testing.

mgmt/rest/v2/metric.go
kjlyon: ok, I'll make that change

mgmt/rest/metric.go
kindermoumoute: Gotcha, specify that in the comment too.

mgmt/rest/rbody/v2/metric.go
kindermoumoute: I guess it was for sorting purpose, but I don't see any import of the package `sort`. It should be something like `sort.Sort(MyMetricsReturned)`.

mgmt/rest/fixtures/mock_tribe_manager.go
```diff
@@ -209,7 +222,7 @@ const (
     "version": 1
   },
   "body": {
-    "name": "",
+    "name": "Imma_Mock",
```
IRCody: 👍

mgmt/rest/metric.go
```diff
@@ -171,7 +174,11 @@ func (s *Server) getMetricsFromTree(w http.ResponseWriter, r *http.Request, para
 	respond(200, b, w)
 }
 
-func respondWithMetrics(host string, mets []core.CatalogedMetric, w http.ResponseWriter) {
+func respondWithMetrics(host string, mets []core.CatalogedMetric, w http.ResponseWriter, version string) {
+	if version == "v2" {
+		apiV2.RespondWithMetrics(host, mets, w)
+		return
+	}
```
kjlyon: So you're saying, rather than checking if version == v2, I should check if version == v1 and the default would be v2 functionality? Is that correct?

IRCody: I was thinking it'd be cleaner to check for each version and route the request to the appropriate version handler here and not have this function have any implementation itself. What would the implications of that be?

kjlyon: I don't think that level of partitioning is necessary yet. Maybe once we get a lot of V2 methods or even a V3 to some of these routes we could create a V1 package and then call the specified method from that package. I don't think we need that yet though.

mgmt/rest/server.go
```diff
@@ -470,8 +470,13 @@ func (s *Server) addRoutes() {
 	s.r.GET("/v1/metrics", s.getMetrics)
 	s.r.GET("/v1/metrics/*namespace", s.getMetricsFromTree)
 
+	s.r.GET("/v2/metrics", s.getMetrics)
+	s.r.GET("/v2/metrics/*namespace", s.getMetricsFromTree)
+
 	// task routes
 	s.r.GET("/v1/tasks", s.getTasks)
+	s.r.GET("/v2/tasks", s.getTasks)
```
kjlyon: It's not necessary to.

IRCody: If you were programming against it wouldn't that create a weird case where the v2 api isn't complete? What happens if I make a request to a v2 route that doesn't exist? Does it get routed down to an earlier version?

mgmt/rest/metric.go
IRCody: I would disagree. I think now (when we are adding the first additional version) is the appropriate time to think about how future versions will be added and how it can be done cleanly.

kjlyon: Are you suggesting to add a v1 package? or just create v1 methods in the same files, and those methods are called by the functions the routes point to?

mgmt/rest/v2/metric.go
kjlyon: Good catch! Fixed this.

jcooklin:
I'd like to propose that we create a "v2 rest api" branch for these changes with the intent being that we hold off on adding a new version until we have a stronger need to version the API.  These changes are all broadly cosmetic and I anticipate that in the near future we may have more compelling reasons to version the API.  Thoughts?

kjlyon:
I'm closing this PR to replace it with #1272 which merges these changes into a feature branch in intelsdi-x/snap/v2_rest_api

PR #1257, opened 2016-10-05
(SDI-2036) Fix s3 output path to snap

nanliu:
We need a minor change so the s3 sub directory is snap instead of
snap/snap.


nanliu:
So #1257 travis CI test ran before #1239 was merged. Due to new changes in #1239 we have additional medium tests that needs to be resolved in the master branch.

PR #1263, opened 2016-10-06
(maint) update plugins list

nanliu:
* Remove jmx plugin
* Add snmp plugin
* Reorg plugin list so it's easier to track changes

### DIFF
--- a/docs/plugins.yml
+++ b/docs/plugins.yml
@@ -15,18 +15,15 @@
 # After merging, we run a job to update PLUGIN_CATALOG.md and
 # the list available on our website: http://snap-telemetry.io/plugins.html
 #####################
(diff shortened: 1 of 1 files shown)


danielkza:
I'm looking to use snap to monitor Java services, and a JMX plugin would be fundamental for my use case. Does the previously linked plugin not work well, or is there another method to acquire JMX metrics?

nanliu:
@danielkza, the repo is still there, but I was under the impression the plugin is not in a working state: https://github.com/Staples-Inc/snap-plugin-collector-jmx

You are welcome to try it and let us know if we should include it in the plugin list.

danielkza:
@nanliu I see. I'll evaluate it and report if I have any success. Thank you!

PR #1266, opened 2016-10-07
Updated documentation API documentation in response to GH#749:

### LINKED ISSUE #749: Improve API documentation based on #633 dicussion
There's a long conversation on state, endpoints and their differentiation. This needs to be captured, probably within docs/REST_API.md and with accompanying blog posts exploring this.

kjlyon:
Fixes #749 

Summary of changes:
-   In REST_API.md added more info on what `watch` means in the tasks section
-   In TASKS.md added state information

Testing done:
- Unit tests
- Previewing document

@intelsdi-x/snap-maintainers

- In REST_API.md, added more info to what `watch` means in tasks section
- In TASKS.md, added state information

### DIFF
--- a/docs/REST_API.md
+++ b/docs/REST_API.md
@@ -449,7 +449,7 @@ _**Example Response**_
 }
 ```
 **GET /v1/tasks/:id/watch**:
-Watch a task activity stream given a task ID
+Watch a task activity stream given a task ID. Watch is an event stream sent over a long running HTTP connection.
 
 _**Example Request**_
 ```
--- a/docs/TASKS.md
+++ b/docs/TASKS.md
@@ -1,11 +1,37 @@
 Tasks
 =====
 
-A task describes the how, what, and when to do for a Snap job.  A task is described in a task _manifest_, which can be either JSON or YAML<sup>1</sup>.
+A task describes the how, what, and when to do for a Snap job.
 
 _Skip to the TL;DR example [here](#tldr)_.
+ 
 
-The manifest can be divided into two parts: Header and Workflow.
+## Task States
+
+A task can be in the following states:
+- **running:** a running task
+- **stopped:** a task that is not running
+- **disabled:** a task in a state not allowed to start. This happens when the task produces consecutive errors. A disabled task must be re-enabled before it can be started again. 
+
+
+![newtaskstatediagram2](https://cloud.githubusercontent.com/assets/21182867/19282545/a4179520-8fa3-11e6-9056-4fc3aa610983.png)
+
+
+	    How To				                        |  Command
+    ----------------------------------------|------------------------
+      Create task                           |  snapctl task create _[command options] [arguments...]_ <br/>  Find more details [here](https://github.com/intelsdi-x/snap/blob/master/docs/SNAPCTL.md#task)
+      List					                        |  snapctl task list
+      Start task                     		    |  snapctl task start _\<task_id>_
+      Stop task                        		  |  snapctl task stop _\<task_id>_
+      Remove task                       		|  snapctl task remove _\<task_id>_
+      Export task                       		|  snapctl task export _\<task_id>_
+      Watch task                        		|  snapctl task watch _\<task_id>_
+      Enable task                       		|  snapctl task enable _\<task_id>_
+
+
+## Task Manifest
+
+A task is described in a task _manifest_, which can be either JSON or YAML<sup>1</sup>. The manifest is divided into two parts: Header and Workflow.
 
 ### The Header
 


docs/TASKS.md
```diff
@@ -3,6 +3,27 @@ Tasks
 
 A task describes the how, what, and when to do for a Snap job.  A task is described in a task _manifest_, which can be either JSON or YAML<sup>1</sup>.
 
+A task can be in the following states:
+- **started:** a running task
+- **stopped:** a task that is not running
+- **disabled:** a task in a state not allowed to start
```
lynxbat: Is this not allowed to start? Or is it "stopped by the Snap"? This question is more for @intelsdi-x/snap-maintainers

docs/REST_API.md
```diff
@@ -449,7 +449,7 @@ _**Example Response**_
 }
 ```
 **GET /v1/tasks/:id/watch**:
-Watch a task activity stream given a task ID
+Watch a task activity stream given a task ID. Watch is an event stream sent overa long running HTTP connection
```
lynxbat: s/overa/over a/

lynxbat: Needs a period on the end.

docs/TASKS.md
kjlyon: Isn't it both? - Not allowed to start because it was stopped by Snap

docs/REST_API.md
kjlyon: Many other sections like this do not have periods at the end. But because this one has two sentences maybe it should end with a period. I'll add that.

docs/TASKS.md
IRCody: A task is disabled on error (by snap) and is not resumable directly. It can be re-enabled which moves it from disabled-> stopped and then is able to be started, iirc. I think phrasing something about it being disabled on error and then requiring the task to be explicitly re-enabled before being started would be good. 

Do we actually have a started state? Looking at the code [here](https://github.com/intelsdi-x/snap/blob/master/core/task.go#L50-L55) it seems like that would be called running. I think this is what would be reported by the API as the state of the task. Would it make sense to stick to those labels then?

The ones I think we'd want to talk about are: 
* running 
* disabled
* stopped
* ended

kjlyon: I updated the PR to reflect these changes.

docs/REST_API.md
```diff
@@ -449,7 +449,7 @@ _**Example Response**_
 }
 ```
 **GET /v1/tasks/:id/watch**:
-Watch a task activity stream given a task ID
+Watch a task activity stream given a task ID. Watch is an event stream sent overa long running HTTP connection.
```
IRCody: need a space in `overa`

docs/TASKS.md
```diff
@@ -1,11 +1,31 @@
 Tasks
 =====
 
-A task describes the how, what, and when to do for a Snap job.  A task is described in a task _manifest_, which can be either JSON or YAML<sup>1</sup>.
+A task describes the how, what, and when to do for a Snap job.  A task is described in a task _manifest_, which can be either JSON or YAML<sup>1</sup>. The manifest can be divided into two parts: Header and Workflow.
```
IRCody: I would say it is divided into two parts, not can be. It's a little confusing to me that the part directly below this isn't going into describing either the Header or the Workflow, maybe move this down below this section and add a header here about "Task states".

docs/TASKS.md
```diff
@@ -1,11 +1,31 @@
...
+
+A task can be in the following states:
+- **running:** a running task
+- **stopped:** a task that is not running
+- **disabled:** a task in a state not allowed to start. This happens when the task produces consecutive errors. A disabled task must be re-enabled before it can be started again. 
+
+
+![newtaskstatediagram](https://cloud.githubusercontent.com/assets/21182867/19250638/4d78384a-8ef0-11e6-91db-2bd9288cf9b5.png)
```
IRCody: Since tasks can be created into the stopped state, should we represent that on this diagram?

docs/TASKS.md
```diff
@@ -1,11 +1,31 @@
...
 
-_Skip to the TL;DR example [here](#tldr)_.
 
-The manifest can be divided into two parts: Header and Workflow.
+	    Commands				                      |  How To
```
IRCody: I like this table but it doesn't feel like it fits with the flow right here.

docs/TASKS.md
```diff
@@ -1,11 +1,31 @@
...
+      Start task                     		    |  snapctl task start _\<task_id>_
+      Stop task                        		  |  snapctl task stop _\<task_id>_
+      Remove task                       		|  snapctl task remove _\<task_id>_
+      Export task                       		|  snapctl task export _\<task_id>_
+      Watch task                        		|  snapctl task watch _\<task_id>_
+      Enable task                       		|  snapctl task enable _\<task_id>_
+
+
+
+_Skip to the TL;DR example [here](#tldr)_.
```
IRCody: This should probably go back towards the top.

docs/TASKS.md
kjlyon: I made some changes to the layout, moved the tldr back to the top, and updated the state diagram. Let me know what you think of these changes.

PR #1269, opened 2016-10-10
Cleaning of examples/ folder

### LINKED ISSUE #1158: Clarify scope and purpose of examples/ folder

katarzyna-z:
Fixes #1158 

Summary of changes:
- Removed the following from the /examples folder:
  - /influxdb-grafana,
  - /riemann,
  - /tasks,
- Removed the reference to influxdb-grafana, riemann, tasks in examples/readme.md
- Added the reference to plugins repositories to readme.md
- Removed the reference to examples/tasks/ in docs/TASKS.md
- Removed the reference to examples/tasks/ in core/task_medium_test.go
- Removed the reference to examples/tasks/ in readme.md
- Removed the reference to examples/tasks/ in docs/SNAPCTL.md
- Removed the reference to examples/tasks/ in docs/REST_API.md
- Replaced "snap" to "Snap" in docs/TASKS.md

@intelsdi-x/snap-maintainers


core/task_medium_test.go
```diff
@@ -72,8 +145,16 @@ func okRoutine(sch schedule.Schedule,
 	return nil, nil
 }
 
-func TestUnmarshalBodyTask(t *testing.T) {
+func createTaskFile(name string, content []byte) {
+	f, _ := os.Create(name)
```
kindermoumoute: Can you handle errors?

docs/TASKS.md
```diff
@@ -110,7 +110,51 @@ Process and Publish nodes in the workflow can also target remote snap nodes via
```
kindermoumoute: Can't comment further but uppercase is missing on "**s**nap" [line 43](https://github.com/intelsdi-x/snap/pull/1269/files#diff-5710b592bec82bde9e7f0bae0a5bbe58R43).

core/task_medium_test.go
```diff
@@ -72,8 +145,16 @@ func okRoutine(sch schedule.Schedule,
...
+	f.Write(content)
+}
 
+func deleteTaskFile(name string) {
+	os.Remove(name)
```
kindermoumoute: Can you handle errors here too?

IRCody:
What is the reasoning behind removing the example tasks?

kindermoumoute:
> What is the reasoning behind removing the example tasks?
Examples already exist in plugin repos.

IRCody:
>>What is the reasoning behind removing the example tasks?

>Examples already exist in plugin repos.

This doesn't seem particular compelling imo.

core/task_medium_test.go
katarzyna-z: I added handling of this error. Please check if it is ok now.

docs/TASKS.md
katarzyna-z: Corrected ;)

core/task_medium_test.go
katarzyna-z: Added handling of this error.

IRCody:
To clarify I'm speaking only about the stuff under examples/tasks that dealt mostly with mock plugins, not the ones that were specific to influx/riemann. I think it's a bad idea to remove all example tasks from the snap repo without replacing them with something.

jcooklin:
@IRCody the thinking was that we didn't want to encourage using mock plugins so we felt link to examples that included proper run scripts was the best approach.

IRCody:
@jcooklin: Maybe I am an outlier but I frequently use the example plugins as "templates" for creating tasks. With this change you'd have to either start from scratch on creating a task, or locate a plugin repo that has examples. From my viewpoint neither of those are that great.

In the issue this is addressing, #1158, I brought up what I thought should go in there based on this thought. Maybe I'm the only one who feels that way. Below is the relevant snippet from my comment on that issue.

>> 2. How should the tasks/ folder differ from example tasks in each plugin repository?

>I think a differentiation between the tasks/ folder in snap and the example tasks in the plugin repos should be that the tasks/ folder in snap focus on what different sorts of things you can do with tasks. An example tasks in a plugin repo should focus on that specific plugins options/etc. Some that come to mind:

>* Collect->Process->Publish -- the "standard" workflow
* Collect-> Publish (no process) -- Might be useful to ensure that people know you can go from collect directly to publish.
* Collect->Publish, Collect->Process->Publish -- display a forking of the workflow that demonstrates collecting raw metrics, as well as publishing processed metrics.
* distributed workflow example
* Example of config on all of {collect,process,publish}

candysmurf:
I often reference example configs and takes in this repo as well. If examples are out date and we have no intention to keep them up. Leaving example configs and tasks could be useful

IzabellaRaulin:
I understand that we don't want to point to mock plugins, however how to present snap features as using query (wildcard, tuple) or specified dynamic metric or tagging? Already these task manifests can be used to check how it works:
- [mock-file_specific_instance.json](https://github.com/intelsdi-x/snap/blob/master/examples/tasks/mock-file_specific_instance.json)
- [mock-file_tuple.json](https://github.com/intelsdi-x/snap/blob/master/examples/tasks/mock-file_tuple.json)
- [mock_tagged-file.json](https://github.com/intelsdi-x/snap/blob/master/examples/tasks/mock_tagged-file.json)
- [distributed-mock-file.json](https://github.com/intelsdi-x/snap/blob/master/examples/tasks/distributed-mock-file.json)

**Question1:** What do you think about keeping these files under "/snap/example/task/mock" or  "/snap/example/mock-task"or sth like that to highlight that is _only for mock plugin_?  

**Question2:** What do you think about keeping one of existing exemplary task manifest as a template? 

**Question3:**  (additional) What do you think about showing the following possibilities:
- skip processing phase (metrics -> publisher)
```
metrics -> publisher
```
- publishing to many places 
```
                      -> publisher1
                     /
metrics -> processor 
                     \
                      -> publisher2
```                                                                               
- chaining metrics through processors in series: 
```
metrics -> processor1 -> processor2 -> publisher
```
- chaining metrics through processors in parallel:
```       
       -> processor1 -> publisher
      /
metrics 
      \
       -> processor2 -> publisher
```

IRCody:
>Question1: What do you think about keeping these files under "/snap/example/task/mock" or "/snap/example/mock-task"or sth like that to highlight that is only for mock plugin?

>Question2: What do you think about keeping one of existing exemplary task manifest as a template?

After thinking about it more, I think I can get behind removing the example tasks. From my perspective it might change a little bit of my workflow but for users/newcomers it would be better not to clutter them with examples that use plugins that are written for testing purposes. In general it's probably good to avoid optimizing to much for our workflows. 

>Question3: (additional) What do you think about showing the following possibilities:

>skip processing phase (metrics -> publisher)
metrics -> publisher
publishing to many places
                      -> publisher1
                     /
metrics -> processor 
                     \
                      -> publisher2
chaining metrics through processors in series:
metrics -> processor1 -> processor2 -> publisher
chaining metrics through processors in parallel:
       -> processor1 -> publisher
      /
metrics 
      \
       -> processor2 -> publisher

I think examples for this stuff is important, but the more I think about it the more I think it should live inside somewhere like docs/Tasks.md with really solid explanations about what is happening. 

>Question4: Why some of examples have been moved from snap to snap-plugin repo without removing "mock" from their name? There is still "mock" in files' names, see https://github.com/intelsdi-x/snap-plugin-collector-disk/tree/master/examples/tasks. I think that is confusing.

I agree and I think we should correct these where we find them.

mjbrender:
TL;DR - Assume Task Manifests are for Operations engineers and do not correlate with our testing strategy. 

From @IRCody: 

> After thinking about it more, I think I can get behind removing the example tasks. From my perspective it might change a little bit of my workflow but for users/newcomers it would be better not to clutter them with examples that use plugins that are written for testing purposes. In general it's probably good to avoid optimizing to much for our workflows.

I couldn't agree more. Work on this folder can get us a little clearer on the separation of **testing** and **examples**. Testing can act as a great basis for contributors to Snap, but it's not safe to assume every user is a Go developer and will correlate the mock plugin design with the framework. The more audience-specific our examples can get, the better they are.

core/task_medium_test.go
kindermoumoute: The function should return the error, so you can test if error is `nil` with the convey framework.

docs/TASKS.md
kindermoumoute: Same than above.

candysmurf:
Are you guys proposing updating xxx.md file to embedded sample configs and task manifests?  Although plugins have those sample task manifests but they could be out of sync when Snap changes.   

Snap seems an ideal place for the latest sample configs and task manifests so that plugins can reference and change accordingly.

candysmurf:
Probably, we should not prevent to get this PR merged if you guys have a better strategy for examples. 
Currently "plugin" package has all mock plugins. Should we rename it? I would not expect "plugin" package has all testing mocks.

core/task_medium_test.go
katarzyna-z: Corrected, now error is returned.

docs/TASKS.md
katarzyna-z: I hope that I replaced all "snap" to "Snap".

PR #1274, opened 2016-10-11
(maint) add default and parallel builds.

nanliu:
This reduces the number of builds and speed up builds by increasing number of parallel plugin builds. We only need mac and linux builds when releasing, so it is no longer the default.

### DIFF
--- a/scripts/build_plugins.sh
+++ b/scripts/build_plugins.sh
@@ -32,7 +32,6 @@ _info "project path: ${__proj_dir}"
 git_version=$(_git_version)
 go_build=(go build -ldflags "-w -X main.gitversion=${git_version}")
 
-_info "snap build version: ${git_version}"
 _info "git commit: $(git log --pretty=format:"%H" -1)"
 
 # Disable CGO for builds.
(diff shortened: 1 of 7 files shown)


scripts/build_plugins.sh
```diff
@@ -0,0 +1,61 @@
...
+# Disable CGO for builds.
+export CGO_ENABLED=0
+
+# rebuild binaries:
+export GOOS=${GOOS:-$(uname -s | tr '[:upper:]' '[:lower:]')}
+export GOARCH=${GOARCH:-"amd64"}
+
+OS=$(uname -s)
+if [[ "${OS}" == "Darwin" ]]; then
+  p=$(type -p sysctl > /dev/null && sysctl -n hw.ncpu || echo "1")
```
kindermoumoute: Is this possible to still have an option to manually define p? Some developer may prefer to wait longer but not have their laptop frozen.

nanliu:
Updated PR to show os/arch when building plugins:
```
2016-10-12 00:08:59 UTC [     info] building plugins for linux/amd64 in 2 parallels
2016-10-12 00:08:59 UTC [     info] building snap-plugin-collector-anothermock1 for linux/amd64
...
```

I added an option to allow you to specify number of parallel builds.
```
BUILD_JOBS=2 make plugins
```

kindermoumoute:
LGTM

PR #1281, opened 2016-10-13
(SDI-1827): Fix #1128 Loading collector twice causes task failure

### LINKED ISSUE #1128: Loading collector twice causes failure of task creation
### Description

The issue appears when one accidentally attempts to load already loaded collector plugin. Framework informs that plugin has been already loaded, but after that creating a task fails. snapd needs to be restarted and all the plugins loaded properly from the beginning in order to run a task. This is quite confusing as it seems that something is wrong with the collector.
### Steps to reproduce this error:
1. Run Snap:
   `snapd -l 1 -t 0`
2. Load collector plugin (tested with iostat, ethtool)
   `snapctl plugin load snap-plugin-collector-iostat/build/rootfs/snap-plugin-collector-iostat`
3. 'Accidentally' load collector plugin once again
   `snapctl plugin load snap-plugin-collector-iostat/build/rootfs/snap-plugin-collector-iostat`
4. Load mock-file-publisher and processor-passthru
5. Create a task:
   [iostat.txt](https://github.com/intelsdi-x/snap/files/407012/iostat.txt)
### Result

Task is created but does not collect metrics.

marcin-krolik:
Fixes #1128 

Summary of changes:
- after plugin response is recieved, additional check added to look if plugin already exists in loaded plugins table

Testing done:
- smal
- legacy

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/plugin_manager.go
+++ b/control/plugin_manager.go
@@ -314,6 +314,16 @@ func (p *pluginManager) LoadPlugin(details *pluginDetails, emitter gomit.Emitter
 		return nil, serror.New(err)
 	}
 
+	// check if plugin is already loaded
+	key := fmt.Sprintf("%s"+core.Separator+"%s"+core.Separator+"%d", resp.Meta.Type.String(), resp.Meta.Name, resp.Meta.Version)
+	if _, exists := p.loadedPlugins.table[key]; exists {
+		return nil, serror.New(ErrPluginAlreadyLoaded, map[string]interface{}{
+			"plugin-name":    resp.Meta.Name,
+			"plugin-version": resp.Meta.Version,
+			"plugin-type":    resp.Type.String(),
+		})
+	}
+
 	ap, err := newAvailablePlugin(resp, emitter, ePlugin)
 	if err != nil {
 		pmLogger.WithFields(log.Fields{


IRCody:
Are you able to reproduce the issue from #1128 on master? I tried again just now and wasn't able to. Curious what steps I might be missing, can you provide the steps so I can properly test the PR?

marcin-krolik:
@IRCody I can't reproduce this issue either on current master or on `df51bec` which is the last commit before the issue was reported. Weird ...

IRCody:
I was able to repro it at the time so that is very weird. I don't feel confident merging this fix though if we can't reproduce the original error and verify it's fixed.

marcin-krolik:
@IRCody I got your point, but at the same time if plugin is already loaded, this change helps to stop loading early, right after response is received.

IRCody:
@marcin-krolik: I think we still need to track down why we can't reproduce the issue and find out what's going on.

marcin-krolik:
@IRCody I used git bisect to track down the issue. This is quite messy, because at that time we were working on simplification of interaction between scheduler and control, and there are Matuesz's commits, mine and Joel's and it is hard to track it, but here is what I found:
- Last commit with issue reproducible `3e91e77`
- Next commit is Matuesz commit and I coudl not compile snap with it
- Following commit has this issue resolved `2afb26b`

IRCody:
@marcin-krolik: Thanks for tracking it down. Do you think this PR is still needed? Is the advantage that it would fail earlier?

jcooklin:
After subscription groups was landed this issue was avoided because we get the loaded plugin through the `pluginManager` [here](https://github.com/intelsdi-x/snap/blob/master/control/subscription_group.go#L391).

The fact that `metricType` has a reference to a loaded plugin that can potentially be invalid is still a problem and should be fixed.

@marcin-krolik  What do you think about changing the type of [`metricType.Plugin`](https://github.com/intelsdi-x/snap/blob/master/control/metrics.go#L133) from `*loadedPlugin` to `core.CatalogedPlugin`.  This will remove the the fact that the catalog has a reference to an invalid `loadedPlugin` (invalid in the sense that the path to the plugin is wrong).  If we don't change the type of the field [this](https://github.com/intelsdi-x/snap/blob/master/control/plugin_manager.go#L452) will come back to bite us at some point.

marcin-krolik:
@jcooklin I like this idea. I'm going to start working on it.

PR #1287, opened 2016-10-17
Added host and port to tribe members api

astub:
Feature 

Summary of changes:
- Added the host name to the tribe members api
- Added the host's port to the tribe members api

Testing done:
- Tested members api route with curl
- Ran small test

@intelsdi-x/snap-maintainers


astub:
I'm updating /mgmt/rest/fixtures/mock_tribe_manager.go which is causing med test to fail.

Issue #1289, opened 2016-10-17
Deprecating GoRPC


IRCody:
## TL;DR

As Snap makes the move to gRPC, we will officially consider the  Go RPC plugin libraries deprecated. All plugin authors should consider migrating to utilize the new gRPC. We currently have gRPC plugin libraries for [Go](https://github.com/intelsdi-x/snap-plugin-lib-go) and [C++](https://github.com/intelsdi-x/snap-plugin-lib-cpp).

Plugins written using Go RPC plugins libs **will continue to work after the 1.0** release. Removal of support for loading plugins written using the Go RPC libs will be announced well in advance of it’s removal.

## Reasoning

It is our goal to provide the ability to write plugins in many languages. We think this is best accomplished by utilizing gRPC. By standardizing on a gRPC interface we can immediately have plugins written in any language that [gRPC supports](http://www.grpc.io/docs/). We plan to offer plugin libs for more popular languages, see: [Go libary](https://github.com/intelsdi-x/snap-plugin-lib-go), [C++ library](https://github.com/intelsdi-x/snap-plugin-lib-cpp) but due to the nature of gRPC nothing is stopping someone from taking the proto file and generating code in their favorite gRPC supported language.If you’re interested in another language you can let us know by opening an [issue](https://github.com/intelsdi-x/snap/issues/new) to let us know or writing your on. We would be happy to link to community created libraries.

## What this means for plugins using Go RPC 

Plugins written using Go RPC will continue to work for the foreseeable future. The exact timing of when support for them will be dropped hasn’t been decided yet. It will be after the 1.0 release and will be communicated well in advance of it’s removal.

## Converting plugins

Converting a plugin to the new libraries should be straightforward. Most of the code will be the same/very similar. For an example of what a converted plugin looks like see the [graphite publisher](https://github.com/intelsdi-x/snap-plugin-publisher-graphite). To get help converting an existing plugin (or writing a new one) come join us on [slack](https://intelsdi-x.herokuapp.com/).

ConnorDoyle:
> Removal of support for loading plugins written using the Go RPC libs will be announced well in advance of it’s removal.

Can we be more specific? e.g. would it be removed in version 2.0?

bjray:
@ConnorDoyle, we are still working out the targeted scope for 2.0, but in all likelihood it would be delivered with that release, particularly given that it represents a change to our contract.

kjlyon:
[...]

PR #1291, opened 2016-10-17
Untangled Go RPC code

kjlyon:
Enables #1289 to be done easily when the time comes

Summary of changes:
- Separated Go RPC plugin libraries and non plugin related functions 
- Created files with deprecated flags to hold code that is used by GoRPC
- Added deprecation warning that will appear when loading legacy plugins

Testing done:
- Unit tests, Running snap

@intelsdi-x/snap-maintainers

- Identified code used by GoRPC plugins
- Created files with deprecated flags to hold code that is used by GoRPC plugins
- Added deprecation warning when loading legacy plugin


control/plugin/plugin_deprecated.go
```diff
@@ -0,0 +1,308 @@
+/*          **  DEPRECATED  **
```
IRCody: We probably want this separate from the copyright notice separated by a line in it's own comment block. We want attention to be drawn to it.

control/available_plugin.go
```diff
@@ -107,6 +107,9 @@ func newAvailablePlugin(resp plugin.Response, emitter gomit.Emitter, ep executab
 			}
 			ap.client = c
 		case plugin.NativeRPC:
+			log.WithFields(log.Fields{
+				"_module": "control-aplugin",
+			}).Warning("This plugin has been deprecated.")
```
IRCody: We need more description here.

control/available_plugin.go
```diff
@@ -107,6 +107,9 @@ func newAvailablePlugin(resp plugin.Response, emitter gomit.Emitter, ep executab
 			}
 			ap.client = c
 		case plugin.NativeRPC:
+			log.WithFields(log.Fields{
```
IRCody: Give me the block also (newAvailablePlugin) so that I know exactly where this message came from.

control/plugin/plugin.go
```diff
@@ -21,21 +21,10 @@ package plugin
...
-	"encoding/json"
-	"fmt"
-	"io" // Don't use "fmt.Print*"
-	"net"
-	"net/http"
-	"net/rpc"
-	"net/rpc/jsonrpc"
-	"regexp"
-	"runtime"
+	"crypto/rsa" // Don't use "fmt.Print*"
```
jcooklin: This comment `// Don't use "fmt.Print*"` should be moved up below the warning or removed.

control/plugin/collector_deprecated.go
```diff
@@ -1,3 +1,5 @@
+/*          **  DEPRECATED  **          */
```
kjlyon: I made the suggested updates

control/plugin/publisher_proxy_deprecated.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (p *gRPCPublisherProxy) Publish(ctx context.Context, arg *rpc.PublishArg) (*common.Empty, error) {
+	defer catchPluginPanic(p.Session.Logger())
+	err := p.Plugin.Publish(arg.ContentType, arg.Content, common.ParseConfig(arg.Config))
+	if err != nil {
+		return &common.Empty{}, err
+	}
+	return &common.Empty{}, nil
+}*/
```
IRCody: Do we want to leave this stuff in here commented out?

kjlyon: I left it there since it was in publisher_proxy before I untangled the code. I don't know if it has a purpose or future purpose.

IRCody: It doesn't have a future purpose and should've been removed previously.

kjlyon: Makes sense. I deleted that section and a couple other places that were just commented out code.

control/available_plugin.go
```diff
@@ -107,6 +107,11 @@ func newAvailablePlugin(resp plugin.Response, emitter gomit.Emitter, ep executab
...
+				"_module": "control-aplugin",
+				"_block":  "newAvailablePlugin",
+				"aplugin": ap.name,
+			}).Warning("This plugin is using a deprecated RPC protocol. Please update " + ap.name)
```
IRCody: I think this would look better if we change `aplugin` to `plugin_name` and then remove the name from the warning string. Instead it could be something like `This plugin is using a deprecated RPC protocol.`. I'm not convinced if we should have some sort of message pointing them somewhere (a url?) where they can find out more. What do you think?

PR #1301, opened 2016-10-20
SDI-2115: remove metric.String() in plugin-lib-go

candysmurf:
Summary of changes:
- added local method toString(ns []string) for GRPC mock test.
- change tests use the local toString instead of from plugin-lib-go.

Testing done:
-  test medium
-  test small

@intelsdi-x/snap-maintainers

### DIFF
--- a/plugin/collector/snap-plugin-collector-mock2-grpc/mock/mock_medium_test.go
+++ b/plugin/collector/snap-plugin-collector-mock2-grpc/mock/mock_medium_test.go
@@ -23,11 +23,11 @@ package mock
 
 import (
 	"math/rand"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/intelsdi-x/snap-plugin-lib-go/v1/plugin"
-	"github.com/intelsdi-x/snap-plugin-utilities/str"
 	. "github.com/smartystreets/goconvey/convey"
 )
 
@@ -96,7 +96,7 @@ func TestCollectMetric(t *testing.T) {
 
 					// only one metric for this specific hostname should be returned
 					So(len(mts), ShouldEqual, 1)
-					So(mts[0].Namespace.String(), ShouldEqual, "/intel/mock/host0/baz")
+					So(mts[0].Namespace.Strings(), ShouldResemble, []string{"intel", "mock", "host0", "baz"})
 
 					Convey("returned metric should have data type integer", func() {
 						_, ok := mts[0].Data.(int)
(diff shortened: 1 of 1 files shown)


IzabellaRaulin:
@candysmurf, why metric.String() has been removed from plugin-lib-go? Should it has both methods String() and Strings() as snap's core.Namespace? Could you explain the motivation of removing this?

I prefer to avoid creating private "toString" function in each package if it's possible - regarding [https://github.com/intelsdi-x/snap-plugin-lib-go/pull/38](https://github.com/intelsdi-x/snap-plugin-lib-go/pull/38/files#diff-7f1fcd671dd1765edb0fc47e386a3258R74.

candysmurf:
@IzabellaRaulin, thanks for your concern. Please refer to this issue https://github.com/intelsdi-x/snap-plugin-lib-go/issues/34 opened in plugin-lib-go. That's the decision made in our backlog grooming. 
Feel free to express your concerns.

IRCody:
@IzabellaRaulin: I think the goal is to remove the implied guarantee that lives behind offering a String() method. With the work that has gone into arbitrary separators now the implementation behind a String() method would be decently complex and subject to future change. We don't want to provide a guarantee that it will work in a specific way.

kindermoumoute:
@candysmurf, as for the [plugin lib PR](https://github.com/intelsdi-x/snap-plugin-lib-go/pull/38), I suggest you replace string "ShouldEqual" by slice "ShouldResemble":
```
So(slice1, ShouldResemble, slice2)
```

Doing it that way avoid creating a "toString" function as @IzabellaRaulin mentioned it.

candysmurf:
@IRCody & @kindermoumoute, if not comparing strings but array of strings, it requires comparison of each array element or using reflect.DeepEqual(). An additional function "stringsEqual" will need to be added. Please confirm you prefer "stringEqual" over "toString". Should we run a benchmark? :-)

candysmurf:
@IRCody, I can see the good use of doing array comparison if we like to recommend this way to all plugins and if it has no performance penalty, I would be glad to change it.

candysmurf:
@IRCody, thanks for doing the benchmark!  We agreed to not use toString() method for two reasons:

* our code often does the conversion first, such as strings.Join(ns, "/"), then comparison. It's less performant than comparing the slices directly.

* we promote using namespace array value or []string slice directly instead of converting []string -> string, then compare.

plugin/collector/snap-plugin-collector-mock2-grpc/mock/mock_medium_test.go
```diff
@@ -96,7 +96,7 @@ func TestCollectMetric(t *testing.T) {
 
 					// only one metric for this specific hostname should be returned
 					So(len(mts), ShouldEqual, 1)
-					So(mts[0].Namespace.String(), ShouldEqual, "/intel/mock/host0/baz")
+					So(mts[0].Namespace.Strings(), ShouldResemble, plugin.NewNamespace("intel", "mock", "host0", "baz").Strings())
```
kindermoumoute: Why not to write a slice of string directly?
`plugin.NewNamespace("intel", "mock", "host0", "baz").Strings()` ==> `[]string{"intel", "mock", "host0", "baz"}`

candysmurf: not sure why it was like that. Probably make sure the correct library is used. Not core.Namespace

kindermoumoute:
LGTM

PR #1306, opened 2016-10-22
Add an API endpoint to profiling tools

### LINKED ISSUE #1305: Add an API endpoint to easily profile snapd
As a Snap contributor, I want to be able to easily use profiling tools, so I can generate clear views of the resources used by snapd, during a customized period.

kindermoumoute:
Fixes #1305, update documentation and add implementation of optional profiling tools endpoints in the Snap API.

Summary of changes:
- Add a flag to start profiling tools.
- When starting a plugin, pass the profiling tools flag.
- Add API endpoints for profiling tools in Snap and in the plugin side (new Go lib https://github.com/intelsdi-x/snap-plugin-lib-go/pull/44)

Testing done:
- start snapd with pprof flag `snapd -t 0 --pprof`.
- load plugins and start a task.
- curl snap API `curl "http://127.0.0.1:8181/v1/debug/pprof/profile?seconds=5" > cpu.pprof`
- `snapctl plugin list --running`, pprof port of running plugins should be displayed.
- curl plugin API on their pprof port `curl "http://127.0.0.1:<PORT>/v1/debug/pprof/profile?seconds=5" > cpu.pprof`

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/available_plugin.go
+++ b/control/available_plugin.go
@@ -75,6 +75,7 @@ type availablePlugin struct {
 	exec               string
 	execPath           string
 	fromPackage        bool
+	pprofPort          string
 }
 
 // newAvailablePlugin returns an availablePlugin with information from a
@@ -92,6 +93,7 @@ func newAvailablePlugin(resp plugin.Response, emitter gomit.Emitter, ep executab
 		healthChan:  make(chan error, 1),
 		lastHitTime: time.Now(),
 		ePlugin:     ep,
+		pprofPort:   resp.PprofAddress,
 	}
 	ap.key = fmt.Sprintf("%s"+core.Separator+"%s"+core.Separator+"%d", ap.pluginType.String(), ap.name, ap.version)
 
@@ -165,6 +167,10 @@ func newAvailablePlugin(resp plugin.Response, emitter gomit.Emitter, ep executab
 	}
 
 	return ap, nil
+}
+
+func (a *availablePlugin) Port() string {
+	return a.pprofPort
 }
 
 func (a *availablePlugin) ID() uint32 {
--- a/mgmt/rest/rbody/plugin.go
+++ b/mgmt/rest/rbody/plugin.go
@@ -107,4 +107,5 @@ type AvailablePlugin struct {
 	LastHitTimestamp int64  `json:"last_hit_timestamp"`
 	ID               uint32 `json:"id"`
 	Href             string `json:"href"`
+	PprofPort        string `json:"pprof_port"`
 }
--- a/mgmt/rest/server.go
+++ b/mgmt/rest/server.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/http/pprof"
 	"strings"
 	"sync"
 	"time"
@@ -60,6 +61,7 @@ const (
 	defaultAuth            bool   = false
 	defaultAuthPassword    string = ""
 	defaultPortSetByConfig bool   = false
+	defaultPprof           bool   = false
 )
 
 var (
@@ -83,6 +85,7 @@ type Config struct {
 	RestAuth         bool   `json:"rest_auth"yaml:"rest_auth"`
 	RestAuthPassword string `json:"rest_auth_password"yaml:"rest_auth_password"`
 	portSetByConfig  bool   ``
+	Pprof            bool   `json:"pprof"yaml:"pprof"`
 }
 
 const (
@@ -115,6 +118,9 @@ const (
 					},
 					"addr" : {
 						"type": "string"
+					},
+					"pprof": {
+						"type": "boolean"
 					}
 				},
 				"additionalProperties": false
@@ -174,6 +180,7 @@ type Server struct {
 	r          *httprouter.Router
 	snapTLS    *snapTLS
 	auth       bool
+	pprof      bool
 	authpwd    string
 	addrString string
 	addr       net.Addr
@@ -191,10 +198,12 @@ func New(cfg *Config) (*Server, error) {
 	https := cfg.HTTPS
 	cpath := cfg.RestCertificate
 	kpath := cfg.RestKey
+	pprof := cfg.Pprof
 	s := &Server{
 		err:        make(chan error),
 		killChan:   make(chan struct{}),
 		addrString: cfg.Address,
+		pprof:      pprof,
 	}
 	if https {
 		var err error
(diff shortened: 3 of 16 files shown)


docs/PROFILING.md
```diff
@@ -59,56 +59,30 @@ go get github.com/uber/go-torch
...
-    signal.Notify(c, os.Interrupt, os.Kill, syscall.SIGTERM)
-
-    //Let's block until someone tells us to quit
-    go func() {
-        sig := <-c
-        p.Stop()    // Stop profiling
-        ...
+You can now get at any time a profile during a defined period (e.g. 120 seconds):
+```bash
+curl "http://127.0.0.1:8181/v1/debug/pprof/profile?seconds=120" > cpu.pprof
```
candysmurf: please mention that whatever time specified, the curl command needs to wait for the time interval.

kindermoumoute:
I'll add a demo gif in the documentation, to show what is supposed to happen.

mgmt/rest/server.go
```diff
@@ -492,6 +500,15 @@ func (s *Server) addRoutes() {
 		s.r.GET("/v1/tribe/members", s.getMembers)
 		s.r.GET("/v1/tribe/member/:name", s.getMember)
 	}
+
+	// profiling tools routes
+	if s.pprof {
+		s.r.GET("/v1/debug/pprof/", s.index)
+		s.r.GET("/v1/debug/pprof/cmdline", s.cmdline)
+		s.r.GET("/v1/debug/pprof/profile", s.profile)
+		s.r.GET("/v1/debug/pprof/symbo", s.symbol)
```
candysmurf: can we use symbol in the GET URI as well.

candysmurf:
I was able to run and see the torch.svg with Sierra.
screen shot 2016-10-31 at 3 20 17 pm


However, tasks were created and ran file with the following error:
```
▶ snapctl task create -t ~/task/task-mock-file.json
Error creating task:Unknown API response: invalid character 'P' looking for beginning of value

 Received: PANIC: runtime error: invalid memory address or nil pointer dereference
goroutine 19535 [running]:
github.com/intelsdi-x/snap/vendor/github.com/codegangsta/negroni.(*Recovery).ServeHTTP.func1(0xd13110, 0xc420396bc0, 0xc420130ac0)
	/Users/egu/.gvm/pkgsets/go1.7/snap-1.7/src/github.com/intelsdi-x/snap/vendor/github.com/codegangsta/negroni/recovery.go:34 +0xf2
panic(0x6d6f80, 0xc4200100b0)
	/Users/egu/.gvm/gos/go1.7/src/runtime/panic.go:458 +0x243
github.com/intelsdi-x/snap/mgmt/rest/rbody.assertSchedule(0xb15c80, 0xc42037e200, 0xc420224e60)
	/Users/egu/.gvm/pkgsets/go1.7/snap-1.7/src/github.com/intelsdi-x/snap/mgmt/rest/rbody/task.go:225 +0x2b6
github.com/intelsdi-x/snap/mgmt/rest/rbody.AddSchedulerTaskFromTask(0xb1d6e0, 0xc4201d28c0, 0xc42029d488)
	/Users/egu/.gvm/pkgsets/go1.7/snap-1.7/src/github.com/intelsdi-x/snap/mgmt/rest/rbody/task.go:111 +0x3d8
github.com/intelsdi-x/snap/mgmt/rest.(*Server).addTask(0xc4200df520, 0xd13110, 0xc420396bc0, 0xc4203d1c20, 0x0, 0x0, 0x0)
	/Users/egu/.gv
```

kindermoumoute:
Documentation will come in another PR. https://github.com/intelsdi-x/snap-plugin-lib-go/pull/44 needs to be merged first so I can update glide.lock in this PR.

mgmt/rest/rbody/plugin.go
```diff
@@ -107,4 +107,5 @@ type AvailablePlugin struct {
 	LastHitTimestamp int64  `json:"last_hit_timestamp"`
 	ID               uint32 `json:"id"`
 	Href             string `json:"href"`
+	PprofPort        string `json:"pprof_port"`
```
candysmurf: why is it using a string type for port?

control/available_plugin.go
```diff
@@ -75,6 +75,7 @@ type availablePlugin struct {
 	exec               string
 	execPath           string
 	fromPackage        bool
+	pprofPort          string
```
candysmurf: string for port?

mgmt/rest/rbody/plugin.go
kindermoumoute: At the end the goal is to display the port as a string when typing `snapctl plugin list --running`. Communication between snapctl and snapd are in json, which mean string anyway.

control/available_plugin.go
kindermoumoute: Same reason, we have no reason to use it as an int. The goal is to displayed it at the end.

mgmt/rest/server.go
```diff
@@ -60,6 +61,7 @@ const (
 	defaultAuth            bool   = false
 	defaultAuthPassword    string = ""
 	defaultPortSetByConfig bool   = false
+	defaultPprof           bool   = false
```
candysmurf: you don't need to set. The default is false. Of course, it does not hurt here.

mgmt/rest/server.go
```diff
@@ -492,6 +507,41 @@ func (s *Server) addRoutes() {
 		s.r.GET("/v1/tribe/members", s.getMembers)
 		s.r.GET("/v1/tribe/member/:name", s.getMember)
 	}
+
+	// profiling tools routes
+	if s.pprof {
```
candysmurf: are you sure that you want to dynamically change it when snapd is already running? Will it need to restart snapd by doing this way?

mgmt/rest/server.go
kindermoumoute: It doesn't change the binary (SSA optimize it), it's just to make it clear.

mgmt/rest/server.go
kindermoumoute: > are you sure that you want to dynamically change it when snapd is already running? 

We don't enable/disable profiling tools without re-starting Snap. We do the same way than tribe. Else we would need another route to the API, in order to handle dynamic config changes. And I'm not sure we want that.

kindermoumoute: This feature is not about profiling tools, it's about all global option we can list with `snapd -h`. Currently we can't update those options without restarting Snap.

IRCody: I think adding the ability to dynamically update the config is outside the scope of this PR.

candysmurf: agreed that we can do dynamiclly update config in a different PR.

PR #1310, opened 2016-10-25
(SDI-1827): Fix #1128 Loading collector twice causes task failure

### LINKED ISSUE #1128: Loading collector twice causes failure of task creation

marcin-krolik:
Fixes #1128

Summary of changes:
- `metricType.Plugin` is now `core.CatalogedPlugin` and contains copy
  - after plugin response is received, additional check added to look if plugin already exists in loaded plugins table
- `loadedPlugin` attributes setting moved just below receiving plugin response

Testing done:
- small
- medium
- legacy

@intelsdi-x/snap-maintainers

---
This change is [Reviewable](https://reviewable.io/reviews/intelsdi-x/snap/1310)

### DIFF
--- a/control/metrics.go
+++ b/control/metrics.go
@@ -131,7 +131,7 @@ func (m *metricCatalogItem) Versions() map[int]core.Metric {
 }
 
 type metricType struct {
-	Plugin             *loadedPlugin
+	Plugin             core.CatalogedPlugin
 	namespace          core.Namespace
 	version            int
 	lastAdvertisedTime time.Time
@@ -163,12 +163,24 @@ func (m *metric) Version() int {
 	return m.version
 }
 
-func (m *metric) Data() interface{}             { return nil }
-func (m *metric) Description() string           { return "" }
-func (m *metric) Unit() string                  { return "" }
-func (m *metric) Tags() map[string]string       { return nil }
-func (m *metric) LastAdvertisedTime() time.Time { return time.Unix(0, 0) }
-func (m *metric) Timestamp() time.Time          { return time.Unix(0, 0) }
+func (m *metric) Data() interface{} {
+	return nil
+}
+func (m *metric) Description() string {
+	return ""
+}
+func (m *metric) Unit() string {
+	return ""
+}
+func (m *metric) Tags() map[string]string {
+	return nil
+}
+func (m *metric) LastAdvertisedTime() time.Time {
+	return time.Unix(0, 0)
+}
+func (m *metric) Timestamp() time.Time {
+	return time.Unix(0, 0)
+}
 
 type processesConfigData interface {
 	Process(map[string]ctypes.ConfigValue) (*map[string]ctypes.ConfigValue, *cpolicy.ProcessingErrors)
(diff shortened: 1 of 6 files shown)


candysmurf:
changes look good to me. Let me test it. @jcooklin & @IRCody welcome to give comments.

control/mttrie.go
```diff
@@ -72,7 +74,8 @@ func NewMTTrie() *MTTrie {
 func (m *MTTrie) String() string {
 	out := ""
 	for _, mt := range m.gatherMetricTypes() {
-		out += fmt.Sprintf("%s => %s\n", mt.Key(), mt.Plugin.Key())
+		pluginKey := fmt.Sprintf("%s"+core.Separator+"%s"+core.Separator+"%d", mt.Plugin.TypeName(), mt.Plugin.Name(), mt.Plugin.Version())
+		out += fmt.Sprintf("%s => %s\n", mt.Key(), pluginKey)
```
IRCody: plugin doesn't have a `Key()` method anymore?

control/plugin_manager.go
```diff
@@ -180,51 +180,51 @@ type loadedPlugin struct {
...
 
 // LoadedTimestamp returns a unix timestamp of the LoadTime of a plugin
 // implements the CatalogedPlugin interface
-func (lp *loadedPlugin) LoadedTimestamp() *time.Time {
+func (lp loadedPlugin) LoadedTimestamp() *time.Time {
 	return &lp.LoadedTime
 }
 
-func (lp *loadedPlugin) Policy() *cpolicy.ConfigPolicy {
+func (lp loadedPlugin) Policy() *cpolicy.ConfigPolicy {
```
IRCody: Why did these need to change?

marcin-krolik: `metricType.Plugin` has copy of `loadedPlugin` now

IRCody: I don't see what the receiver type has to do with having a copy vs reference? In either case you're going to need to copy the individual elements as well because loadedPlugin has reference elements. Maybe I am missing something?
```
type loadedPlugin struct {
	Meta         plugin.PluginMeta
	Details      *pluginDetails
	Type         plugin.PluginType
	State        pluginState
	Token        string
	LoadedTime   time.Time
	ConfigPolicy *cpolicy.ConfigPolicy
}
```

I'd also be curious if, since there are still places that use loadedPlugin as a reference, this doesn't break something else.

control/metrics.go
```diff
@@ -289,7 +301,7 @@ func (mc *metricCatalog) AddLoadedMetricType(lp *loadedPlugin, mt core.Metric) e
 		return err
 	}
 	newMt := metricType{
-		Plugin:             lp,
+		Plugin:             *lp,
```
jcooklin: Consider creating a `catalogedPlugin` struct and `newCatalogedPlugin` function that returns a core.CatalogedPlugin (implementing a deep copy).

control/plugin_manager.go
marcin-krolik: @IRCody I think it is similar to below example
```go
type Fooer interface {
	foo()
}

type Foo struct {}
func (f *Foo) foo(){}

func doStuff(){
  var fi Fooer
  // works ok since *Foo implements interface Fooer
  fi = &Foo{}
  fi.foo()
  
  // Compile error, Foo does not implement Fooer (foo method has pointer receiver)
  fi = Foo{}
  fi.foo()
}
```
If there is something I'm missing, please let me know.

control/metrics.go
marcin-krolik: That was my initial implementation, but I thought it's too verbose and changed it to dereferencing `loadedPlugin`. I guess it was mistake and I missed the need of deep copy.

control/plugin_manager.go
IRCody: > In either case you're going to need to copy the individual elements as well because loadedPlugin has reference elements.

Here I was talking about cpolicy/plugin details being pointers not about how receivers work.

>All these methods are 'getters', so they should only be used to retrieve values, so it should not be the case.

My comment wasn't really directed at this but getter methods that return pointers still allow you to mutate the object, even though the receiver is a value receiver.

I was trying to bring up that changing loadedPlugin from pointer receivers to value receivers could result in weird inconsistencies since there are still places there use loadedPlugin as a pointer. Could be confusing in the future.

control/metrics.go
```diff
@@ -250,6 +262,62 @@ func (m *metricType) Unit() string {
 	return m.unit
 }
 
+type catalogedPlugin struct {
```
IRCody: Is there really not another type that implements core.catalogedPlugin in control?

control/metrics.go
```diff
@@ -250,6 +262,62 @@ func (m *metricType) Unit() string {
...
+func newCatalogedPlugin(lp *loadedPlugin) core.CatalogedPlugin {
+	return &catalogedPlugin{
+		name:         lp.Name(),
+		version:      lp.Version(),
+		signed:       lp.IsSigned(),
+		typeName:     lp.Type,
+		state:        lp.State,
+		path:         lp.PluginPath(),
+		loadedTime:   lp.LoadedTime,
+		configPolicy: *lp.Policy(),
```
IRCody: I don't think this helps because Config Policy's definition is:

```go
type ConfigPolicy struct {
		config *ctree.ConfigTree
}	
```

marcin-krolik: Added proper copy of config policy

control/metrics.go
marcin-krolik: I don't see any. Is there a problem with definition of private structure which implements this interface? It is quite common in other packages as well.

IRCody: I was specifically asking if there was another one in the control package that implemented it. It seems like there would be.

control/plugin/cpolicy/node.go
```diff
@@ -71,6 +71,56 @@ func NewPolicyNode() *ConfigPolicyNode {
 	}
 }
 
+func (c *ConfigPolicyNode) CopyRules() ([]Rule, error) {
+	rules := make([]Rule, len(c.rules))
+	i := 0
+	for _, rule := range c.rules {
```
IRCody: I think this would be clearer as `for i := 0; i < len(c.rules); i++ { rule := c.rules[i];...`.

marcin-krolik: `c.rules` is a `map[string]Rule`, so it's not possible to use `i` as key. I would need to create slice of map values firsts.

IRCody: That's a good point! Do you think it'd be clearer without the i then and just using append.

IzabellaRaulin:
Travis tests failed because of:
```
# github.com/intelsdi-x/snap/control/plugin/cpolicy
control/plugin/cpolicy/node.go:118: undefined: i
make[1]: *** [snap] Error 1
make[1]: Leaving directory `/home/travis/gopath/src/github.com/intelsdi-x/snap'
make: *** [default] Error 2
```

jcooklin:
LGTM @IRCody do you have further suggestions, questions?

IRCody:
LGTM but can we rebase into a single commit before merging?

PR #1311, opened 2016-10-26
Added docs/MAINTAINERS.md

lmroz:
Summary of changes:
- Added document listing snap maintainers and plugins maintaintainers

Because it's not possible to list team members from outside of our organization, this information has to be delivered another way. This list is automatically generated so it may require adjustments.

@intelsdi-x/snap-maintainers

### DIFF
--- a/docs/MAINTAINERS
+++ b/docs/MAINTAINERS
@@ -1,0 +1,38 @@
+snap-maintainers, plugin-maintainers:
+  Joel Cooklin (@jcooklin)
+  Cody Roseborough (@IRCody)
+  Emily Gu (@candysmurf)
+  Marcin Krolik (@marcin-krolik)
+  Izabella Raulin (@IzabellaRaulin)
+  Matt Brender (@mjbrender)
+  Andrzej Kuriata (@andrzej-k)
+  Connor Doyle (@ConnorDoyle)
+  Olivier Cano (@kindermoumoute)
+  Niklas Quarfot Nielsen (@nqn)
+
+snap-maintainers:
+  Nicholas Weaver (@lynxbat)
+  dan pittman (@danielscottt)
+  Kelly Lyon (@kjlyon)
+
+plugin-maintainers:
+  Pawel Palucki (@ppalucki)
+  Patryk Matyjasek (@PatrykMatyjasek)
+  Marcin Spoczynski (@sandlbn)
+  Szymon Konefal (@skonefal)
+  Mateusz Kleina (@mkleina)
+  Łukasz Mróz (@lmroz)
+  Maria Rolbiecka (@mkuculyma)
+  Katarzyna Kujawa (@katarzyna-z)
+  iwankgb (@iwankgb)
+  B.J. Ray (@bjray)
+  Maciej Cichocki (@kimbarbel)
+  Maciej Patelczyk (@mpatelcz)
+  Paweł Adamczyk (@ppadamcz)
+  Piotr (@padamowi)
+  akwasnie (@akwasnie)
+  maciej-wisniewski (@maciej-wisniewski)
+  marcintao (@marcintao)
+  ptbuilder (@ptbuilder)
+  rashmigottipati (@rashmigottipati)
+


lmroz:
Use this link to view rendered version: https://github.com/lmroz/snap/blob/48c9e2302ac5fce6e218249cf177ba58508351a0/docs/MAINTAINERS.md

bjray:
#snapmaintainers, any objections to displaying our email if it is already available in your github profile?

mjbrender:
@candysmurf I'm a fan of a maintainer file format - it's something I noticed in CoreOS files. Given that people who aren't part of our GitHub organization cannot use the `@intelsdi-x/snap-maintainers` tag, we were looking for an alternative. 

This file gives a way of identifying active maintainers for Snap. Given that projects like this one have shifting maintainers, it's important for other contributors to know who is available for them *today* if they have questions. 

@bjray I'm okay with sharing the email from my profile.

mjbrender:
@lmroz what tool do you use to generate this file? I'm curious how we'll generate this in the future and where it fits into the toolchain @nanliu has been maintaining.

candysmurf:
@mjbrender, agreed that showing active maintainers is helpful. Notifying a team/alias seems more useful than an individual. Also with so many maintainers, do we need timezone info?:-)

ConnorDoyle:
Would prefer not to list the profile emails here unless we want to encourage direct mail to maintainers; a link to each maintainer's github profile should suffice.

candysmurf:
separating snap-maintainers and snap-plugin-maintainers into two lists and no group info?

lmroz:
@mjbrender so the list can look this way https://gist.github.com/lmroz/2194e498d6632ac3fb0ca3b3d53f8d80 - I just need some kind of general agreement about layout :tada: 

And there are tools I'm playing with to sort out these problems with maintainers https://github.com/lmroz/org2docmd
Me & @bjray tested notification bot yesterday and seems like it's working :8ball:
Maintainers list generator is rather customizable if you can compile it (template still hardcoded).
And I'm not guaranteeing it's bug free (today I found very nasty one in generator, but maybe that was the last one).

docs/MAINTAINERS.md
```diff
@@ -0,0 +1,261 @@
...
+</ul></td></tr>
+<tr><td colspan="3"></td></tr>
+<tr><td colspan="3"><h4><a name="katarzyna-z" href="#katarzyna-z">#</a> Katarzyna Kujawa</h4></td></tr>
+<td><img src="https://avatars.githubusercontent.com/u/16666942?v=3" width="90px" /></td>
+<td><b>Github:</b> <a href="https://github.com/katarzyna-z">@katarzyna-z</a><br /><br /><b>Company:</b> Intel Corporation</td>
+<td><b>Groups:</b> <br /><br /><ul>
+    <li><a href="https://github.com/orgs/intelsdi-x/teams/plugin-maintainers">@intelsdi-x/plugin-maintainers</a></li>
+</ul></td></tr>
+<tr><td colspan="3"></td></tr>
+<tr><td colspan="3"><h4><a name="marea-cobb" href="#marea-cobb">#</a> Marea Cobb</h4></td></tr>
```
lmroz: Now some members can be skipped https://github.com/lmroz/org2docmd/blob/master/maintainersgen/consts.go#L68 [WIP version]

kindermoumoute:
Do we want to update GoMit in the same time? https://github.com/intelsdi-x/gomit#maintainers

bjray:
@lmroz - sorry to weigh in so late on this, but I would like to simplify this list the maintainer's name and github id in a simple plain text format without the HTML formatting.  The goal is to make changes to this list extremely simple.  Also, I will have the maintainer's list updated shortly.

lmroz:
@bjray I've uploaded the almost simplest version possible. It's still based on actual contents of github gropus, because I'm assuming after updating these groups simply generating this list again will make contents appropriate.

bjray:
@lmroz - thanks for the update.  It looks good for now.  We will come back later with a reduced maintainer list next week.

PR #1313, opened 2016-10-27
Fixes #1134, fixed support for windowed schedule

### LINKED ISSUE #1134: A task manifest does not support "windowed" schedule types
For a task manifest, the manifest includes a schedule plus the workflow. When the task manifest is unmarshalled, the schedule is unmarshalled into the following struct.

``` go
type Schedule struct {
    // Type specifies the type of the schedule. Currently, the type of "simple", "windowed" and "cron" are supported.
    Type string
    // Interval specifies the time duration.
    Interval string
    // StartTime specifies the beginning time.
    StartTime *time.Time
    // StopTime specifies the end time.
    StopTime *time.Time
}
```

For a `windowed` schedule type, the schedule would need to include Type, Interval, StartTime, and StopTime.

Golang json encoder does not support `time.Time` type by default, and to correctly unmarshal into a time.Time type requires writing a UnmarshalJSON function on the Schedule type, which currently does not exist.

katarzyna-z:
Fixes #1134 

Summary of changes:
- Fixed support for windowed schedule:
  - Added tags for JSON parser in Schedule struct (mgmt/rest/client/task.go). JSON parser requires tags to deal with unmarshalling into  Schedule struct,
  - unified structures for schedule configuration,
- Added user friendly errors to inform about missing parameters in configuration of schedule,
- Tests to cover changes,
- Updated documentation for windowed schedule.

Testing done:
- small, medium, legacy tests,
- manual tests

@intelsdi-x/snap-maintainers

### DIFF
--- a/docs/TASKS.md
+++ b/docs/TASKS.md
@@ -51,7 +51,28 @@ The header contains a version, used to differentiate between versions of the tas
 
 The schedule describes the schedule type and interval for running the task.  The type of a schedule could be a simple "run forever" schedule, which is what we see above as `"simple"` or something more complex.  Snap is designed in a way where custom schedulers can easily be dropped in.  If a custom schedule is used, it may require more key/value pairs in the schedule section of the manifest.  At the time of this writing, Snap has three schedules:
 - **simple schedule** which is described above,
-- **window schedule** which adds a start and stop time,
+- **window schedule** which adds a start and stop time for the task. The time must be given as a quoted string in [RFC 3339](https://www.ietf.org/rfc/rfc3339.txt) format, for example with specific timezone offset:
+```json
+    "version": 1,
+    "schedule": {
+        "type": "windowed",
+        "interval": "1s",
+        "start_timestamp": "2016-10-27T16:39:57+01:00",
+        "stop_timestamp": "2016-10-28T16:39:57+01:00"
+    },
+    "max-failures": 10,
+```
+or without timezone offset (in that cases uppercase'Z' must be present):
+```json
+    "version": 1,
+    "schedule": {
+        "type": "windowed",
+        "interval": "1s",
+        "start_timestamp": "2016-10-27T16:39:57Z",
+        "stop_timestamp": "2016-10-28T16:39:57Z"
+    },
+    "max-failures": 10,
+```
 - **cron schedule** which supports cron-like entries in ```interval``` field, like in this example (workflow will fire every hour on the half hour):
 ```json
     "version": 1,
--- a/mgmt/rest/rbody/task.go
+++ b/mgmt/rest/rbody/task.go
@@ -222,13 +222,11 @@ func assertSchedule(s schedule.Schedule, t *AddScheduledTask) {
 		}
 		return
 	case *schedule.WindowedSchedule:
-		startTime := v.StartTime.Unix()
-		stopTime := v.StopTime.Unix()
 		t.Schedule = &core.Schedule{
 			Type:           "windowed",
 			Interval:       v.Interval.String(),
-			StartTimestamp: &startTime,
-			StopTimestamp:  &stopTime,
+			StartTimestamp: v.StartTime,
+			StopTimestamp:  v.StopTime,
 		}
 		return
 	case *schedule.CronSchedule:
--- a/cmd/snapctl/task.go
+++ b/cmd/snapctl/task.go
@@ -170,15 +170,15 @@ func (t *task) setWindowedSchedule(start *time.Time, stop *time.Time, duration *
 		// if start is set and stop is not then use duration to create stop
 		if start != nil && stop == nil {
 			newStop := start.Add(*duration)
-			t.Schedule.StartTime = start
-			t.Schedule.StopTime = &newStop
+			t.Schedule.StartTimestamp = start
+			t.Schedule.StopTimestamp = &newStop
 			return nil
 		}
 		// if stop is set and start is not then use duration to create start
 		if stop != nil && start == nil {
 			newStart := stop.Add(*duration * -1)
-			t.Schedule.StartTime = &newStart
-			t.Schedule.StopTime = stop
+			t.Schedule.StartTimestamp = &newStart
+			t.Schedule.StopTimestamp = stop
 			return nil
 		}
 		// otherwise, the start and stop are both undefined but a duration was passed in,
@@ -187,19 +187,19 @@ func (t *task) setWindowedSchedule(start *time.Time, stop *time.Time, duration *
 		// and the duration
 		newStart := time.Now().Add(createTaskNowPad)
 		newStop := newStart.Add(*duration)
-		t.Schedule.StartTime = &newStart
-		t.Schedule.StopTime = &newStop
+		t.Schedule.StartTimestamp = &newStart
+		t.Schedule.StopTimestamp = &newStop
 		return nil
 	}
 	// if a start date/time was specified, we will use it to replace
 	// the current schedule's start date/time
 	if start != nil {
-		t.Schedule.StartTime = start
+		t.Schedule.StartTimestamp = start
 	}
 	// if a stop date/time was specified, we will use it to replace the
 	// current schedule's stop date/time
 	if stop != nil {
-		t.Schedule.StopTime = stop
+		t.Schedule.StopTimestamp = stop
 	}
 	// if we get this far, then just return a nil error (indicating success)
 	return nil
--- a/core/schedule.go
+++ b/core/schedule.go
@@ -21,21 +21,26 @@ package core
 
 import (
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/intelsdi-x/snap/pkg/schedule"
 )
 
 type Schedule struct {
-	Type           string `json:"type,omitempty"`
-	Interval       string `json:"interval,omitempty"`
-	StartTimestamp *int64 `json:"start_timestamp,omitempty"`
-	StopTimestamp  *int64 `json:"stop_timestamp,omitempty"`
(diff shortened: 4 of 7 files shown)


docs/TASKS.md
```diff
@@ -51,7 +51,17 @@ The header contains a version, used to differentiate between versions of the tas
...
 The schedule describes the schedule type and interval for running the task.  The type of a schedule could be a simple "run forever" schedule, which is what we see above as `"simple"` or something more complex.  Snap is designed in a way where custom schedulers can easily be dropped in.  If a custom schedule is used, it may require more key/value pairs in the schedule section of the manifest.  At the time of this writing, Snap has three schedules:
 - **simple schedule** which is described above,
-- **window schedule** which adds a start and stop time,
+- **window schedule** which adds a start and stop time for the task. The time must be given as a quoted string in [RFC 3339](https://www.ietf.org/rfc/rfc3339.txt) format, for example:
+```json
+    "version": 1,
+    "schedule": {
+        "type": "windowed",
+        "interval": "1s",
+        "start_timestamp": "2016-10-27T16:39:57+01:00",
```
candysmurf: can you give another alternative by specifying Z/z?

mgmt/rest/rbody/task.go
```diff
@@ -222,13 +222,14 @@ func assertSchedule(s schedule.Schedule, t *AddScheduledTask) {
...
-		stopTime := v.StopTime.Unix()
+		startTime := v.StartTime
+		stopTime := v.StopTime
+
 		t.Schedule = &core.Schedule{
 			Type:           "windowed",
 			Interval:       v.Interval.String(),
-			StartTimestamp: &startTime,
-			StopTimestamp:  &stopTime,
+			StartTimestamp: startTime,
```
candysmurf: startTime and stopTime only used once. It seems using v.StartTime and v.StopTime should suffice.

katarzyna-z: Thanks for catching it. Corrected.

docs/TASKS.md
katarzyna-z: I added example with 'Z', please check if now it looks good.

katarzyna-z:
@candysmurf 

> If stop time is current but the start time is in the past, it ran. 

I see the following test case in the code
```
Convey("start time in past is ok (as long as window ends in the future)", func() {
			start := time.Now().Add(time.Second * -10)
			stop := time.Now().Add(time.Second * 10)
			w := NewWindowedSchedule(time.Millisecond*100, &start, &stop)
			err := w.Validate()
			So(err, ShouldEqual, nil)
})
```
 but the boundary for the now and the past could be thin if you set current time in the task manifest as stop time... I don't know how you set the stop time as current. Please check if it works as it was desired.

candysmurf:
@katarzyna-z, thanks for your work. I tried a few scenarios. Issues are not necessary caused by this PR.  

* When 2 tasks are already running, when posting a new task, the newly created task is in the "stopped" state:

```
 curl -vPOST http://localhost:8181/v1/tasks -d @task-mock-file.json --header "Content-Type: application/json"
```

* Error messages are different depending on the way to create tasks.
**When creating from snapctl with options:**
```
▶ snapctl task create -t ~/task/task-mock-file-1.json -i 2s --start-time 2016-11-01T16:39:57Z
Error creating task:
parsing time "2016-11-01T16:39:57Z": hour out of range
```
**When creating from task manifest only**
```
▶ snapctl task create -t ~/task/task-mock-file-1.json
Usage error (missing interval value); when constructing a new task schedule an interval must be provided
```

The task manifest used:
```
 "schedule": {
        "type": "windowed",
        "start_timestamp": "2016-10-17T16:39:57Z"
    },
```

Same for the interval. E.g:
```
▶ snapctl task create -t ~/task/task-mock-file-1.json -i 2s
Error creating task:missing parameter/parameters in configuration of windowed schedule,start_timestamp: <nil>, stop_timestamp: <nil>, interval: 2s
```

* When using interval only from CLI as below, it gave a weird message. There is no indication of what's a good format. 

```
▶ snapctl task create -t ~/task/task-mock-file-1.json -i 2s --start-date 2016-10-31
Error creating task:
parsing time "2016-10-31": month out of range
```

There is the inconsistency between the existing code base and your PR. How do you like to address this? Should we open a new issue to address these?

katarzyna-z:
For those errors :
```
snapctl task create -t ~/task/task-mock-file-1.json -i 2s --start-date 2016-10-31
Error creating task:
parsing time "2016-10-31": month out of range
```
and 
```
snapctl task create -t ~/task/task-mock-file-1.json -i 2s --start-time 2016-11-01T16:39:57Z
Error creating task:
parsing time "2016-11-01T16:39:57Z": hour out of range
```
please take a look at `mergeDataTime` function which is called from `setScheduleFromCliOptions` method, see:
```
var (
	// padding to picking a time to start a "NOW" task
	createTaskNowPad = time.Second * 1
	timeParseFormat  = "3:04PM"
	dateParseFormat  = "1-02-2006"
	unionParseFormat = timeParseFormat + " " + dateParseFormat
)
```
and
```
func mergeDateTime(tm, dt string) *time.Time {
	reTm := time.Now().Add(createTaskNowPad)
	if dt == "" && tm == "" {
		return nil
	}
	if dt != "" {
		t, err := time.Parse(dateParseFormat, dt)
		if err != nil {
			fmt.Printf("Error creating task:\n%v\n", err)
			os.Exit(1)
		}
		reTm = t
	}

	if tm != "" {
		_, err := time.ParseInLocation(timeParseFormat, tm, time.Local)
		if err != nil {
			fmt.Printf("Error creating task:\n%v\n", err)
			os.Exit(1)
		}
		reTm, err = time.ParseInLocation(unionParseFormat, fmt.Sprintf("%s %s", tm, reTm.Format(dateParseFormat)), time.Local)
		if err != nil {
			fmt.Printf("Error creating task:\n%v\n", err)
			os.Exit(1)
		}
	}
	return &reTm
}
```
you can quickly reproduce issues with this example:
```
package main

import (
	"fmt"
	"os"
	"time"
)

var (
	createTaskNowPad = time.Second * 1
	timeParseFormat  = "3:04PM"
	dateParseFormat  = "1-02-2006"
	unionParseFormat = timeParseFormat + " " + dateParseFormat
)

func mergeDateTime(tm, dt string) *time.Time {
	reTm := time.Now().Add(createTaskNowPad)
	if dt == "" && tm == "" {
		return nil
	}
	if dt != "" {
		t, err := time.Parse(dateParseFormat, dt)
		if err != nil {
			fmt.Printf("Error creating task:\n%v\n", err)
			os.Exit(1)
		}
		reTm = t
	}

	if tm != "" {
		_, err := time.ParseInLocation(timeParseFormat, tm, time.Local)
		if err != nil {
			fmt.Printf("Error creating task:\n%v\n", err)
			os.Exit(1)
		}
		reTm, err = time.ParseInLocation(unionParseFormat, fmt.Sprintf("%s %s", tm, reTm.Format(dateParseFormat)), time.Local)
		if err != nil {
			fmt.Printf("Error creating task:\n%v\n", err)
			os.Exit(1)
		}
	}
	return &reTm
}

func main() {
	fmt.Println(mergeDateTime("", "2016-10-31"))
}
```
so snapctl uses specific format  for start/stop time and date but it seems there is nowhere information about it. I think that this should be addressed as a new issue.

katarzyna-z:
Error messages are different depending on the way to create tasks. 

snapctl builds POST request which could be send also using for example curl. snapctl gives the user possibility to create API request using options. There is specific logic implemented to check options and give user some tips, see [setScheduleFromCliOptions](https://github.com/intelsdi-x/snap/blob/master/cmd/snapctl/task.go#L209) method in cmd/snapctl/tasks.go.
If user sends the raw request (for example using curl) then the layer which is created through snapctl is omitted so it seems that errors messages could be different depending on the way to create tasks because parameters are checked on different level.

But the logic implemented to check options and build schedule from cli options is quite complicated, please see [setScheduleFromCliOptions](https://github.com/intelsdi-x/snap/blob/master/cmd/snapctl/task.go#L209) method in cmd/snapctl/tasks.go especially 
this part of code:
```go
// if a start, stop, or duration value was provided, or if the existing schedule for this task
	// is 'windowed', then it's a 'windowed' schedule
isWindowed := (start != nil || stop != nil || duration != nil || t.Schedule.Type == "windowed")
```
There were some intentions and it must be checked carefully so I think this should be also addressed through a new issue.

katarzyna-z:
> When 2 tasks are already running, when posting a new task, the newly created task is in the "stopped" state:
 curl -vPOST http://localhost:8181/v1/tasks -d @task-mock-file.json --header "Content-T

The issue is more general,  if task is created through raw API request and should start in the future then the state of task is "Stopped" when the same task is created through snapctl then the state of task is "Running". I tested with this task manifest:
```json
{
    "version": 1,
    "schedule": {
        "type": "windowed",
        "interval": "1s",
        "start_timestamp": "2017-10-27T16:39:57+01:00",
        "stop_timestamp": "2017-10-28T16:39:57+01:00"
    },
    "max-failures": 10,
    "workflow": {
        "collect": {
            "metrics": {
                "/intel/mock/foo": {},
                "/intel/mock/bar": {},
                "/intel/mock/*/baz": {}
            },
            "config": {
                "/intel/mock": {
                    "name": "root",
                    "password": "secret"
                }
            },
            "process": [
                {
                    "plugin_name": "passthru",
                    "process": null,
                    "publish": [
                        {
                            "plugin_name": "mock-file",
                            "config": {
                                "file": "/tmp/snap_published_mock_file.log"
                            }
                        }
                    ]
                }
            ]
        }
    }
}
```

PR #1316, opened 2016-10-28
Fix get config policy rules from plugin catalog

kindermoumoute:
#1199 introduced a new key format for plugin catalog, that didn't got followed by #1225 .

Summary of changes:
- Now use the good format.

Testing done:
- Use default values in config policy rules.


candysmurf:
LGTM

PR #1323, opened 2016-11-01
Fix path to plugins.yml

jpetazzo:
I suppose this file was at the top-level and was moved to `docs` at some points. The link to `plugins.yml` points to `docs/plugins.yml`, and expands to `docs/docs`, which is invalid.

@intelsdi-x/snap-maintainers


candysmurf:
LGTM

Issue #1286, opened 2016-10-14
Flaky test in control_test.go


IRCody:
This test has been failing a decent amount lately. Posting an issue so we come back to it.

```
Failures:
  * /home/travis/gopath/src/github.com/intelsdi-x/snap/control/control_test.go 
  Line 1099:
  Expected: '11'
  Actual:   '1'
  (Should be equal)
```

PR #1330, opened 2016-11-03
Fixed flaky control_test

IzabellaRaulin:
Fixes https://github.com/intelsdi-x/snap/issues/1286

### Summary of changes:
- Refactored `control_test.go` especially in test case _TestCollectDynamicMetrics_ - this is a main change, because resolves issue reported in #1286   
- Others: small refactoring another tests and added additional ones

 
###  Before:
In this test  _TestCollectDynamicMetrics_ which is dedicated for dynamic metric, as a first step `GetMetrics()` is called - that returns **2 mts** and the first one is taken without any verification if it's really dynamic metric

```
mts, err := c.metricCatalog.GetMetrics(core.NewNamespace("intel", "mock", "*", "baz"), 2)
So(err, ShouldBeNil)
So(len(mts), ShouldEqual, 2)
m := mts[0]
```

For current snap-plugin-collector-mock2, GetMetrics(core.NewNamespace(**"intel", "mock", "*", "baz"**) will return the following two metrics:
     -  `/intel/mock/*/baz` - which is DYNAMIC
     - `/intel/mock/all/baz` - which is STATIC

Taking just the first element from slice (which is build based on items from mtree where maps are used) do NOT ensure if this dynamic-metric will be gathered. It's not deterministic. 

**What that caused:**
For test [snap master, control_test.go#L1099](https://github.com/intelsdi-x/snap/blob/master/control/control_test.go#L1099):
a) when mts[0] is `/intel/mock/*/baz`, CollectMetrics returns 11 metrics (TEST PASSES)
b) when mts[0] is `/intel/mock/all/baz`, CollectMetrics returns 1 metric (TEST FAILS)

### After:
Added checking if dynamic metric has been chosen for the next tests cases, see more [here](https://github.com/intelsdi-x/snap/compare/master...IzabellaRaulin:fixed_flaky_tests?expand=1#diff-f99b446c9cc93ea5ac9418ce97183447R1120)

**What benefit it brings**:
Error reported in #1286 no more occurs
  
  
  
### Testing done:
- legacy tests

@intelsdi-x/snap-maintainers

---
This change is [Reviewable](https://reviewable.io/reviews/intelsdi-x/snap/1330)

### DIFF
--- a/control/control_test.go
+++ b/control/control_test.go
@@ -136,64 +136,75 @@ func getPort() int {
 
 func TestPluginControlGenerateArgs(t *testing.T) {
 	c := New(getTestConfig())
-	err := c.Start()
-	Convey("pluginControl starts successfully", t, func() {
+	Convey("starts pluginControl", t, func() {
+		err := c.Start()
+		So(err, ShouldBeNil)
 		So(c.Started, ShouldBeTrue)
 		So(err, ShouldBeNil)
 		So(c.Name(), ShouldResemble, "control")
+		Convey("sets monitor duration", func() {
+			c.SetMonitorOptions(MonitorDurationOption(time.Millisecond * 100))
+			So(c.pluginRunner.Monitor().duration, ShouldResemble, 100*time.Millisecond)
+		})
+		c.Stop()
 	})
-	c.SetMonitorOptions(MonitorDurationOption(time.Millisecond * 100))
-	Convey("sets monitor duration", t, func() {
-		So(c.pluginRunner.Monitor().duration, ShouldResemble, 100*time.Millisecond)
-	})
-	c.Stop()
 }
 
 func TestSwapPlugin(t *testing.T) {
-	if fixtures.SnapPath != "" {
+	// These tests only work if SNAP_PATH is known.
+	// It is the responsibility of the testing framework to
+	// build the plugins first into the build dir.
+	if fixtures.SnapPath == "" {
+		t.Fatal("SNAP_PATH not set. Cannot test swapping plugins.")
+	}
+	Convey("Starting plugin control", t, func() {
 		c := New(getTestConfig())
-		c.Start()
+		err := c.Start()
+		So(err, ShouldBeNil)
 		time.Sleep(100 * time.Millisecond)
+
 		lpe := newListenToPluginEvent()
 		c.eventManager.RegisterHandler("Control.PluginsSwapped", lpe)
 
 		_, e := load(c, fixtures.PluginPathMock2)
-		Convey("Loading first plugin", t, func() {
+		Convey("Loading first plugin", func() {
 			Convey("Should not error", func() {
 				So(e, ShouldBeNil)
 			})
 		})
-		// Travis optimization: If for some reason we can not load
-		// the plugin three times, we will fail the test immediately
-		// as we wait on a channel to be closed before proceeding with
-		// additional tests. If the plugin never loads, the channel will not
-		// close and just hang the test indefinitely.
+
 		if e != nil {
-			t.FailNow()
+			t.Fatal(err)
 		}
 		<-lpe.done
-		Convey("First plugin in catalog", t, func() {
+
+		// available one plugin in plugin catalog
+		So(len(c.PluginCatalog()), ShouldEqual, 1)
+		Convey("First plugin in catalog", func() {
 			Convey("Should have name mock", func() {
 				So(c.PluginCatalog()[0].Name(), ShouldEqual, "mock")
 			})
 		})
+
 		mockRP, mErr := core.NewRequestedPlugin(fixtures.PluginPathMock1)
-		Convey("Loading a plugin should not error", t, func() {
+		Convey("Requested collector plugin should not error", func() {
 			So(mErr, ShouldBeNil)
 		})
-		err := c.SwapPlugins(mockRP, c.PluginCatalog()[0])
-		Convey("Swapping plugins", t, func() {
+
+		err = c.SwapPlugins(mockRP, c.PluginCatalog()[0])
+		Convey("Swapping plugins with a different version", func() {
 			Convey("Should not error", func() {
 				So(err, ShouldBeNil)
 			})
 		})
+
 		if err != nil {
-			t.FailNow()
+			t.Fatal(err)
 		}
 		<-lpe.done
 
 		// Swap plugin that was loaded with a different version of the plugin
-		Convey("Swapping plugins", t, func() {
+		Convey("Successfull swapping plugins", func() {
 			Convey("Should generate a swapped plugins event", func() {
 				Convey("So first plugin in catalog after swap should have name mock", func() {
 					So(c.PluginCatalog()[0].Name(), ShouldEqual, "mock")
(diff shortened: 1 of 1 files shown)


control/control_test.go
```diff
@@ -136,127 +136,133 @@ func getPort() int {
...
-			t.FailNow()
-		}
-		<-lpe.done
 
-		// Swap plugin that was loaded with a different version of the plugin
-		Convey("Swapping plugins", t, func() {
-			Convey("Should generate a swapped plugins event", func() {
-				Convey("So first plugin in catalog after swap should have name mock", func() {
+			if e != nil {
+				t.FailNow()
```
lmroz: Having `t.Fatal("err =", e)` might be helpful in quickly determining why this eventually failed.

control/control_test.go
```diff
@@ -136,127 +136,133 @@ func getPort() int {
...
+				})
 
-		Convey("Rollback failure returns error", t, func() {
-			lp := c.PluginCatalog()[0]
-			pm := new(MockPluginManagerBadSwap)
-			pm.ExistingPlugin = lp
-			c.pluginManager = pm
+			})
+			Convey("Swap plugin with a different type of plugin", func() {
+				filePath := strings.Replace(fixtures.PluginPathMock2, "snap-plugin-collector-mock2", "snap-plugin-publisher-mock-file", 1)
```
lmroz: Consider adding path to fixtures or just use `github.com/intelsdi-x/snap/plugin/helper.helper.PluginFilePath`.

control/control_test.go
```diff
@@ -136,127 +136,133 @@ func getPort() int {
...
+					So(err, ShouldNotBeNil)
+				})
 			})
+			c.Stop()
+			time.Sleep(100 * time.Millisecond)
 		})
-
-		c.Stop()
-		time.Sleep(100 * time.Millisecond)
+	} else {
```
lmroz: Generally it's better to have
```
if !happyPath {
    quit
}
code...
```
than
```
if happyPath {
    code...
} else {
    quit
}
```

control/control_test.go
```diff
@@ -136,127 +136,133 @@ func getPort() int {
...
+		fmt.Printf("SNAP_PATH not set. Cannot test swapping plugins.\n")
```
lmroz: Perhaps reporting this error using `t.Log` or even `t.Error` (last one will fail test) is better idea.

IRCody:
Thanks for tracking this down!

IzabellaRaulin:
@lmroz, thanks for your comments. I will address them, however most of them referred to existing code which is not changed by this pull request, only nested into next new goconvey. For some tests, the presented diff is hardly readable to find out what was changed. 

The diff presented by [reviewable.io](https://reviewable.io/reviews/intelsdi-x/snap/1330#-) is more readable, so I added it.

control/control_test.go
```diff
@@ -395,61 +401,101 @@ func TestLoad(t *testing.T) {
...
+				_, err := load(c, fixtures.PluginPathMock1)
+				Convey("Should not return an error", func() {
+					So(err, ShouldBeNil)
+				})
 			})
 			c.Stop()
-			time.Sleep(100 * time.Millisecond)
 		})
-		Convey("pluginControl.Load returns error with trust enabled and signing not validated", t, func() {
+		Convey("pluginControl.Load with trust disabled", t, func() {
```
IzabellaRaulin: Only diff shows that as a renaming what can be confusing. All changes in this _TestLoadWithSignedPlugins_ needs to be taken into account to catch what have been changed, mostly these three cases were reorganized and another tests cases were added - I will explain that below:

#### Before there was 3 tests cases (see [code](https://github.com/intelsdi-x/snap/blob/master/control/control_test.go#L406)):
1. Convey("pluginControl.Load should successufully load a signed plugin with trust enabled"
2. Convey("pluginControl.Load should successfully load unsigned plugin when trust level set to warning"
3. Convey("pluginControl.Load returns error with trust enabled and signing not validated"


#### After reorganization and adding another tests cases (see [code](https://github.com/IzabellaRaulin/snap/blob/fixed_flaky_tests/control/control_test.go#L437)):
1.Convey("pluginControl.Load with trust enabled"
1.1. Convey("Loading a signed plugin"
1.2. Convey("Loading an unsigned plugin"

2.Convey("pluginControl.Load when trust level set to warning"
2.1. Convey("Loading a signed plugin"
2.2. Convey("Loading an unsigned plugin"

3.Convey("pluginControl.Load with trust disabled
3.1. Convey("Loading a signed plugin"
3.2. Convey("Loading an unsigned plugin"

Please notice, that:
old_1 was moved to new_1.1
old_2 was moved to new 2.2
old_3 was moved to new_1.2

control/control_test.go
IzabellaRaulin: Ok, I will change that through the whole control_test file.

control/control_test.go
IzabellaRaulin: This is existing code - see on snap master [control_test.go#226](https://github.com/intelsdi-x/snap/blob/master/control/control_test.go#226). 

I suppose that adding path to fixture will involved renaming the rest of existing vars defined there. Now there is 
- `PluginPathMock1` which point to snap-plugin-collector-mock1 
- `PluginPathMock2` which point to snap-plugin-collector-mock2

So, based on what you proposed, adding mock-file path might involve renaming these two existing to distinguish which of them point to collectors/publisher. Renaming will impact also on others tests files, so I suppose keeping that as it is now on master and addressing your proposal as separate change will be better approach. What do You think about it?



There are used couple tests files (not only in control_tests.go).

control/control_test.go
IzabellaRaulin: @lmroz, @IRCody, @jcooklin - simply question: **What will be better to do when SNAP_PATH is not set** - skip test (option A) or fail (option B)?

**Option A:**
```
       if fixtures.SnapPath == "" {
              t.Log("SNAP_PATH not set. Cannot test swapping plugins.\n")
              t.SkipNow()
        }
```
results:
image

**Option B**:
```
 if fixtures.SnapPath == "" {
                t.Error("SNAP_PATH not set. Cannot test swapping plugins.\n")
        }
```
results:
image

In my opinion, option B is better, because:
- these tests are more like basic tests (testing loading, unloading, swapping, etc.), so passed Travis tests should equals that all of them were executed and passed. They should not be just skipped.
- lack of SNAP_PATH means that [scripts/test.sh](https://github.com/intelsdi-x/snap/blob/master/scripts/test.sh#L32) is wrong 

Let me know do You agree with me. Thanks.

control/control_test.go
IzabellaRaulin: yes, I agree

control/control_test.go
IzabellaRaulin: Changed from `t.Fail()` to `t.Fatal(e)` through the whole _control_test.go_

control/control_test.go
```diff
@@ -136,127 +136,133 @@ func getPort() int {
...
 			})
-		})
 
-		//
-		// TODO: Write a proper rollback test as previous test was not testing rollback
-		//
-
-		// Rollback will throw an error if a plugin can not unload
+			if err != nil {
+				t.FailNow()
```
IzabellaRaulin: done

control/control_test.go
lmroz: solved by next one :+1:

control/control_test.go
lmroz: Thanks for detailed explanation, I was not reading the logic, just had some time to pick simple mistakes or errors :)

candysmurf:
LGTM

PR #1345, opened 2016-11-11
(maint) remove binary from repo

nanliu:


nanliu:
In commit bb06420, we snuck a binary into the repo. This should be removed.

kindermoumoute:
LGTM

PR #1347, opened 2016-11-13
Adds support for plugin pkgs w\ multiple commands

### LINKED ISSUE #1346: Support multiple commands when packaging a plugin
Plugins that don't compile down to a static binary may need to be started by launching them through their respective runtimes.  Today the command to be executed is described with a string instead of slice.

Example: 
 - `java -jar snap-plugin-collector-xyz.jar <args>`
- `python snap-plugin-collector-xyz.py <args>`

jcooklin:
Fixes #1346 

Summary of changes:
- Adds support for plugin packages with multiple commands
  -  Some plugins need to support declaring the interpreter and the entry point along with the command line args. 

Testing done:
- Unit
- Manual

The screen capture below demonstrates creating 3 python based plugin packages, loading and starting a task which uses them **using this branch**.  
demo

@intelsdi-x/snap-maintainers

- Plugin packages need to support declaring the interpreter and the entrypoint
  along with the command line args.

### DIFF
--- a/control/plugin/execution.go
+++ b/control/plugin/execution.go
@@ -35,6 +35,7 @@ import (
 var execLogger = log.WithField("_module", "plugin-exec")
 
 type ExecutablePlugin struct {
+	name   string
 	cmd    command
 	stdout io.Reader
 	stderr io.Reader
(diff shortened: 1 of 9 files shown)


control/plugin/execution.go
```diff
@@ -76,14 +76,14 @@ func (cw *commandWrapper) Kill() error {
 func (cw *commandWrapper) Start() error { return cw.cmd.Start() }
 
 // Initialize a new ExecutablePlugin from path to executable and daemon mode (true or false)
```
andrzej-k: This comment doesn't seem to be valid anymore...

andrzej-k:
I think we may need some guide or tutorial (possibly as a part of Plugin Authoring doc) on how to package plugins that need special runtime or commands to execute - I believe ```pkg_examples``` script that you use could be a good start for how to use acbuild.

jcooklin:
@andrzej-k Plugin packaging is generically documented in [PLUGIN_PACKAGING.md](https://github.com/intelsdi-x/snap/blob/master/docs/PLUGIN_PACKAGING.md) and more specific to Python based plugins at https://intelsdi-x.github.io/snap-plugin-lib-py/plugin_authoring/packaging.html.

mgmt/rest/plugin.go
```diff
@@ -168,10 +168,10 @@ func (s *Server) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter
...
 			restLogger.Error(err)
 			restLogger.Debugf("Removing file (%s)", rp.Path())
-			err2 := os.RemoveAll(filepath.Dir(rp.Path()))
-			if err2 != nil {
-				restLogger.Error(err2)
-			}
+			// err2 := os.RemoveAll(filepath.Dir(rp.Path()))
+			// if err2 != nil {
+			// 	restLogger.Error(err2)
+			// }
```
jcooklin: It was for debugging and shouldn't be commented out.  Good catch!

PR #1349, opened 2016-11-14
SDI-2081: expand supported types for grpc

candysmurf:
Summary of changes:
- added support for uint32 and uint64

Testing done:
- small, medium
- integration test with statistics processor

@intelsdi-x/snap-maintainers


control/plugin/rpc/plugin.pb.go
```diff
@@ -99,9 +99,9 @@ func (*Time) ProtoMessage()               {}
 func (*Time) Descriptor() ([]byte, []int) { return fileDescriptor0, []int{2} }
 
 type NamespaceElement struct {
-	Value       string `protobuf:"bytes,1,opt,name=Value,json=value" json:"Value,omitempty"`
```
candysmurf: It's GRPC generated code, we cannot change it.

candysmurf: also, you know, it's similar like JSON, YAML, it has to be upper case

IRCody:
The  related PR on [plugin-lib-go](https://github.com/intelsdi-x/snap-plugin-lib-go/pull/48) is also relevant.

Is this change backwards compatible with the [cpp](https://github.com/intelsdi-x/snap-plugin-lib-cpp) lib?

candysmurf:
@IRCody, a good question.  My guess is that it's compatible as only new types are added. The definite answer will have to be tested.

candysmurf:
@IRCody, as we chatted. it's 100% sure that this is backward compatible by adding types.

jcooklin:
Is there a reason we don't also add support for all other not currently supported (proto) types (sint32, sint64, fixed32, fixed64, sfixed32, sfixed64)?  

Since we will need to update the plugin libs to support additional types adding them all at once would be easier IMO.

candysmurf:
@jcooklin, we can add them but cannot support them in plugin-lib-go? Probably other x-language repos.

jcooklin:
@candysmurf OK.  Let's not include them for now as it would imply that we support them to a greater extent than is reasonable in the short term.

kindermoumoute:
FYI, https://github.com/intelsdi-x/snap-plugin-lib-go/pull/48 needs to be merged first, so you can update the commit id for glide in this PR.  ([this line](https://github.com/intelsdi-x/snap/blob/master/glide.lock#L52))

kindermoumoute:
LGTM

PR #1404, opened 2016-12-01
Provides additional information about check location in code - WIP

lmroz:
This is WIP, done in meanwhile to easier debugging the issues. Lacks tests and anything besides basic functionality.
Summary of changes:
- Adds info to log about file location
- Log contains path of execution that causes the message to appear

Testing done:
- Close to nothing, it's very early stage of development :)

@intelsdi-x/snap-maintainers

### DIFF
--- a/snapteld.go
+++ b/snapteld.go
@@ -190,7 +190,62 @@ type managesTribe interface {
 	GetMember(name string) *agreement.Member
 }
 
+type verboseErrorsHook struct {
+}
+
+func (hook *verboseErrorsHook) Levels() []log.Level {
+	return []log.Level{
+		// log.DebugLevel,
+		// log.InfoLevel,
+		// log.WarnLevel,
+		log.ErrorLevel,
+		log.FatalLevel,
+		log.PanicLevel,
+	}
+
+}
+
+func (hook *verboseErrorsHook) Fire(entry *log.Entry) error {
+	type Info struct {
+		FullPath, Func string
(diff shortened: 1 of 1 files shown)


IRCody:
@lmroz: I went ahead and added the WIP label. Are you looking for feedback on specific things before going forward?

lmroz:
I've done this as POC, for internal use it's good enough but when users will report that snap misbehaves such feature  might turn out useful.
Let me know if you are interested in this, so I'll refine this, add some tests and maybe config options.

IRCody:
@lmroz: If you want to get some opinions on it it would probably be useful to run through:

* An example of a problem that would be easier to debug because of this addition.
* What this adds over current logs with examples.
* Any trade-offs or decisions that were made (I noticed that what log levels it applies to seems to be one).

It also might be useful to talk about logrus hooks (or at least link to the documentation) in general and why they are a good fit for this.

snapteld.go
```diff
@@ -190,7 +190,62 @@ type managesTribe interface {
...
+			return err
+		}
+	}
+	entry.Data["source"] = infos
+
+	return nil
+}
+
 func main() {
+	log.AddHook(&verboseErrorsHook{})
```
IRCody: According to documentation this should happen in an init()?

andrzej-k:
It seems that no working is happening on this. Closing due to inactivity. Please re-open if needed.

PR #1406, opened 2016-12-01
SDI-2274: GRPC based collectors **must** provide LastAdvertisedTime o…

### LINKED ISSUE #1394: GRPC based collectors **must** provide LastAdvertisedTime or framework panics
The client call to GetMetricTypes sets the LastAdvertisedTime ([here](https://github.com/intelsdi-x/snap/blob/master/control/plugin/client/grpc.go#L256)) but this doesn't happen for the collect call.

candysmurf:
Fixes #1394 

Summary of changes:
- added check for metric timeStamp and lastAdvertisedTime.
- set it to time.Now() if they are nil.

Testing done:
- small and medium

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/plugin/client/grpc.go
+++ b/control/plugin/client/grpc.go
@@ -252,10 +252,6 @@ func (g *grpcClient) GetMetricTypes(config plugin.ConfigType) ([]core.Metric, er
 		return nil, errors.New(reply.Error)
 	}
 
-	for _, metric := range reply.Metrics {
-		metric.LastAdvertisedTime = ToTime(time.Now())
-	}
-
 	results := ToCoreMetrics(reply.Metrics)
 	return results, nil
 }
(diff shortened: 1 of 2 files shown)


control/plugin/client/grpc.go
```diff
@@ -305,6 +301,14 @@ func ToCoreMetrics(mts []*rpc.Metric) []core.Metric {
 }
 
 func ToCoreMetric(mt *rpc.Metric) core.Metric {
+	if ts := mt.Timestamp; ts == nil {
+		mt.Timestamp = ToTime(time.Now())
+	}
+
+	if lastAdvTime := mt.LastAdvertisedTime; lastAdvTime == nil {
```
kjlyon: Could you simplify the conditional to be like this: 
`if mt.LastAdvertisedTime == nil {...`

IRCody: This applies to the other if statement also.

kjlyon:
Cody is out on vacation, this issue looks good to me. Thank you for adding the test.

PR #1431, opened 2016-12-09
Interpolate environment variable

### LINKED ISSUE #791: Support ENV variables inside configuration files for paths
When a snapd flag is set on the command line with an environment variable as part of the value such as `snapd -a $PATH/to/somewhere` the cli package we use will expand `$PATH` before passing the value into snapd. Support should be added for expanding environment variables from configurations set inside of config files before we try to use the values inside of snapd.

The below example should work:

``` yaml

---
control:
  auto_discover_path: $SNAP_PATH/plugin
```

kindermoumoute:
Fixes #791 

Summary of changes:
- Interpolate environment variable (`$var` and `${var}`) in global config and task manifest.

Testing done:
- Build and run the daemon.
- Load global config and task manifest

Notes:

- Adding env variables in task manifest (or global config) make those files are note JSON/YAML until we interpolate variables. The question is now, where should we interpolate those variables?

### On `snapteld`
pros:
- easy to implement
cons:
- We should remove [JSON validation on `snaptel`](https://github.com/intelsdi-x/snap/blob/master/cmd/snaptel/task.go#L344). 

### On `snaptel`
pros:
- The API receive static JSON. 
cons:
- Still need to implement on `snapteld` side for global config and autodiscover path.

### On both
This PR changes both side, but we should definitely choose a side.

### DIFF
--- a/cmd/snaptel/task.go
+++ b/cmd/snaptel/task.go
@@ -322,7 +322,7 @@ func createTaskUsingTaskManifest(ctx *cli.Context) error {
 	if e != nil {
 		return fmt.Errorf("File error [%s] - %v\n", ext, e)
 	}
-
+	file = []byte(os.ExpandEnv(string(file)))
 	// create an empty task struct and unmarshal the contents of the file into that object
 	t := task{}
 	switch ext {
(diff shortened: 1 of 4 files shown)


jcooklin:
I like this approach.  The interpolation happens server side before the body is unmarshalled.  

cc: @jtlisi

kjlyon:
LGTM 👍

PR #1433, opened 2016-12-09
Add additional package info and scsi plugin.

nanliu:
@intelsdi-x/snap-maintainers

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ $ curl -s https://packagecloud.io/install/repositories/intelsdi-x/snap/script.de
 $ sudo apt-get install -y snap-telemetry
 ```
 
-Linux Mint 17/17.* (for Linux Mint 18/18.* use `dist=xenial`)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -100,7 +100,7 @@ $ curl -s https://packagecloud.io/install/repositories/intelsdi-x/snap/script.de
 $ sudo apt-get install -y snap-telemetry
 ```
 
-Linux Mint 17/17.* (for Linux Mint 18/18.* use `dist=xenial`)
+For distros that are compatible with RedHat/Ubuntu packages, use the environment variable `os=` and `dist=` to override the OS detection script. For example Linux Mint 17/17.* (use `dist=xenial` for Linux Mint 18/18.*):
```
mjbrender: Do you think we need to note we don't specifically test against these other flavors of Linux so YMMV? Or is that clearly enough noted?

nanliu: Yep good idea. I'll add a note.

PR #1440, opened 2016-12-15
Global tags

### LINKED ISSUE #1280: Snapd Config Tags and Vars
## Global Tags

**Proposition:** A configurable array of tags that are applied globally to any metrics collected by tasks scheduled on the snap agent. 
- Every metric collected and brought into the agent would have this tag appended, similar to how the `'plugin_running_on'` tag is currently applied to every collected metric
- It would exist in the config file supplied to snap upon startup and would look like this

```
"tags":{
    "foo":"bar",
    "krabby":"patty"
}
```
## Global Vars

**Proposition:** A key value pair array in the snapd config that could be accessed globally and could be inserted as dynamic variables within tasks that are loaded into the agent

Config  

```
"vars":{
    "foo":"bar",
    "krabby":"patty"
}
```

Task Example

```
{
    "version": 1,
    "schedule": {
        "type": "simple",
        "interval": "1s"
    },
    "workflow": {
        "collect": {
            "metrics": {
                "/intel/mock/foo": {<% foo %>},
            },
            "config": {
                "/intel/mock": {
                    "password": "<% bar %>",
                }
            },
            "process": null
        }
    }
}
```

**Value:** The value in these two propositions is made more clear when viewed with respect to the use of Tribe. When using tribe a task manifest loaded into an agreement is shared across the tribe. This feature allows snapd agents to differentiate themselves from other members in terms of there collection/publishing configurations but at the same time allows them to leverage shared workflow made possible by the tribe

kindermoumoute:
Fixes #1280 

Summary of changes:
- Clean up some code
- Apply standard tag before workflow tags
- Support global tags
- Add tests
- Updated documentation

How to verify it:
- each commit can be tested independently (the first 3 commits are just code clean up).
- Run Snap daemon with a config with tags, e.g.:
```json
{
    "control":{
        "tags": {
            "/intel/psutil":{
                "context": "config_example"
            }
        }
    }
}
```
- then run a task and verify published metrics are tagged.


control/config.go
```diff
@@ -351,114 +351,62 @@ func (p *pluginConfig) deletePluginConfigDataNodeFieldAll(key string) {
...
+	if res, ok := configItem.Plugins[name]; ok {
+		if res2, ok2 := res.Versions[ver]; ok2 {
+			res2.DeleteItem(key)
 			return
 		}
-		p.Publisher.All.DeleteItem(key)
+		res.DeleteItem(key)
+		return
 	}
+	p.Collector.All.DeleteItem(key)
```
IRCody: Should this be configItem.all.DeleteItem?

control/metrics.go
```diff
@@ -625,6 +625,9 @@ func addStandardAndWorkflowTags(m core.Metric, allTags map[string]map[string]str
 	if tags == nil {
 		tags = map[string]string{}
 	}
+	// apply standard tag
+	tags[core.STD_TAG_PLUGIN_RUNNING_ON] = hostname
+
```
IRCody: To be clear this is done to ensure that the tags from the workflow can overwrite this if they want to?

control/config.go
```diff
@@ -152,71 +152,6 @@ func GetDefaultConfig() *Config {
 	}
 }
 
-// UnmarshalJSON unmarshals valid json into a Config.  An example Config can be found
```
IRCody: Can you give an explanation of why this is "useless"?

control/control.go
```diff
@@ -127,6 +127,8 @@ type managesPlugins interface {
 	SetMetricCatalog(catalogsMetrics)
 	GenerateArgs(logLevel int) plugin.Arg
 	SetPluginConfig(*pluginConfig)
+	SetPluginTags(map[string]map[string]string)
+	AddStandardAndWorkflowTags(core.Metric, map[string]map[string]string) core.Metric
```
IRCody: What is the advantage of adding this to the managesPlugin interface?

control/control_test.go
```diff
@@ -71,13 +71,17 @@ func (m *MockPluginManagerBadSwap) LoadPlugin(*pluginDetails, gomit.Emitter) (*l
...
+func (m *MockPluginManagerBadSwap) teardown()                                  {}
+func (m *MockPluginManagerBadSwap) SetPluginConfig(*pluginConfig)              {}
+func (m *MockPluginManagerBadSwap) SetPluginTags(map[string]map[string]string) {}
+func (m *MockPluginManagerBadSwap) AddStandardAndWorkflowTags(met core.Metric, allTags map[string]map[string]string) core.Metric {
+	return nil
+}
+func (m *MockPluginManagerBadSwap) SetPluginLoadTimeout(int)         {}
+func (m *MockPluginManagerBadSwap) SetMetricCatalog(catalogsMetrics) {}
+func (m *MockPluginManagerBadSwap) SetEmitter(gomit.Emitter)         {}
+func (m *MockPluginManagerBadSwap) GenerateArgs(int) plugin.Arg      { return plugin.Arg{} }
```
IRCody: If this mocked plugin manager is going to be used for more things than BadSwap should it's name be changed?

examples/configs/snap-config-sample.json
```diff
@@ -1,88 +1,93 @@
...
-        "seed": "1.1.1.1:16000"
+    "tribe":{
+        "enable":true,
+        "bind_addr":"127.0.0.1",
+        "bind_port":16000,
+        "name":"localhost",
+        "seed":"1.1.1.1:16000"
     }
-}
+}
```
IRCody: Not easy to see what changed here with all the noise from whitespace changes.

control/config.go
kindermoumoute: [`json.Unmarshal`](https://github.com/intelsdi-x/snap/blob/master/pkg/cfgfile/cfgfile.go#L120-L121) decodes the json with the default algorithm, unless `UnmarshalJSON` is defined for a type. In this particular case , `control.Config.UnmarshalJSON` implements the default algorithm. The only value would be if the JSON is not valid the logs give more information. But this case is already catch by [the schema validation](https://github.com/intelsdi-x/snap/blob/master/pkg/cfgfile/cfgfile.go#L115-L119). So it sounds more like code duplicate.

control/control.go
kindermoumoute: Short answer: test won't pass.

Global tags are stored in the pluginManager; which is a `managesPlugin` interface too.

examples/configs/snap-config-sample.json
kindermoumoute: I added line 65 to 69.

control/control.go
IRCody: I don't understand "test won't pass" as an explanation. The test were passing before without that method on the interface?

control/plugin_manager.go
```diff
@@ -624,3 +638,64 @@ func (p *pluginManager) all() map[string]*loadedPlugin {
...
+	for i := range ns2 {
+		if i > len(ns1)-1 || ns1[i] != ns2[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func split(ns string) []string {
+	if len(ns) <= 1 {
```
IRCody: Can we add a comment here about why it is <= 1 and not 0.

docs/SNAPTELD_CONFIGURATION.md
```diff
@@ -134,6 +134,14 @@ control:
...
             user: tiffany
             password: new password
+
+  # tags section contains global tags that will be applied on collected metrics
+  # across tasks.
+  tags:
+    /intel/psutil:
+      datacenter: rennes
+    /intel:
+      country: france
```
IRCody: Add examples here for adding tags to all metrics.

examples/configs/snap-config-sample.yaml
```diff
@@ -103,6 +103,14 @@ control:
             user: tiffany
             password: new password
 
+  # tags section contains global tags that will be applied on collected metrics
+  # across tasks.
+  tags:
+    /intel/psutil:
+      datacenter: rennes
+    /intel:
+      country: france
```
IRCody: Duplicate the all tags from snapteld_configuration.md here?

control/plugin_manager.go
```diff
@@ -624,3 +638,64 @@ func (p *pluginManager) all() map[string]*loadedPlugin {
 	defer p.loadedPlugins.RUnlock()
 	return p.loadedPlugins.table
 }
+
+func hasPrefix(ns1 []string, ns2 []string) bool {
```
IRCody: Can this be replaced with [strings.HasPrefix](https://golang.org/pkg/strings/#HasPrefix)?

PR #1451, opened 2016-12-23
Fixes #1450 - check publish response for error

### LINKED ISSUE #1450: Errors returned in the publish (grpc) response are ignored
If a grpc based plugin returns an error in the publish reply object the framework ignores the error and continues like it was a successful execution of the workflow.  

The task given below should eventually go into a disabled state however currently the error is undetected when it's passed back in the reply object.
```
rand-file.yml
---
  version: 1
  schedule:
    type: "simple"
    interval: "1s"
  workflow:
    collect:
      metrics:
        /random/integer: {}
      publish:
        -
          plugin_name: test-file-publisher
          config:
            file: /tmp/rand-file.out
            return_error: true
```
before

jcooklin:
Fixes #1450

Summary of changes:
- [x] checks publish response (grpc client) for an error

after 

related: https://github.com/intelsdi-x/snap-plugin-lib-go/pull/54
@intelsdi-x/snap-maintainers

### DIFF
--- a/control/available_plugin.go
+++ b/control/available_plugin.go
@@ -453,8 +453,8 @@ func (ap *availablePlugins) publishMetrics(metrics []core.Metric, pluginName str
 	pool.RLock()
 	defer pool.RUnlock()
 
-	p, err := pool.SelectAP(taskID, config)
-	if err != nil {
+	p, serr := pool.SelectAP(taskID, config)
+	if serr != nil {
 		return []error{serr}
 	}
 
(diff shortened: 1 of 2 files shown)


control/available_plugin.go
```diff
@@ -457,8 +455,7 @@ func (ap *availablePlugins) publishMetrics(metrics []core.Metric, pluginName str
 
 	p, err := pool.SelectAP(taskID, config)
 	if err != nil {
-		errs = append(errs, err)
-		return errs
+		return []error{serr}
```
candysmurf: is it []error{err}?

IRCody: Good catch, I didn't notice that!

jcooklin: @candysmurf how does it look now?

PR #1456, opened 2017-01-05
(SDI-2393) Fix #1448 log and document maxpluginrestarts

### LINKED ISSUE #1448: Snap plugin 'connection is shut down' error message
There's several plugin bugs reporting the error `connection is shut down` which seems related to the go rpc client. I'm listing them below so we can centrally track them all:

https://github.com/intelsdi-x/snap-plugin-collector-snmp/issues/16
https://github.com/intelsdi-x/snap-plugin-collector-perfevents/issues/24
https://github.com/intelsdi-x/snap-plugin-publisher-opentsdb/issues/28

nanliu:
Fixes #1448

Summary of changes:
- support -1 for max_plugin_restarts, so it works better with max-failure: -1.
- improve logging of MaxPluginRestarts so users are aware why a plugin is disabled.
- document changes and relationship to max-failure settings.

Example of disabling plugin due to exceeding maxpluginrestartcount:
```
time="2017-01-04T17:31:25-08:00" level=warning msg="plugin disabled due to exceeding restart limit: 10" _block=handle-events _module=control-runner plugin-name=opentsdb plugin-version=9
```

Testing done:
- manual kill of publisher plugin

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/runner.go
+++ b/control/runner.go
@@ -244,7 +244,7 @@ func (r *runner) HandleGomitEvent(e gomit.Event) {
 		}
 
 		if pool.Eligible() {
-			if pool.RestartCount() < MaxPluginRestartCount {
+			if pool.RestartCount() < MaxPluginRestartCount || MaxPluginRestartCount == -1 {
 				e := r.restartPlugin(v.Key)
 				if e != nil {
 					runnerLog.WithFields(log.Fields{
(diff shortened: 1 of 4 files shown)


control/runner.go
```diff
@@ -256,10 +256,10 @@ func (r *runner) HandleGomitEvent(e gomit.Event) {
 				pool.IncRestartCount()
 
 				runnerLog.WithFields(log.Fields{
-					"_block":        "handle-events",
-					"event":         v.Name,
-					"aplugin":       v.Version,
-					"restart_count": pool.RestartCount(),
+					"_block":         "handle-events",
+					"plugin-name":    v.Name,
+					"plugin-version": v.Version,
```
IRCody: I'm not clear what the difference is between v.Name + v.Version and v.String. Right before we log this we log an error with "aplugin": v.String. Not sure what the difference is but it seems like we should be consistent? I tried looking into core/control_event for the definitions of DeadAvailablePluginEvent but it's not commented on what the fields are meant to hold.

nanliu: I changed to v.String since that provides more info such as plugin type.

PR #1462, opened 2017-01-09
Add debug log instructions for Snap.

nanliu:
Summary of changes:
- add debug output example for Snap
- use `--log-path ''` for debugging examples since we need to override the package config file defaults.

@intelsdi-x/snap-maintainers

### DIFF
--- a/docs/SNAPTEL.md
+++ b/docs/SNAPTEL.md
@@ -21,19 +21,23 @@ limitations under the License.
 A powerful telemetry framework
 
 ## Usage
+The `snap-telemetry` package installs `snaptel` in `/usr/local/sbin/snaptel`, and `snapteld` in `/usr/local/sbin/snapteld`. Either ensure `/usr/local/bin:/usr/local/sbin` is in your path, or use fully qualified filepath to the `snaptel`/`snapteld` binary:
+
 ```
-$ $SNAP_PATH/bin/snaptel [global options] command [command options] [arguments...]
+$ snaptel [global options] command [command options] [arguments...]
 ```
+
 ### Global Options
 ```
 --url, -u 'http://localhost:8181'    Sets the URL to use [$SNAP_URL]
 --insecure                           Ignore certificate errors when Snap's API is running HTTPS
 --api-version, -a 'v1'               The Snap API version
---password, -p			             Password for REST API authentication
---config, -c 			             Path to a config file [$SNAPTEL_CONFIG_PATH]
+--password, -p                       Password for REST API authentication
+--config, -c                         Path to a config file [$SNAPTEL_CONFIG_PATH]
 --help, -h                           show help
 --version, -v                        print the version
 ```
+
 ### Commands
 ```
 metric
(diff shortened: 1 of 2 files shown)


nanliu:
I added a small section on getting debug logs. We didn't use -o '' in the past, but since people installing snap-telemetry as a package, if we don't specify this option, they will get logs rerouted to /var/log/snap/snapteld.log because that's the setting in snapteld.conf. This is a bit more verbose, but will always behave the way we expected when we ask end users to collect debug logs.

cc: @mjbrender

docs/SNAPTEL.md
```diff
@@ -21,9 +21,12 @@ limitations under the License.
 A powerful telemetry framework
 
 ## Usage
+The `snap-telemetry` package installs `snaptel` in `/usr/local/sbin/snaptel`, and `snaptel` in `/usr/local/sbin/snapteld`. Either ensure `/usr/local/bin:/usr/local/sbin` is in your path, or use fully qualified filepath to the `snaptel`/`snapteld` binary:
```
IRCody: The second `snaptel` should be `snapteld.`

docs/SNAPTEL.md
```diff
@@ -103,7 +111,7 @@ Example Usage
 
 In one terminal window, run snapteld (log level is set to 1 and signing is turned off for this example):
 ```
-$ $SNAP_PATH/bin/snapteld -l 1 -t 0
+$ snapteld -l 1 -o '' -t 0
```
IRCody: This works because we are overriding anything set explicitly in a config with '' and there is a check somewhere that if the value is empty we go to stdout?

Also does it make sense to be explicit in the arguments here so that it's more clear what is going on?

docs/SNAPTELD.md
```diff
@@ -57,38 +59,92 @@ $ $SNAP_PATH/bin/snapteld [global options] command [command options] [arguments.
...
-$SNAP_PATH/bin/snapteld
-$SNAP_PATH/bin/snapteld -log-level 4
-$SNAP_PATH/bin/snapteld -l 1 -t 2 -k <keyringPath>
-$SNAP_PATH/bin/snapteld -a $SNAP_PATH/plugins/
-$SNAP_PATH/bin/snapteld --version
+$ snapteld
+$ snapteld --version
+$ snapteld --log-level 4
+$ snapteld --auto-discover /opt/snap/plugins/
+$ snapteld -l 1 -t 2 -k <keyringPath>
```
IRCody: Same question here w/r/t verbose command flags.

docs/SNAPTEL.md
nanliu: Yeah, because our default package configuration is `log-path: /var/log/snap/snapteld.log` (which is sensible), we must override `log-path` to '' to see the output on the cli. Otherwise snap will read the default config and write logs to `/var/log` and users mistakenly think nothing is happening on the system.

nanliu:
Good catch, I've updated the PR based on feedback.

docs/SNAPTEL.md
```diff
@@ -184,14 +192,14 @@ and then:
...
+$ snaptel plugin load /opt/snap/plugin/snap-plugin-collector-mock1
+$ snaptel plugin load /opt/snap/plugin/snap-plugin-processor-passthru
+$ snaptel plugin load /opt/snap/plugin/snap-plugin-publisher-mock-file
+$ snaptel plugin list
+$ snaptel task create -t mock-file.json
+$ snaptel task create -w workflow.json -i 1s -d 10s
+$ snaptel task list
+$ snaptel plugin unload -t collector -n mock -v <version>
+$ snaptel plugin unload -t processor -n passthru -v <version>
+$ snaptel plugin unload -t publisher -n publisher -v <version>
```
IzabellaRaulin: @nanliu, could you check twice plugin unload command. I checked this trying to unload some loaded plugin and this command with flags **is no longer valid** as I supposed (see below my reasoning step by step):

a) load plugin
```
$ snaptel plugin load build/linux/x86_64/snap-plugin-collector-perfevents
Plugin loaded
Name: perfevents
Version: 8
Type: collector
Signed: false
Loaded Time: Wed, 11 Jan 2017 11:35:57 CET
```
b) unload plugin (deprecated command with flags)
```
$ snaptel plugin unload -t collector -n perfevents -v 8
Incorrect Usage: flag provided but not defined: -t

NAME:
   snaptel plugin unload - unload <plugin_type> <plugin_name> <plugin_version>

USAGE:
   snaptel plugin unload [arguments...]
flag provided but not defined: -t
```
**c) unload plugin with proper command**
```
$ snaptel plugin unload collector perfevents 8

Plugin unloaded
Name: perfevents
Version: 8
Type: collector
```
Based on that, please remove flags from all unloading plugin commands.

docs/SNAPTELD.md
```diff
@@ -57,38 +59,92 @@ $ $SNAP_PATH/bin/snapteld [global options] command [command options] [arguments.
...
+    ```
+    RedHat 7/Ubuntu 16.04:
+    ```
+$ systemctl stop snap-telemetry
+$ pgrep snap
+    ```
+
+* run snap with debug log `--log-level 1`, no log file `--log-path ''`, along with any other custom startup options (it will use the snapteld.conf file settings if an option is omited):
+    ```
+$ /usr/local/sbin/snapteld --log-level l 1 --log-path '' --plugin-trust 0
```
IzabellaRaulin: This command: 
` /usr/local/sbin/snapteld --log-level l 1 --log-path '' --plugin-trust 0` 
should be without this single "l" letter like below: 
` /usr/local/sbin/snapteld --log-level 1 --log-path '' --plugin-trust 0` 

Also, please check is it fine in this place to skip the part "/usr/local/sbin/" in this command (just light suggestion)

docs/SNAPTEL.md
```diff
@@ -72,9 +78,10 @@ watch        watch <task_id>
 enable       enable <task_id>
 help, h      Shows a list of commands or help for one command
 ```
+
 #### plugin
 ```
-$ $SNAP_PATH/bin/snaptel plugin command [command options] [arguments...]
+$ snaptel plugin command [command options] [arguments...]
```
IzabellaRaulin: Please, update commands for snaptel plugin (unchanged fragment below), especially for `snaptel plugin unload` (it's related to my next comment)

nanliu:
@IzabellaRaulin, yeah looks like lots of options have been added/modified, so I recreated the docs based on the current --help output.

docs/SNAPTEL.md
IzabellaRaulin: Thanks, now looks fine

PR #1464, opened 2017-01-10
PoC v2 API

kindermoumoute:
## Definitions
- **v2 API**: current API exposes endpoint with `v1` in the URL (e.g.`http://localhost:8181/v1/metrics`). In this way, when we want to drastically change the API, we can create routes with a new version `v2` in the URL (e.g.`http://localhost:8181/v2/metrics`), without breaking the v1 routes.

## Motivation
- Remove CLI logic from the API
  - API response contains [Meta fields](https://github.com/intelsdi-x/snap/blob/master/mgmt/rest/rbody/body.go#L51-L56) that only exist to return messages to the CLI. Actually, those messages should be generated by the CLI.
- We may change the API for the next Tribe.

## Benefits
- Encourage Snap contributions
  - By making the code base clearer, we reduce the amount of time needed to implement new features / fix bugs.
- Prevent bugs in the API
  - The ultimate goal is to split API and CLI code bases, in order to be sure the API not only work with our CLI, but also with any client / curl request.

## PoC
- Reorganize files for identifying v1 API and v2 api code
  - Introducing new extension `*_v2.go` and `*_v1.go`. v1 files have dependencies on v2, but v2 files don't have dependencies to v1 files. Same for `*_v2_test.go` and `*_v1_test.go`
  - v2 handlers and functions are named `<name>V2`.
  - Introduce packages `mgmt/rest/v2/fixtures` and `mgmt/rest/v2/rbody`.
- Implement GET /v2/metrics and GET /v2/tasks
  - Response body doesn't have meta fields anymore.
- Add medium tests
  - Copied existing test for the v1 to the v2, and updated expected response bodies.
  - Changed legacy test to medium, because they share same context and work the same way (not testing the same thing).

### How to test it
- Build and run snapteld
- `curl -v http://localhost:8181/v2/tasks` or `curl -v http://localhost:8181/v2/metrics`
- Compare results with the v1


IzabellaRaulin:
@kindermoumoute - why did you change `mgmt/rest/config_test.go` to `mgmt/rest/config_v2_test.go`? I suppose this should be renamed to `mgmt/rest/config_v1_test.go`, but maybe I am wrong. Could you check this?

kindermoumoute:
> v1 files have dependencies on v2, but v2 files don't have dependencies to v1 files.

@IzabellaRaulin That's what I had in mind, because this file is used by v2 tests too. But now you say that I could leave the name to `mgmt/rest/config_test.go` (this file only contains a mock config).

candysmurf:
@kindermoumoute, I haven't looked all changes yet. One question first. If v1 and v2 are two different folders, file names won't need to be changed. Don't you think it's easier to manage?

kindermoumoute:
@candysmurf http handler are method of the server:
```Go
func (s *Server) myHandler(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
```

For that reason, we can't have v1 and v2 handlers in different package without making things complicated.

However, I already split v1 and v2 dependencies in 2 folders:
```
rest
├── fixtures
├── rbody
├── v2
    ├── fixtures
    └── rbody
```

candysmurf:
@kindermoumoute, I disagree. I think the design should have a clear separation of versions. At the top level, there are three folders, v1, v2, tools/libs. Where tools/libs are absolutely common tools/utils shared by v1 and v2. Inside v1 and v2, there may be a bit duplication, so be it! It's easy to maintain and retire. Don't mingle them together. v1, v2,...vx, all should run as different instances not all together.

nanliu:
Do we need to keep rbody for v2? I thought one of the reason for v2 was to move away from this.

kindermoumoute:
@nanliu `rbody` means response body. In this PoC the response is static though.

candysmurf:
@kindermoumoute, thanks for showing me problems and the proposed solutions. Unfortunately, I'm standing on my ground. If team all agree with this approach, I'm fine. Let me give more details of my thoughts.

**Pure API Versioning**

If Snap is going to claim that Snap supports API versioning, then I suggest structuring our code tree somehow like:
snap
├── libs
├── v1
├── v2
└── vn

Libs folder has common utility code for all versions where v1, v2,...vn are different versions of APIs. Although there will be a bit duplication, but it's easy to do EOL support and releases. Note that each version is it own standalone daemon.

**New Release**
Often time new releases need to not break existing clients but also be able to introduce new features, API changes, even refactor. Snap cannot force any existing clients to upgrade to a newer release but only can announce EOF support of blah by xx time. How can we have a robust release support for different releases of Snap?  I suggest:

* For each release, we always cut a release branch. Any bugfixes for that release need to be fixed on that branch and merged back to the master branch.
* master is always a dev branch. Snap always cuts a new release from the master branch. Master branch allows breakage changes and supports nightly builds.
* cuts throttle branch for customer specific release from the master. Depending on the scope of bug fixes, if it's customer specific, it'll never be merged back to the master.

Please share your thoughts.

nanliu:
@candysmurf, this proposal maintains backwards compatibility with some tools that have used the existing Snap client as a library and expect the current v1 endpoints while allowing us to move forward. Snap daemon can service both /v1/* and /v2/* requests without requiring two different Snap binaries. Similar to [Docker api](https://docs.docker.com/v1.9/engine/reference/api/docker_remote_api/#v1-21-api-changes), snap version != snap api version, and for a given snap daemon version, it does not mean the service only respond to one specific snap api version request. We probably also need a similar notice as docker:

>  Docker’s Remote API uses an open schema model. In this model, unknown properties in incoming messages are ignored. Client applications need to take this behavior into account to ensure they do not break when talking to newer Docker daemons.

Let's discuss this in our next grooming session, and make sure we understand the pro/cons of each approach.

candysmurf:
it was a great discussion at our grooming session. It's OK that we go with API versioning for now. To clarify one request. I think the folder separation of v1, v2, ...vn,  is better than using file names with "metric_v1.go" and "metric_v2.go" if that's possible. Folder separation may introduce a bit duplication but it'll save us time ahead.

jcooklin:
To organize the API versions into packages and avoid having metric_v1.go and metric_v2.go side-by-side we could create fields for `v1` and `v2` in the server struct where the type would be a new api struct implementing the version in the new package (mgmt/rest/v2).  This would allow the relevant code to be in separate packages.  Today the management interfaces (control, scheduler, etc) are added to the server instance by calling BindMetricManager.  In this call we would bind them to the new v2 and V1 fields of the server struct.  

Before:
```
type Server struct {
	mm         managesMetrics
	mt         managesTasks
	tr         managesTribe
	mc         managesConfig
	n          *negroni.Negroni
	r          *httprouter.Router
        ...
}

func (s *Server) addRoutes() {
	s.r.GET("/v2/metrics", s.getMetricsV2)
	s.r.GET("/v1/plugins", s.getPlugins)
        ...
}
```

After:
```
type Server struct {
	v2         v2.Api
	v1         v1.Api	
	n          *negroni.Negroni
	r          *httprouter.Router
        ...
}
func (s *Server) addRoutes() {
	// plugin routes
	s.r.GET("/v1/plugins", s.v1.getPlugins)
	s.r.GET("/V2/metrics", s.v2.getPlugins)
        ...
}
```

candysmurf:
@kindermoumoute, will you like to rename two files back: tribe_v1_test.go and metric_v1.go?

kindermoumoute:
Here is the plan for the incoming work on v2:
1. merge this PoC on v2_rest_api
2. I'll open another big PR with all v2 route implementations (I'll try to keep small changes in multiple commits)
3. merge that big PR on v2_rest_api
4. new PR with restructured files
5. merge restructured files on v2_rest_api
6. merge v2_rest_api to master

candysmurf:
@kindermoumoute, Look forward to your coming PRs. LGTM

PR #1474, opened 2017-01-17
V2 rest api

kindermoumoute:
Summary of changes:
- Implement all v2 routes
- Re-structure files to have v1 and v2 routes in a different package.

How to test it:
- review commit code one by one and read commit description.
- curl the API with v1 and v2 routes.


mgmt/rest/v2/response/response.go
```diff
@@ -2,7 +2,7 @@
 http://www.apache.org/licenses/LICENSE-2.0.txt
 
 
-Copyright 2015 Intel Corporation
+Copyright 2016 Intel Corporation
```
kjlyon: Why change in just this file? And should it be 2017 now?

mgmt/rest/v2/config.go
```diff
@@ -0,0 +1,158 @@
...
+	if errCode != 0 && err != nil {
+		response.Write(400, response.FromError(err), w)
+		return
+	}
+
+	var res cdata.ConfigDataNode
+	if styp == "" {
+		res = s.configManager.MergePluginConfigDataNodeAll(src)
+	} else {
+		res = s.configManager.MergePluginConfigDataNode(typ, name, iver, src)
```
candysmurf: what if the name is empty

mgmt/rest/v2/metric.go
```diff
@@ -0,0 +1,178 @@
+/*
+http://www.apache.org/licenses/LICENSE-2.0.txt
+
+
+Copyright 2015 Intel Corporation
```
candysmurf: 2017

mgmt/rest/v2/config.go
```diff
@@ -0,0 +1,158 @@
+/*
+http://www.apache.org/licenses/LICENSE-2.0.txt
+
+
+Copyright 2015 Intel Corporation
```
candysmurf: 2017

mgmt/rest/v2/metric.go
```diff
@@ -0,0 +1,178 @@
...
+func (s *V2) getMetrics(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	ver := 0 // 0: get all metrics
+
+	// If we are provided a parameter with the name 'ns' we need to
+	// perform a query
+	q := r.URL.Query()
+	v := q.Get("ver")
+	ns_query := q.Get("ns")
+	if ns_query != "" {
+		ver = 0 // 0: get all versions
```
candysmurf: assign it again 45. It was not changed at line 37

mgmt/rest/v2/metric.go
```diff
@@ -0,0 +1,178 @@
...
+		if v != "" {
+			var err error
+			ver, err = strconv.Atoi(v)
```
candysmurf: use shorthand

mgmt/rest/v2/plugin.go
```diff
@@ -0,0 +1,412 @@
...
+		if err != nil {
+			return "", err
+		}
+	}
+	return f.Name(), nil
+}
+
+func (s *V2) unloadPlugin(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	plName := p.ByName("name")
+	plType := p.ByName("type")
```
candysmurf: no name and type empty check and handling

mgmt/rest/v2/plugin.go
```diff
@@ -0,0 +1,412 @@
...
+	plType := params.ByName("type")
+	nbFilter := Btoi(plName != "") + Btoi(plType != "")
+
+	if _, detail := r.URL.Query()["running"]; detail {
+		// get running plugins
+		plugins := runningPluginsBody(r.Host, s.metricManager.AvailablePlugins())
+		filteredPlugins := []response.RunningPlugin{}
+		if nbFilter > 0 {
+			for _, p := range plugins {
+				if nbFilter == 1 && (p.Name == plName || p.Type == plType) || nbFilter == 2 && (p.Name == plName && p.Type == plType) {
```
candysmurf: can this if be simplified as if (nbFilter == 1  || nbFilter == 2) && ()

mgmt/rest/v2/plugin.go
```diff
@@ -0,0 +1,412 @@
...
+		}
+		response.Write(200, filteredPlugins, w)
+	} else {
+		// get plugins from the plugin catalog
+		plugins := pluginCatalogBody(r.Host, s.metricManager.PluginCatalog())
+		filteredPlugins := []response.Plugin{}
+
+		if nbFilter > 0 {
+			for _, p := range plugins {
+				if nbFilter == 1 && (p.Name == plName || p.Type == plType) || nbFilter == 2 && (p.Name == plName && p.Type == plType) {
```
candysmurf: simply if  as above

mgmt/rest/v2/plugin.go
```diff
@@ -0,0 +1,412 @@
...
+		// get running plugins
+		plugins := runningPluginsBody(r.Host, s.metricManager.AvailablePlugins())
+		filteredPlugins := []response.RunningPlugin{}
+		if nbFilter > 0 {
+			for _, p := range plugins {
+				if nbFilter == 1 && (p.Name == plName || p.Type == plType) || nbFilter == 2 && (p.Name == plName && p.Type == plType) {
+					filteredPlugins = append(filteredPlugins, p)
+				}
+			}
+		} else {
```
candysmurf: get rid of else by initializing it first

mgmt/rest/v2/plugin.go
```diff
@@ -0,0 +1,412 @@
...
+		plugins := pluginCatalogBody(r.Host, s.metricManager.PluginCatalog())
+		filteredPlugins := []response.Plugin{}
+
+		if nbFilter > 0 {
+			for _, p := range plugins {
+				if nbFilter == 1 && (p.Name == plName || p.Type == plType) || nbFilter == 2 && (p.Name == plName && p.Type == plType) {
+					filteredPlugins = append(filteredPlugins, p)
+				}
+			}
+		} else {
```
candysmurf: get rid of else by initializing it first

mgmt/rest/v2/plugin.go
```diff
@@ -0,0 +1,412 @@
...
+				Name:     r.Name,
+				Type:     r.Type,
+				Default:  r.Default,
+				Required: r.Required,
+				Minimum:  r.Minimum,
+				Maximum:  r.Maximum,
+			})
+		}
+
+	} else {
```
candysmurf: get rid of else

mgmt/rest/v2/response/response.go
```diff
@@ -28,16 +28,19 @@ import (
...
 		w.WriteHeader(code)
 	}
 
-	j, err := json.MarshalIndent(body, "", "  ")
-	if err != nil {
-		panic(err)
+	if body != nil {
+		j, err := json.MarshalIndent(body, "", "  ")
+		if err != nil {
+			panic(err)
```
candysmurf: log?

mgmt/rest/v2/task.go
```diff
@@ -0,0 +1,273 @@
+/*
+http://www.apache.org/licenses/LICENSE-2.0.txt
+
+
+Copyright 2015,2016 Intel Corporation
```
candysmurf: 2017

pkg/api/api.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/julienschmidt/httprouter"
+)
+
+type API interface {
+	GetRoutes() []Route
+	BindMetricManager(Metrics)
+	BindTaskManager(Tasks)
+	BindTribeManager(Tribe)
+	BindConfigManager(Config)
+}
```
candysmurf: nice change

mgmt/rest/rest_v1_test.go
```diff
@@ -46,8 +46,8 @@ import (
 	"github.com/intelsdi-x/snap/core"
 	"github.com/intelsdi-x/snap/core/cdata"
 	"github.com/intelsdi-x/snap/core/ctypes"
-	"github.com/intelsdi-x/snap/mgmt/rest/fixtures"
-	"github.com/intelsdi-x/snap/mgmt/rest/rbody"
+	"github.com/intelsdi-x/snap/mgmt/rest/v1/fixtures"
+	"github.com/intelsdi-x/snap/mgmt/rest/v1/rbody"
```
kindermoumoute: In order to differentiate v1 and v2 dependencies / testing, I moved fixtures and rbody under the folder v1.

v2 has the same dependencies but I renamed it to mock / response.

mgmt/rest/rest_v2_test.go
```diff
@@ -50,6 +60,29 @@ var (
...
+// Since we do not have a global snap package that could be imported
+// we create a mock config struct to mock what is in snapteld.go
+
+type mockConfig struct {
+	LogLevel   int    `json:"-"yaml:"-"`
+	GoMaxProcs int    `json:"-"yaml:"-"`
+	LogPath    string `json:"-"yaml:"-"`
+	Control    *control.Config
+	Scheduler  *scheduler.Config `json:"-",yaml:"-"`
+	RestAPI    *Config           `json:"-",yaml:"-"`
```
kindermoumoute: This fields are under a `mockConfig` struct, it sounds clear to me that `mockConfig.RestAPI` is a mock config for the rest API.

mgmt/rest/rest_v2_test.go
```diff
@@ -101,29 +134,209 @@ func startV2API(cfg *mockConfig, testType string) *restAPIInstance {
...
+			// to the pipe so we can read from it in the http call
+			go func() {
+				part, err := mwriter.CreateFormFile("snap-plugins", "mock")
+				c.So(err, ShouldBeNil)
+				bufin.WriteTo(part)
+				mwriter.Close()
+				writer.Close()
+			}()
+
+			resp1, err1 := http.Post(
```
kindermoumoute: This is how we did test that on the v1 API. AFAIK I never seen this test fail, maybe @kjlyon has more inside on it. Looks good to me though.

mgmt/rest/v2/api.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"sync"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/intelsdi-x/snap/pkg/api"
+)
+
+const (
+	version = "v2"
```
kindermoumoute: > it seems better to include the version inside V1/V2 data type.

Can you develop this thought?

mgmt/rest/v2/config.go
```diff
@@ -0,0 +1,158 @@
...
+		return
+	}
+
+	typ, err := getPluginType(styp)
+	if err != nil {
+		response.Write(400, response.FromError(err), w)
+		return
+	}
+
+	name := p.ByName("name")
```
kindermoumoute: Do you mean we should skip the next block (verifying the version) if the name is empty?

mgmt/rest/v2/config.go
```diff
@@ -0,0 +1,158 @@
...
+	if sver != "" {
+		if iver, err = strconv.Atoi(sver); err != nil {
+			response.Write(400, response.FromError(err), w)
+			return
+		}
+	} else {
+		iver = -2
+	}
+
+	cdn := s.configManager.GetPluginConfigDataNode(typ, name, iver)
```
kindermoumoute: It will get every config from the type ("collector", "processor", ...).

mgmt/rest/v2/metric.go
kindermoumoute: Good catch

mgmt/rest/v2/metric.go
kindermoumoute: Doing `ver, err := strconv.Atoi(v)` will create a new variable `ver` that is not in the same context that the original one. Basically it will throw an error "variable unused".

mgmt/rest/v2/plugin.go
kindermoumoute: This is done line 225.

mgmt/rest/v2/plugin.go
kindermoumoute: The 2 following conditions are not the same:
```
p.Name == plName || p.Type == plType
p.Name == plName && p.Type == plType
```
Can't factorize it.

mgmt/rest/v2/plugin.go
kindermoumoute: That changes the behavior and would break the filter feature.

mgmt/rest/v2/plugin.go
kindermoumoute: Same answer than above.

mgmt/rest/v2/plugin.go
kindermoumoute: Same answer than above.

mgmt/rest/v2/plugin.go
kindermoumoute: Good catch 👍

mgmt/rest/rest_v2_test.go
candysmurf: agreed. object name is clear.

mgmt/rest/rest_v2_test.go
candysmurf: why did you need a goroutine without achieving anything.

mgmt/rest/v2/api.go
candysmurf: I'm not a fan of many global variables. I see the difficulty for you to change it as there are mixed methods and functions. It's ok. leave it.

mgmt/rest/v2/config.go
candysmurf: if the name is empty, will you need to execute all those statements below?

mgmt/rest/v2/plugin.go
candysmurf: I saw it later. It didn't let me remove my comments.

mgmt/rest/v2/plugin.go
candysmurf: Ah, || and &&.  it's hard to read statement though.

mgmt/rest/v2/plugin.go
candysmurf: ??

mgmt/rest/v2/api.go
kindermoumoute: It's a global constant, which is more reasonable than a global variable because it can't change. (the compiler will inline the constant everywhere it's used)

mgmt/rest/v2/plugin.go
kindermoumoute: I agree, I did that to avoid looping twice for filtering twice. I can jump a line between the 2 conditions if it makes it clearer.

mgmt/rest/v2/plugin.go
kindermoumoute: If you move that to the top it'll return every plugins + the append of the filtered plugins.

mgmt/rest/v2/config.go
kindermoumoute: If the name is empty, the version doesn't matter.

mgmt/rest/plugin_v2.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+	return plugins
+}
+
+func pluginURI(host, version string, c core.Plugin) string {
+	return fmt.Sprintf("%s://%s/%s/plugins/%s/%s/%d", protocolPrefix, host, version, c.TypeName(), c.Name(), c.Version())
+}
+
+func (s *Server) getPluginV2(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	plName := p.ByName("name")
```
kjlyon: This section (lines 324-337) is repeated in unloadPlugin, should it be broken out into a separate method? If it's only those two places it doesn't have to be, but if it is in other places too it should be broken out.

pkg/api/task.go
```diff
@@ -0,0 +1,19 @@
+package api
+
+import (
+	"github.com/intelsdi-x/snap/core"
+	"github.com/intelsdi-x/snap/core/serror"
+	"github.com/intelsdi-x/snap/pkg/schedule"
+	"github.com/intelsdi-x/snap/scheduler/wmap"
+)
+
+type Tasks interface {
```
kjlyon: 👍  I like how simple these files are

mgmt/rest/v2/api.go
```diff
@@ -0,0 +1,73 @@
...
+		// metric routes
+		api.Route{Method: "GET", Path: prefix + "/metrics", Handler: s.getMetrics},
+		api.Route{Method: "GET", Path: prefix + "/metrics/*namespace", Handler: s.getMetricsFromTree},
+
+		// task routes
+		api.Route{Method: "GET", Path: prefix + "/tasks", Handler: s.getTasks},
+		api.Route{Method: "GET", Path: prefix + "/tasks/:id", Handler: s.getTask},
+		api.Route{Method: "GET", Path: prefix + "/tasks/:id/watch", Handler: s.watchTask},
+		api.Route{Method: "POST", Path: prefix + "/tasks", Handler: s.addTask},
+		api.Route{Method: "PUT", Path: prefix + "/tasks/:id", Handler: s.updateTask},
```
kjlyon: This method `updateTask` now captures what `startTask`, `stopTask`, and `enableTask` does in V1/api.go. Does this need to be documented in some way? Will we have a section in our documentation that says what the new changes to the v2 API are and how to use them?

kindermoumoute: I'm currently working on a documentation using swagger. This will be documented. 👍

mgmt/rest/v1/api.go
```diff
@@ -34,41 +34,41 @@ func NewV1(wg *sync.WaitGroup, killChan chan struct{}, protocol string) *V1 {
...
-		api.Route{Method: "GET", Path: prefix + "/plugins", Handler: s.getPlugins},
-		api.Route{Method: "GET", Path: prefix + "/plugins/:type", Handler: s.getPlugins},
-		api.Route{Method: "GET", Path: prefix + "/plugins/:type/:name", Handler: s.getPlugins},
-		api.Route{Method: "GET", Path: prefix + "/plugins/:type/:name/:version", Handler: s.getPlugin},
-		api.Route{Method: "POST", Path: prefix + "/plugins", Handler: s.loadPlugin},
-		api.Route{Method: "DELETE", Path: prefix + "/plugins/:type/:name/:version", Handler: s.unloadPlugin},
-		api.Route{Method: "GET", Path: prefix + "/plugins/:type/:name/:version/config", Handler: s.getPluginConfigItem},
-		api.Route{Method: "PUT", Path: prefix + "/plugins/:type/:name/:version/config", Handler: s.setPluginConfigItem},
-		api.Route{Method: "DELETE", Path: prefix + "/plugins/:type/:name/:version/config", Handler: s.deletePluginConfigItem},
+		api.Route{Method: "GET", Path: prefix + "/plugins", Handle: s.getPlugins},
```
kjlyon: What was purpose for rename? Was it just for clarity?

kindermoumoute: Yeah, the type is `httprouter.Handle` so I wanted to match the type name.

mgmt/rest/plugin_v2.go
kindermoumoute: I like the idea, that's duplicate code so it makes sense to break it out. 👍

mgmt/rest/v2/plugin.go
candysmurf: isn't that the default behavior.

kindermoumoute: No, the current behavior is it returns only filtered plugins (when there is a filter).

candysmurf: ok, I'll take another look carefully

core/api/metric.go
```diff
@@ -0,0 +1,18 @@
...
+	MetricCatalog() ([]core.CatalogedMetric, error)
+	FetchMetrics(core.Namespace, int) ([]core.CatalogedMetric, error)
+	GetMetricVersions(core.Namespace) ([]core.CatalogedMetric, error)
+	GetMetric(core.Namespace, int) (core.CatalogedMetric, error)
+	Load(*core.RequestedPlugin) (core.CatalogedPlugin, serror.SnapError)
+	Unload(core.Plugin) (core.CatalogedPlugin, serror.SnapError)
+	PluginCatalog() core.PluginCatalog
+	AvailablePlugins() []core.AvailablePlugin
+	GetAutodiscoverPaths() []string
+}
```
IRCody: What is the advantage of having this api dir inside core as opposed to rest?

IRCody:
In v1 there is a metric.go, plugin.go, and plugin_test.go but in v2 there is a metric.go, metric_test.go, and plugin.go. Seems a little weird, are test missing or are different things tested between apis?

core/api/metric.go
kindermoumoute: I'll answer the 2 ways I understand your question:
1. We can't have a the API interface in `mgmt/rest` because `mgmt/rest/v2` (or v1) has a dependency on it ==> it would be a dependency loop
2. why having this package in `core/api` instead of `mgmt/rest/api`? I just thought it would make more sense because managers are implementing interfaces from `core`.

kindermoumoute:
@IRCody, tests in `plugin_test.go` are not required anymore because: (1) getPlugins got moved inside the Handler, (2) medium test are already testing the getplugins handler.

`metric_test.go` should be present in v1 too. I forgot to copy it back there.

core/api/metric.go
IRCody:
```
why having this package in core/api instead of mgmt/rest/api? I just thought it would make more sense because managers are implementing interfaces from core.
```

This seems circular since the interfaces are only in core (as opposed to rest) because they have been moved?

mgmt/rest/v2/api.go
```diff
@@ -72,3 +77,21 @@ func (s *V2) BindTribeManager(tribeManager api.Tribe) {}
...
+	if !w.(negroni.ResponseWriter).Written() {
+		w.WriteHeader(code)
+	}
+
+	if body != nil {
+		j, err := json.MarshalIndent(body, "", "  ")
+		if err != nil {
+			restLogger.Fatalln(err)
+		}
+		j = bytes.Replace(j, []byte("\\u0026"), []byte("&"), -1)
```
IRCody: Maybe a comment to explain what's going on here?

core/api/metric.go
kindermoumoute: Good point, I'll move it to `mgmt/rest/api`.

mgmt/rest/v2/task.go
```diff
@@ -29,147 +28,83 @@ import (
...
-	"github.com/intelsdi-x/snap/mgmt/rest/v2/response"
+	"github.com/intelsdi-x/snap/pkg/schedule"
+	"github.com/intelsdi-x/snap/scheduler/wmap"
 	"github.com/julienschmidt/httprouter"
 )
 
-var (
-	// The amount of time to buffer streaming events before flushing in seconds
-	StreamingBufferWindow = 0.1
+type Task struct {
```
IRCody: These exported types would be better with comments explaining expected use.

mgmt/rest/v2/api.go
```diff
@@ -0,0 +1,96 @@
...
+	version = "v2"
+	prefix  = "/" + version
+)
+
+var (
+	restLogger     = log.WithField("_module", "_mgmt-rest-v2")
+	protocolPrefix = "http"
+)
+
+type V2 struct {
```
IRCody: Should this (and V1) be unexported to avoid instantiation outside of calls to New? Maybe renamed also to avoid confusion around type V2 inside package v2?

mgmt/rest/api/api.go
```diff
@@ -0,0 +1,18 @@
...
+type API interface {
+	GetRoutes() []Route
+	BindMetricManager(Metrics)
+	BindTaskManager(Tasks)
+	BindTribeManager(Tribe)
+	BindConfigManager(Config)
+}
+
+type Route struct {
+	Method, Path string
```
IRCody: Move path to it's own line.

mgmt/rest/v2/response/metric.go
```diff
@@ -91,3 +84,8 @@ func (s StreamedMetrics) Less(i, j int) bool {
 func (s StreamedMetrics) Swap(i, j int) {
 	s[i], s[j] = s[j], s[i]
 }
+
+// cdata.ConfigDataNode implements it's own UnmarshalJSON
```
IRCody: Is this comment still true?

mgmt/rest/v2/response/response.go
```diff
@@ -17,14 +17,30 @@ See the License for the specific language governing permissions and
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/urfave/negroni"
+)
+
+func Write(code int, body interface{}, w http.ResponseWriter) {
```
IRCody: Can we comment exported methods?

mgmt/rest/v2/response/task.go
```diff
@@ -79,12 +53,35 @@ type Task struct {
 	Href               string            `json:"href"`
 }
 
+type Tasks []Task
```
IRCody: And comments on these exported types.

mgmt/rest/rest_v2_test.go
IRCody: Did this get resolved?

mgmt/rest/v2/response/metric.go
kindermoumoute: Yes. Actually, I did add the comment because unmarshalling `ConfigDataNode` is customized.

mgmt/rest/v2/api.go
kindermoumoute: Comment added.

mgmt/rest/rest_v2_test.go
kjlyon: From the comments in the file,  ` A go routine is needed since we must write the multipart file to the pipe so we can read from it in the http call ` 
I agree with @kindermoumoute, I don't recall ever seeing this fail.

mgmt/rest/v2/api.go
kindermoumoute: I didn't add a comment, instead I disabled HTML escape (forced us to unescape `&`).

mgmt/rest/v2/response/metric.go
IRCody: i.c. I was looking for PluginConfigitem unmarshall.

mgmt/rest/rest_v2_test.go
candysmurf: although goroutine is needed but there is no handling for it, test always passes. It's unrelated to Olivier's change though

kindermoumoute:
This PR will be merged with #1489.

Issue #1482, opened 2017-01-20
Impossible to stop Snap task


iwankgb:
We have a task that runs for approximately 2 minutes and then is stopped. The task is being run in a loop of configurable number of iterations. Sometimes it is not possible to stop the task.

- OS version
````
[root@*********** ~]# uname -a
Linux ******* 4.9.2-1.el7.elrepo.x86_64 #1 SMP Tue Jan 10 11:30:19 EST 2017 x86_64 x86_64 x86_64 GNU/Linux
````
- Snap version
````
[root@*********** ~]# snapteld --version
snapteld version 1.0.0
````
- Environment details (virtual, physical, etc.) - bare metal
- Steps to reproduce - keep creating and stopping Snap task.
- Expected results: - task stoppped
- Actual results: task is not stopped and pretends to be running; one of the plugins remains running despite of attempt to kill it
````
[root@*********** ~]# snaptel task list
ID 					 NAME 						 STATE 		 HIT 	 MISS 	 FAIL 	 CREATED 		 LAST FAILURE
dc2810f5-76e2-4c91-a4b4-08918d55df8f 	 serenity                                  	 Running 	 434 	 26 	 1 	 1:17PM 1-20-2017 	 CollectMetrics call error : could not obtain load metric: no history stored: could not calculate current load
````

````
[root@*********** ~]# ps ax | grep [s]nap
16600 ?        Ssl    0:13 /opt/snap/sbin/snapteld
32595 ?        Sl     0:01 /tmp/478065240/snap-plugin-publisher-influxdb {"LogLevel":5,"PingTimeoutDuration":10000000000,"NoDaemon":false,"Pprof":false}
````

@marcin-krolik suggested that it might be related to https://github.com/intelsdi-x/snap/issues/1454.

Log file is available at: https://gist.github.com/iwankgb/03ca6ac4d18cd1247fd22905645b3dbb

candysmurf:
@iwankgb, thanks for logging this issue. Do you mind to share your script for easy to reproduce this issue?

marcin-krolik:
I think I managed to debug it. Locking mutex in [control/available_plugin.go:findLatestPool](https://github.com/intelsdi-x/snap/blob/master/control/available_plugin.go#L510) 

```go
func (ap *availablePlugins) findLatestPool(pType, name string) (strategy.Pool, serror.SnapError) {
        ap.RLock()
        ap.RUnlock()
	// see if there exists a pool at all which matches name version.
[...]

PR #1487, opened 2017-01-21
Adding default timeout to Snap REST client

iwankgb:
Fixes issue of infinitely blocking HTTP connections (mitigates #1482 and similar issues).

Summary of changes:
- added default timeout of 10 seconds to secure and insecure transports

Testing done:
- added a functional test for the timeout feature

It does not seem to be possible to use `opts` parameter of `client.New() as client's `http` property is unexported and therefore unavailble from another packages.

@intelsdi-x/snap-maintainers

### DIFF
--- a/mgmt/rest/client/client.go
+++ b/mgmt/rest/client/client.go
@@ -102,6 +102,13 @@ func Username(u string) metaOp {
 	}
 }
 
+//Timeout is an option that can be provided to the func client.New in order to set HTTP connection timeout.
+func Timeout(t time.Duration) metaOp {
+	return func(c *Client) {
+		c.http.Timeout = t
+	}
+}
+
 var (
 	secureTransport = &http.Transport{
 		TLSClientConfig: &tls.Config{InsecureSkipVerify: false},
--- a/mgmt/rest/client/client_func_test.go
+++ b/mgmt/rest/client/client_func_test.go
@@ -26,6 +26,7 @@ package client
 import (
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


mgmt/rest/client/client.go
```diff
@@ -104,12 +104,14 @@ func Username(u string) metaOp {
 
 var (
 	secureTransport = &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: false},
-		IdleConnTimeout: time.Second,
+		TLSClientConfig:       &tls.Config{InsecureSkipVerify: false},
+		IdleConnTimeout:       time.Second,
+		ResponseHeaderTimeout: 10 * time.Second,
```
IRCody: Does it make sense to add this as a const somewhere so that it can be edited in one place or potentially modified via config in the future?

mgmt/rest/client/client_func_test.go
```diff
@@ -627,4 +628,20 @@ func TestSnapClient(t *testing.T) {
 		So(err, ShouldNotBeNil)
 		So(c, ShouldBeNil)
 	})
+
+	go http.ListenAndServe("127.0.0.1:65000", timeoutHandler{})
+	c, err = New("http://127.0.0.1:65000", "", true)
+	FocusConvey("Client should timeout", t, func() {
```
IRCody: What's the advantage of FocusConvey here?

IRCody:
```
It does not seem to be possible to use opts parameter of client.New() as client'shttp` property is unexported and therefore unavailble from another packages.
```

Since the metaOpt is defined in the client package it has access to the clients unexported http property, unless I'm misunderstanding something. I also think we can use the http.Client's Timeout property rather than the transports since the transport can be shared among clients and we don't want to enforce identical timeouts (necessarily) across all clients. Something like this inside client.go:

```go
func Timeout (t time.Duration) metaOp {
        return func(c *Client) {
               c.http.Timeout = t
       }
}
```
Then you would call into the New function with:

```go
c, _ := client.New("address", "", true, client.Timeout(10*time.Second))
```

Does this seem like it would solve your problem?

mgmt/rest/client/client_func_test.go
iwankgb: None. I have not cleaned the code up after writing a test, apparently. Removed.

mgmt/rest/client/client.go
iwankgb: Constant created.

iwankgb:
I was originally thinking about using `metaOp` in our application code but adding such a function in `client` package would solve the issue indeed. Do you want to have such a function implemented and default timeout setting removed?

IRCody:
@iwankgb: Yes I think that's a little cleaner than the my original thought about just having the const. I also like the idea of using the http.Client timeout instead of the Transport since modifying the Transport would affect all clients.

iwankgb:
OK :)

IRCody:
LGTM @iwankgb. Could you squash to a single commit so it can be merged?

iwankgb:
Thanks a lot, @IRCody :)

PR #1509, opened 2017-02-07
SDI-1659: support for CORS header

### LINKED ISSUE #998: support for CORS headers
It doesn't look like snapd currently supports [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS) at least with v0.14.0-beta. This would be useful to allow javascript based tooling to interact with the snapd without needing to be on the same host or use a proxy layer.

Is this planned or are there any objections to adding CORS support?

candysmurf:
Fixes #998 

Summary of changes:
- Added CORS support.
- Cors origins have to be specified in the global config. Otherwise it's the same origin.

Testing done:
- swagger-ur which is in a different origin of Snap.

@intelsdi-x/snap-maintainers

### DIFF
--- a/mgmt/rest/server.go
+++ b/mgmt/rest/server.go
@@ -25,16 +25,27 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
+	"strconv"
 	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/julienschmidt/httprouter"
+	"github.com/rs/cors"
 	"github.com/urfave/negroni"
+
+	"strings"
 
 	"github.com/intelsdi-x/snap/mgmt/rest/api"
 	"github.com/intelsdi-x/snap/mgmt/rest/v1"
 	"github.com/intelsdi-x/snap/mgmt/rest/v2"
+)
+
+const (
+	allowedMethods = "GET, POST, DELETE, PUT, OPTIONS"
+	allowedHeaders = "Origin, X-Requested-With, Content-Type, Accept"
+	maxAge         = 3600
 )
 
 var (
@@ -45,18 +56,19 @@ var (
 )
 
 type Server struct {
-	apis       []api.API
-	n          *negroni.Negroni
-	r          *httprouter.Router
-	snapTLS    *snapTLS
-	auth       bool
-	pprof      bool
-	authpwd    string
-	addrString string
-	addr       net.Addr
-	wg         sync.WaitGroup
-	killChan   chan struct{}
-	err        chan error
+	apis           []api.API
+	n              *negroni.Negroni
+	r              *httprouter.Router
+	snapTLS        *snapTLS
+	auth           bool
+	pprof          bool
+	authpwd        string
+	addrString     string
+	addr           net.Addr
+	wg             sync.WaitGroup
+	killChan       chan struct{}
+	err            chan error
+	allowedOrigins map[string]bool
 	// the following instance variables are used to cleanly shutdown the server
 	serverListener net.Listener
 	closingChan    chan bool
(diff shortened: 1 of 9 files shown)


mgmt/rest/server.go
```diff
@@ -92,6 +93,15 @@ func New(cfg *Config) (*Server, error) {
 		negroni.HandlerFunc(s.authMiddleware),
 	)
 	s.r = httprouter.New()
+
+	// Deals with CORS
+	c := cors.New(cors.Options{
+		AllowedOrigins: []string{"*"},
```
andrzej-k: I don't think that it's safe to allow unrestricted access from any domain. Is there a possibility to constrain allowed origins - via snapteld configuration maybe?

mgmt/rest/server.go
```diff
@@ -92,6 +93,15 @@ func New(cfg *Config) (*Server, error) {
...
+		AllowedMethods: []string{"GET, POST, DELETE, PUT, PATCH, OPTIONS"},
```
marcintao: Not all of the HTTP methods were allowed previously. I believe it would be the safest to keep the list of allowed methods to a required minimum.
`PATCH` - I didn't find it declared for any of REST controllers. Are you sure we need to allow it? 
`OPTIONS` - the same as for `PATCH`. It's not needed for pre-flight requests. Negroni allows it implicitly only for the purpose of CORS. Furthermore, if it's needed for Snap API, it should be enabled with a flag `OptionsPassthrough`.

mgmt/rest/server.go
candysmurf: @andrzej-k, agreed your concerns. Should we allow localhost for now until we know whoch domains are allowed?

mgmt/rest/server.go
candysmurf: OPTIONS is needed for preflight. I can remove PATCH for now.

mgmt/rest/server.go
jcooklin: Configurable with 'localhost' as the default?

candysmurf: @jcooklin, sounds good. I'll add it as a global option.

mgmt/rest/server.go
marcintao: Yes, you're right - OPTIONS are needed for preflight. But isn't it handled before the actual request? I'm thinking about this: https://github.com/rs/cors/blob/master/cors.go#L189. I guess the actual request handling and method checking takes place later.

candysmurf: @marcintao, thanks for looking into it. However, Snap has two places to be patched. One place cannot use the lib. I added maxAge to ease your concerns. Please let me know any other suggestions you have.

candysmurf:
@andrzej-k, @marcintao, @jcooklin, CORS PR is update. Please review when you have a chance.

mgmt/rest/server_test.go
```diff
@@ -161,5 +163,45 @@ func TestRestAPIDefaultConfig(t *testing.T) {
...
+	cfg := GetDefaultConfig()
+
+	Convey("Test cors origin list", t, func() {
+		cfg.Corsd = "http://127.0.0.1:80, http://example.com"
+		s, o := NewMockServer(cfg)
+
+		Convey("Origins match results", func() {
+			So(s.allowedOrigins, ShouldResemble, map[string]bool{"http://127.0.0.1:80": true, "http://example.com": true})
+			So(len(o), ShouldEqual, 2)
+		})
```
andrzej-k: Consider adding more rainy day scenarios, ie. incorrect origin format, incorrect separator, etc.

mgmt/rest/server.go
```diff
@@ -259,6 +304,18 @@ func (s *Server) addRoutes() {
...
+func (s *Server) getAllowedOrigins(corsd string) []string {
+	if corsd == "" {
+		return []string{}
+	}
+	os := strings.Split(corsd, ",")
+	s.allowedOrigins = map[string]bool{}
+	for _, o := range os {
+		s.allowedOrigins[strings.TrimSpace(o)] = true
+	}
+	return os
```
andrzej-k: Consider adding more validation, for example if origin format is correct, something like: 
- httpa://domain.com
- 127.0.0.0.1
- http://?domain.com
- ...

should not pass, and user should be warned early that configuration is not valid.

mgmt/rest/server.go
```diff
@@ -60,6 +70,7 @@ type Server struct {
 	// the following instance variables are used to cleanly shutdown the server
 	serverListener net.Listener
 	closingChan    chan bool
+	allowedOrigins map[string]bool
```
candysmurf: not sure your comment here. It's used in authMiddleware method.

candysmurf:
@andrzej-k, thanks for your awesome review. I changed the name, added URL validation and more negative tests as you suggested. Please let me know if anything else.

mgmt/rest/server.go
andrzej-k: I'm just saying that "allowedOrigins" variable is added in a section which is described with a comment (3 lines above): 
```// the following instance variables are used to cleanly shutdown the server```

This may be misleading, so I just proposed to move "allowedOrigins" before that comment, i.e. 4 lines up.

candysmurf:
good catch! Changed. Thanks @andrzej-k!

jcooklin:
LGTM

PR #1510, opened 2017-02-07
Fix #1499 update explaination to auto-discovery plugin.

### LINKED ISSUE #1499: --auto-discover option needs better documentation
SNAPTELD documentation page lacks proper explanation of --auto-discover option. The option description is "Auto discover paths separated by colons", but doesn't say anything about the purpose.
There is a single example below with "snapteld --auto-discover /opt/snap/plugins/" line suggesting this option might be used to automatically load plugins, but that's all.

In the SNAPTELD_CONFIGURATION page at the end there is a comment on restarting snapteld process and it says that not only plugins in the auto_discover_path will be reloaded, but also tasks! (which wasn't mentioned anywhere else).

And there is the TASKS page which doesn't say anything about auto-discovery and auto-loading tasks at all.

I guess that auto-loading tasks feature was added later and documentation wasn't properly updated...
Anyway, I suggest adding an explanation of --auto-discover option on the SNAPTELD page with additional examples (with comments) showing clearly it migt be also used for auto-loading tasks.

Additionally, the TASKS page should also be updated with the proper information regarding auto-loading tasks.

nanliu:


nanliu:
This is related to: https://github.com/intelsdi-x/snap-packaging/pull/34. There are changes to the packaging repo config file, so we take advantage of new settings such as `max_plugin_restarts`. If the we decide to change the comments in the config file, I will update the packaging repo to match the updates here.

docs/SNAPTELD_CONFIGURATION.md
```diff
@@ -71,9 +71,9 @@ The control section contains settings for configuring the Control module within
 
 ```yaml
 control:
-  # auto_discover_path sets a directory to auto load plugins on the start
-  # of the Snap daemon
-  auto_discover_path: /opt/snap/plugins
+  # auto_discover_path sets the directory(s) to auto load plugins and tasks on
+  # the start of the snap daemon. This can be a comma separated list of directories.
```
candysmurf: giving an example of a comma separated list?

nanliu: The example below is two directories, one for plugins, one for tasks.

nanliu:
Since the packaging repo isn't public, I just want to check and make sure the follow changes are sensible defaults for snapteld.conf.

```
auto_discover_path: /opt/snap/plugins:/opt/snap/tasks
max_plugin_restarts: -1
```

I've also created a PR for [snap-docker](https://github.com/intelsdi-x/snap-docker/pull/25/files), but left out max_plugin_restarts because that is not compatible with older versions of snap, and will prevent the containers from starting if we ci older versions such as: SNAP_VERSION=0.19.0.

cc: @intelsdi-x/snap-maintainers

docs/SNAPTELD_CONFIGURATION.md
candysmurf: oh, it does not use command instead of ":"?

jcooklin:
I would be more conservative with `max_plugin_restarts`.  If there is a problem with the system or a plugin I don't think we should continually restart the plugin by default.

candysmurf:
LGTM

nanliu:
I'll switch max_plugin_restarts back to the default for the package to keep everything consistent.

Issue #1502, opened 2017-01-31
Do not execute binaries directly from the auto-load path


jcooklin:
Currently, plugins loaded by the snaptel cli via 'snaptel plugin load file_name' are copied to a temporary directory before they are executed. Plugins placed in the auto-load directory are loaded directly from that filepath without going through this process. This prevents plugin files from getting updated when Snap is running since they are in use. We would like snap plugins to behave the same whether they are loaded via the autoloader or snaptel plugin load cli.

jtlisi:
#1494 is something that relates very closely to this. If the temporary directory could default to `/opt/snap/plugins/tmp/` or something along those lines that would be better than the linux default tmp directory

jcooklin:
@jtlisi:  agreed.  @rashmigottipati is currently working on #1491.

PR #1512, opened 2017-02-08
(SDI-2475) Configure the TempDir where the API loads plugins.

### LINKED ISSUE #1494: Configure the TempDir where the API loads plugins.
One issue we are facing with loading plugins through the api is that the temporary directory which houses the temporary files is not configurable in snapteld's configuration. 

I noticed on line 192 of [mgmt/rest/v1/plugin.go](https://github.com/intelsdi-x/snap/blob/8080135b12e250cdba74937d679fa767d9c3e5ac/mgmt/rest/v1/plugin.go#L192)
the function `writeFile` uses the the ioutil.TempDir function: 
```golang
func writeFile(filename string, b []byte) (string, error) {
	// Create temporary directory
	dir, err := ioutil.TempDir("", "")
        ...
```

ioutil.TempDir defaults to the TMPDIR env variable or to `/tmp` directory if it is not set. I was hoping it could default to either a directory in `/opt/snap/` or be configurable in the snapteld configuration.

Let me know what you guys think.

Thanks!

rashmigottipati:
Fixes #1494

Summary of changes:
- Added a configurable flag in snapteld for loading plugins

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/flags.go
+++ b/control/flags.go
@@ -70,5 +70,11 @@ var (
 		EnvVar: "SNAP_CONTROL_LISTEN_ADDR",
 	}
 
-	Flags = []cli.Flag{flNumberOfPLs, flPluginLoadTimeout, flAutoDiscover, flPluginTrust, flKeyringPaths, flCache, flControlRpcPort, flControlRpcAddr}
+	flTempDirPath = cli.StringFlag{
+		Name:   "temp_dir_path",
+		Usage:  fmt.Sprintf("Temporary path for loading plugins (default: %v)", defaultTempDirPath),
+		EnvVar: "SNAP_TEMP_DIR_PATH",
+	}
+
+	Flags = []cli.Flag{flNumberOfPLs, flPluginLoadTimeout, flAutoDiscover, flPluginTrust, flKeyringPaths, flCache, flControlRpcPort, flControlRpcAddr, flTempDirPath}
 )
--- a/docs/SNAPTELD.md
+++ b/docs/SNAPTELD.md
@@ -43,6 +43,7 @@ $ snapteld [global options] command [command options] [arguments...]
 --cache-expiration value                     The time limit for which a metric cache entry is valid (default: 500ms) [$SNAP_CACHE_EXPIRATION]
 --control-listen-port value                  Listen port for control RPC server (default: 8082) [$SNAP_CONTROL_LISTEN_PORT]
 --control-listen-addr value                  Listen address for control RPC server [$SNAP_CONTROL_LISTEN_ADDR]
+--temp_dir_path value                        Temporary path for loading plugins [$SNAP_TEMP_DIR_PATH]
 --work-manager-queue-size value              Size of the work manager queue (default: 25) [$WORK_MANAGER_QUEUE_SIZE]
 --work-manager-pool-size value               Size of the work manager pool (default: 4) [$WORK_MANAGER_POOL_SIZE]
 --disable-api, -d                            Disable the agent REST API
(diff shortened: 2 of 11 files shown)


docs/SNAPTELD.md
```diff
@@ -52,6 +52,7 @@ $ snapteld [global options] command [command options] [arguments...]
 --rest-cert value                            A path to a certificate to use for HTTPS deployment of Snap's REST API
 --rest-key value                             A path to a key file to use for HTTPS deployment of Snap's REST API
 --rest-auth                                  Enables Snap's REST API authentication
+--rest-load-path                             A configurable path from which the API loads plugins
```
jcooklin: We should consider exposing this as top level (control) config.  It could be called `--temp-dir-path`.  This will line up when we resolve #1502.

mgmt/rest/common/common.go
```diff
@@ -0,0 +1,39 @@
...
+	"path"
+	"runtime"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/intelsdi-x/snap/control"
+)
+
+func WriteFile(filename string, b []byte) (string, error) {
+	// Create temporary directory
+	dir, err := ioutil.TempDir(control.GetDefaultConfig().TempDirPath, "snap-plugin-")
```
IRCody: Why the trailing `-`?

rashmigottipati: The trailing isn't required. Each time a plugin loads, a tmp file is being generated with a random number. snap-plugin- would be the prefix with which the file is generated. It might as well just be snap-plugin.

jcooklin: It's a prefix.  The method will add a unique identifier.

control/flags.go
```diff
@@ -70,5 +70,11 @@ var (
 		EnvVar: "SNAP_CONTROL_LISTEN_ADDR",
 	}
 
-	Flags = []cli.Flag{flNumberOfPLs, flPluginLoadTimeout, flAutoDiscover, flPluginTrust, flKeyringPaths, flCache, flControlRpcPort, flControlRpcAddr}
+	flTempDirPath = cli.StringFlag{
+		Name:   "temp_dir_path",
+		Usage:  "Temporary path for loading plugins",
+		EnvVar: "TEMP_DIR_PATH",
```
jcooklin: Add `SNAP_` to the EnvVar value.

docs/SNAPTELD.md
```diff
@@ -43,6 +43,7 @@ $ snapteld [global options] command [command options] [arguments...]
 --cache-expiration value                     The time limit for which a metric cache entry is valid (default: 500ms) [$SNAP_CACHE_EXPIRATION]
 --control-listen-port value                  Listen port for control RPC server (default: 8082) [$SNAP_CONTROL_LISTEN_PORT]
 --control-listen-addr value                  Listen address for control RPC server [$SNAP_CONTROL_LISTEN_ADDR]
+--temp_dir_path value                        Temporary path for loading plugins [$TEMP_DIR_PATH]
```
jcooklin: Add SNAP_ to $TEMP_DIR_PATH

mgmt/rest/common/common.go
IRCody: Ok so the path for the plugin that's written will be something like:
`<TempDirPath>/snap-plugin-<unique identifier>/snap-plugin-<plugin-name>`? And the reason for separate directories per plugin is that multiple plugins might have the same binary name?

jcooklin:
LGTM

PR #1534, opened 2017-02-23
(SDI-2483) Auto-Load plugins via temp similar to cli

### LINKED ISSUE #1502: Do not execute binaries directly from the auto-load path

rashmigottipati:
Fixes #1502 

Summary of changes:
- Load the plugins via auto-loader using the temp path similar to running snaptel plugin load cli
- Removed the autoload field

Testing done:
- Manual
- Legacy tests

@intelsdi-x/snap-maintainers


core/plugin.go
```diff
@@ -95,21 +99,56 @@ type SubscribedPlugin interface {
...
-	}
-	err := rp.generateCheckSum()
-	if err != nil {
-		return nil, err
+	path      string
+	checkSum  [sha256.Size]byte
+	signature []byte
+}
+
+func NewRequestedPlugin(path, tmp string, b []byte) (*RequestedPlugin, error) {
```
IRCody: Can we add a comment about the expected behavior of this? Especially with b being nil vs not being nil.

pkg/fileutils/file.go
```diff
@@ -0,0 +1,41 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// WriteFile creates a temporary directory for loading plugins.
+// Plugins loaded by the snaptel cli and plugins placed in the auto-load directory go through this route of copying the plugin binaries to the temp dir and executing from temp.
+// This function takes the name of the original file, path to the original file and the content of the file, and copies it to a a temp file after creating one.
```
jcooklin: Include a fragment that states what it returns specifically.

core/plugin.go
```diff
@@ -95,21 +99,60 @@ type SubscribedPlugin interface {
...
-	}
-	err := rp.generateCheckSum()
-	if err != nil {
-		return nil, err
+	path      string
+	checkSum  [sha256.Size]byte
+	signature []byte
+}
+
+// NewRequestedPlugin returns a Requested Plugin which represents the plugin path, check sum and signature
```
jcooklin: Provide details about the expected arguments (what are path, tmp, b? can they be empty or nil, etc)

core/plugin.go
```diff
@@ -95,21 +99,60 @@ type SubscribedPlugin interface {
...
+// Both REST API and snaptel cli would call this before loading a plugin.
+func NewRequestedPlugin(path, tmp string, b []byte) (*RequestedPlugin, error) {
```
jcooklin: name or fileName would be more meaningful than tmp

jcooklin:
LGTM

PR #1538, opened 2017-02-27
(SDI-2482) Support environment variables for snaptel CLI options

### LINKED ISSUE #1498: snaptel configuration file
Per discussion with @jtlisi, the snaptel client could use a default configuration file to simplify cli usage. Our current snaptel cli does not support ENV var for most options, and the `-c config` option only supports rest-auth-pwd.

We should support all cli options within the config file, or support more environment variable for CLI options to simplify cli usage.

rashmigottipati:
Fixes #1498 

Summary of changes:
- Added support for environment variables for the snaptel CLI options

@intelsdi-x/snap-maintainers

### DIFF
--- a/cmd/snaptel/config.go
+++ b/cmd/snaptel/config.go
@@ -28,6 +28,7 @@ import (
 	"strconv"
 	"text/tabwriter"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/intelsdi-x/snap/core/ctypes"
 	"github.com/urfave/cli"
 )
@@ -40,6 +41,11 @@ type restAPIConfig struct {
 }
 
 func (c *config) loadConfig(path string) error {
+	log.WithFields(log.Fields{
+		"_module":     "snaptel-config",
+		"_block":      "loadConfig",
+		"config_path": path,
+	}).Warning("The snaptel configuration file will be deprecated. Find more information here: https://github.com/intelsdi-x/snap/issues/1539")
 	b, err := ioutil.ReadFile(path)
 	if err != nil {
 		return fmt.Errorf("Unable to read config. File might not exist.")
--- a/cmd/snaptel/flags.go
+++ b/cmd/snaptel/flags.go
@@ -35,21 +35,24 @@ var (
 		Value:  "http://localhost:8181",
 	}
 	flAPIVer = cli.StringFlag{
-		Name:  "api-version, a",
-		Usage: "The Snap API version",
-		Value: "v1",
+		Name:   "api-version, a",
+		Usage:  "The Snap API version",
+		EnvVar: "SNAP_API_VERSION",
+		Value:  "v1",
 	}
 	flSecure = cli.BoolFlag{
-		Name:  "insecure",
-		Usage: "Ignore certificate errors when Snap's API is running HTTPS",
+		Name:   "insecure",
+		Usage:  "Ignore certificate errors when Snap's API is running HTTPS",
+		EnvVar: "SNAP_INSECURE",
 	}
 	flRunning = cli.BoolFlag{
 		Name:  "running",
 		Usage: "Shows running plugins",
 	}
 	flPassword = cli.BoolFlag{
-		Name:  "password, p",
-		Usage: "Password for REST API authentication",
+		Name:   "password, p",
+		Usage:  "Require password for REST API authentication",
+		EnvVar: "SNAP_REST_PASSWORD",
 	}
 	flConfig = cli.StringFlag{
 		Name:   "config, c",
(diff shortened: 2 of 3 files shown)


nanliu:
Since we are updating this file, we might as well change snap plugin version option to use SNAP_PLUGIN_VERSION, instead of SNAP_VERSION. It seems really confusing to use SNAP_VERSION environment variable, and I'm surprised it doesn't cause problems already.

rashmigottipati:
@nanliu I agree. Fixed that.

nanliu:
Could you add a deprecation warning at the beginning of this function:
https://github.com/intelsdi-x/snap/blob/master/cmd/snaptel/config.go#L42

cmd/snaptel/config.go
```diff
@@ -40,6 +41,7 @@ type restAPIConfig struct {
 }
 
 func (c *config) loadConfig(path string) error {
+	log.Warning("The snaptel configuration file will be deprecated. Find more information here: https://intelsdi.atlassian.net/browse/SDI-2580 ")
```
nanliu: Let's move this to a github issue link since not everyone have access to Alassian. Otherwise this PR LGTM.

IzabellaRaulin: If you change this log message, what do you think about something similar to others existing (e.g. https://github.com/intelsdi-x/snap/blob/master/control/available_plugin.go#L103). So it might look like below:
```
log.WithFields(log.Fields{
   "_module":     "snaptel-config",
   "_block":      "loadConfig",
   "config_path": path, 
}).Warning("The snaptel configuration file will be deprecated. Find more information here: <GH issue>")
```
<sup>*</sup>where GH issue is an announcement of depreciation loading config by snaptel - is there any already open?

nanliu:
Let's people a chance to comment on #1539 if there are any concerns, and we can move forward with this change.

cmd/snaptel/config.go
IzabellaRaulin: Now, this log looks good to me.

cmd/snaptel/flags.go
```diff
@@ -77,8 +80,9 @@ var (
 		Usage: "The plugin name",
 	}
 	flPluginVersion = cli.IntFlag{
-		Name:  "plugin-version, v",
-		Usage: "The plugin version",
+		Name:   "plugin-version, v",
+		Usage:  "The plugin version",
+		EnvVar: "SNAP_PLUGIN_VERSION",
```
IzabellaRaulin: This flag is used only in the following snaptel plugin commands:
- ` snaptel plugin swap <load_plugin_path> <unload_plugin_type>:<unload_plugin_name>:<unload_plugin_version>`
- ` snaptel plugin config get  <plugin_type>:<plugin_name>:<plugin_version>`
So it is used only to swapping plugins or get info about configuration requested by the plugin.

Based on those, could you tell me what is the use-case to declare the env var "SNAP_PLUGIN_VERSION"?

cmd/snaptel/flags.go
```diff
@@ -35,21 +35,24 @@ var (
...
 	flRunning = cli.BoolFlag{
 		Name:  "running",
 		Usage: "Shows running plugins",
 	}
 	flPassword = cli.BoolFlag{
-		Name:  "password, p",
-		Usage: "Password for REST API authentication",
+		Name:   "password, p",
+		Usage:  "Password for REST API authentication",
+		EnvVar: "SNAP_PASSWORD",
```
IzabellaRaulin: @rashmigottipati - do you have the same feeling like me that it might be confusing? I know that this is not related to this PR, but maybe simple changing the usage description to 'Require password for REST API authentication'  or sth like that will be more clear for users that this is a boolean type.  Also, the name of env variable "SNAP_PASSWORD" might suggest that some string (password) should be set there.  Maybe renaming that to "SNAP_REQUEST_PASSWORD" will help to avoid such mistakes. 

This is the only light suggestion. If you agree - that is great, if not - that is also ok for me.

cmd/snaptel/flags.go
```diff
@@ -35,21 +35,24 @@ var (
...
 	flRunning = cli.BoolFlag{
 		Name:  "running",
 		Usage: "Shows running plugins",
 	}
 	flPassword = cli.BoolFlag{
-		Name:  "password, p",
-		Usage: "Password for REST API authentication",
+		Name:   "password, p",
+		Usage:  "Require password for REST API authentication",
+		EnvVar: "SNAP_REQUIRE_PASSWORD",
```
nanliu: Would SNAP_API_PASSWORD, or SNAP_REST_PASSWORD be a better description?

Issue #985, opened 2016-06-10
RFC - Single Run Task


IRCody:
## Definition

A Single Run Task is a task where the workflow is going to be scheduled for execution immediately (at most once). The task is saved so that it may be run again easily simply by starting it again.
## Purpose

A single run task serves the following purposes:
- Allows a user to test a task end to end easily with 1 run through the workflow to ensure a task will work.
- Forms a sort of 'task library' for infrequently run tasks. 
## Desired User experience
- Specify a task as single run, Via task.json posted to rest 
- task specified as a single run can be run again easily by doing snapctl task start <task id>.
- not necessarily related – On the command line via snapctl by overwriting schedule type. 
## Implementation notes

Implement new, single run, scheduler type that runs a workflow once.
## Questions
- Should a request to run a single task block in rest to wait for exit code to send back as the response?
  Would be potentially useful for users who want to manage scheduler of a task directly without polling snap for task completion.

lynxbat:
Wouldn't a schedule time that has a `start time` and `count` satisfy this need in the existing framework?
The UX for the CLI could have a friendly switch for single-run (set count==1).

The scheduler would execute the Task when the start time is current or in the past until the run count meets the count. If it is one it would execute once.

I am opposed to schedule type on the CLI unless the UX is considered. Manifest structure shouldn't impose on CLI UX.

IRCody:
>Wouldn't a schedule time that has a start time and count satisfy this need in the existing framework?
The UX for the CLI could have a friendly switch for single-run (set count==1).

>The scheduler would execute the Task when the start time is current or in the past until the run count meets the count. If it is one it would execute once.

You meant schedule type not time? If so I think this makes sense. Seems simple enough implementation wise -- run n times vs run once.

>I am opposed to schedule type on the CLI unless the UX is considered. Manifest structure shouldn't impose on CLI UX.

The RFC wasn't attempting to take a position on whether the schedule type changing is explicit (via --schedule-type=x) or implicit (i.e. via something like `snapctl task create --run-count=1 -t task.json`). 

Is your position that we should only allow single-run to be defined in the task manifest? I think it could be useful to run an existing task just once without having to modify the manifest directly but don't feel super strongly about it.

jackson-intel:
We're working on deploying some Intel collectors with SNAP and would like a feature like this one.

IzabellaRaulin:
[...]

PR #1544, opened 2017-03-06
Schedule workflow execution immediately

IzabellaRaulin:
Related to https://github.com/intelsdi-x/snap/issues/985
This PR is a part of the work being planned related to a single run task.

### Before:
- waiting a period of time at the beginning of every spin() 

### After:
- for the first spin(), do not wait on interval and schedule workflow execution immediately

### Summary of changes:
- the workflow is going to be scheduled for execution immediately 
    - reset the lastFireTime at each `task.Spin()`(see [scheduler/task.go#255](https://github.com/intelsdi-x/snap/pull/1544/files#diff-77f163e381dd113e90ebd1bd76479b54R255)), so for the first run lastFireTime equals time.Time{}
    - do now wait on interval for the first run (see [pkg/schedule/schedule.go#L53](https://github.com/intelsdi-x/snap/pull/1544/files#diff-d856bf186a791e0d54f236ef9fcdb6e0R53))
- some typos fixed

 
Testing done:
- manual tests
- medium tests

@intelsdi-x/snap-maintainers

### DIFF
--- a/pkg/schedule/windowed_schedule_test.go
+++ b/pkg/schedule/windowed_schedule_test.go
@@ -14,11 +14,57 @@ import (
 func TestWindowedSchedule(t *testing.T) {
 	log.SetLevel(log.DebugLevel)
 	Convey("Windowed Schedule", t, func() {
+		Convey("nominal window without misses", func() {
+			startWait := time.Millisecond * 50
+			windowSize := time.Millisecond * 200
+			interval := time.Millisecond * 10
+
+			start := time.Now().Add(startWait)
+			stop := time.Now().Add(startWait + windowSize)
+			w := NewWindowedSchedule(
+				interval,
+				&start,
+				&stop,
+			)
+
+			err := w.Validate()
+			So(err, ShouldBeNil)
+
+			var r []Response
+			last := *new(time.Time)
+
+			state := Active
+			before := time.Now()
+			for state == Active {
+				r1 := w.Wait(last)
+				state = r1.State()
+				last = time.Now()
+				r = append(r, r1)
+			}
+
+			// there are 0 missed responses, so for this schedule
+			// we expect to get between 19 - 22 responses
+			So(len(r), ShouldBeBetweenOrEqual, 19, 22)
+
+			var missed uint
+			for _, x := range r {
+				missed += x.Missed()
+			}
+			So(missed, ShouldEqual, 0)
+
+			// the task is expected to fire immediately on determined start-time
+			So(
+				r[0].LastTime().Sub(before).Seconds(),
+				ShouldBeBetweenOrEqual,
+				(startWait).Seconds(),
+				(startWait + interval).Seconds(),
+			)
+		})
+
 		Convey("nominal window with a few misses", func() {
 			startWait := time.Millisecond * 50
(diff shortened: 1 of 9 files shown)


pkg/schedule/windowed_schedule_test.go
```diff
@@ -14,11 +14,55 @@ import (
...
+
+			state := Active
+			before := time.Now()
+			for state == Active {
+				r1 := w.Wait(last)
+				state = r1.State()
+				last = time.Now()
+				r = append(r, r1)
+			}
+			// we should have either 21 or 23 minus 0 missed
```
jcooklin: "We should have either 21 or 23?", is an odd/inaccurate comment given we are checking for a value between 20 and 22.  It should probably read, "we expect a range between x and y".

Also, this test is flaky >50% of the time on my system.

img

pkg/schedule/windowed_schedule_test.go
```diff
@@ -50,26 +94,28 @@ func TestWindowedSchedule(t *testing.T) {
 					time.Sleep(w.Interval * 2)
 				}
 			}
-			// we should have either 16 or 17 minus 3 missed
-			So(len(r), ShouldBeBetweenOrEqual, 15, 17)
+			// we should have either 17 or 19 minus 4 missed
```
jcooklin: The comment should read, "We should have results between x and y".

This test comes back with 15 around 10% of the time for me.

img

pkg/schedule/windowed_schedule_test.go
```diff
@@ -100,14 +147,23 @@ func TestWindowedSchedule(t *testing.T) {
 					time.Sleep(w.Interval * 2)
 				}
 			}
-			// we should have either 16 or 17 minus 3 missed
-			So(len(r), ShouldBeBetweenOrEqual, 10, 12)
+			// we should have either 12 or 14 results minus 4 missed
```
jcooklin: "Should have between 12 and 14 results"

pkg/schedule/windowed_schedule_test.go
IzabellaRaulin: Yes, I see. I changed the comment through all tests to be more appropriate. Now for each of tests cases, window-width is 200ms, so for interval 10ms it is expected to get in a range 19 - 22 responses (or 15-18 for tests when there are some missed responses). 
@jcooklin, could you run those tests on your system to confirm that?

pkg/schedule/windowed_schedule_test.go
IzabellaRaulin: windowed width has been changed to 200ms to be more corresponding to the other tests-cases.

pkg/schedule/windowed_schedule_test.go
IzabellaRaulin: For the test with 4 missed responses,  the range has been changed to 15 -18.
Notice that for a test without missed responses, 19- 22 responses are expected to receive.

pkg/schedule/windowed_schedule_test.go
jcooklin: 100/100 runs were successful!

PR #1548, opened 2017-03-15
Run task `count` times (available for simple and window schedule)

IzabellaRaulin:
Related to #985

A workflow is going to be scheduled for execution immediately (has been already delivered in PR 1544)
A workflow is going to be scheduled until the run count meets the „count”, what means that a user starts a task with a count of 100 and interval of 1s, the expected behavior is that the task should run 100s.
A task is going to state _Ended_  after the run count meets the „count”. It might be successfully restarted.
Default value of the "count" is zero what means no limit

### TO DO:
- [x] small and medium tests for windowed schedule
- [x] added `count` to exemplary task manifest in [snap/examples/task](https://github.com/IzabellaRaulin/snap/tree/run_task_x_times/examples/tasks)
- [x] tests for others packages

### Summary of changes:
- run task `count` times - the field of schedule section, possible for simple and window schedule
- the code of simple schedule has been removed (simple schedule is a window schedule without start and stop time) - this change does NOT impact on existing task

### Examples:
1. An example of defining **_single run task_** in task manifest:
a) simple schedule, start a task immediately
  ```json
  "schedule": {
    "type": "simple",
    "interval": "1s",
    "count": 1
  },
  ```

b) windowed schedule, start a task on defined _start_timestamp_
  ```json
  "schedule": {
    "type": "windowed",
    "interval": "1s"
    "start_timestamp": "2017-03-10T00:58:37+01:00",
    "count": 1
  },
  ```

c) try to run windowed schedule with determined both _stop_timestamp_ and _count
  The task will run as normal, however, the value of _count_ will be ignored. The following warning will be informed about that:
  `WARN[2017-03-10T00:58:51+01:00] The window stop timestamp and the count cannot be specified simultaneously. The parameter count has been ignored.  _block=NewWindowedSchedule _module=schedule`

2. An example of setting count as a param passing by CLI:

  `snaptel task create -t <my-manifest> --count 1`
  
  The value of `count` in task manifest will be overwritten.

Testing done:
- manual tests
- small, medium tests


cmd/snaptel/task.go
```diff
@@ -115,6 +115,20 @@ func stringValToInt(val string) (int, error) {
...
+	valInt, err := stringValToInt(val)
+	if err != nil {
+		return uint(0), err
+	}
+	if valInt < 0 {
+		return uint(0), fmt.Errorf("Value '%v' is a negative number and cannot be parsed as an unsigned integer", val)
+	}
+	// return the unsigned integer equivalent of the input string and a nil error (indicating success)
+	return uint(valInt), err
+}
```
andrzej-k: Not sure but seems to me that max here is MAX_UINT/2, right? Meaning that when we advertise count as uint it may not take values bigger than MAX_UINT/2

docs/SNAPTEL.md
```diff
@@ -186,16 +187,18 @@ and then:
 3. load a publishing plugin
 4. list the plugins
 5. start a task with a task manifest
-6. start a task with a workflow manifest
-7. list the tasks
-8. unload the plugins
+6. make a single run task
```
andrzej-k: Hmm, it seems that no special task is created for single run task - how about saying "start a single run task with a task manifest" ?

docs/TASKS.md
```diff
@@ -48,46 +48,104 @@ The header contains a version, used to differentiate between versions of the tas
...
+	"version": 1,
+	"schedule": {
+		"type": "simple",
+		"interval": "1s",
+		"count": 1
+	},
+	"max-failures": 1,
+  ```
+Set `count` to 1 to define _single run task_.   
+You can expect that your task will finish in time equals to _count_ multiplied by _interval_.
```
andrzej-k: Should we indicate what is min and max for count?

pkg/schedule/windowed_schedule.go
```diff
@@ -49,12 +85,24 @@ func (w *WindowedSchedule) Validate() error {
 	if w.Interval <= 0 {
 		return ErrInvalidInterval
 	}
+
+	// the schedule passed validation, set is as an active
```
andrzej-k: "set is as an active" -> "set as active"

pkg/schedule/windowed_schedule.go
```diff
@@ -74,32 +122,32 @@ func (w *WindowedSchedule) Wait(last time.Time) Response {
...
 		} else {
 			w.state = Ended
 			m = 0
 		}
 	} else {
-		logger.WithFields(log.Fields{
-			"_block":   "windowed-wait",
-			"last":     last,
-			"interval": w.Interval,
-		}).Debug("waiting for interval")
```
andrzej-k: Why this debug log was removed?

IzabellaRaulin: Removed by mistake, this log should be still there. Thanks

pkg/schedule/windowed_schedule.go
IzabellaRaulin: done

pkg/schedule/windowed_schedule.go
```diff
@@ -12,35 +12,71 @@ var (
...
+	Count      uint
+	state      ScheduleState
+	stopOnTime *time.Time
 }
 
 // NewWindowedSchedule returns an instance of WindowedSchedule given duration,
 // start and stop time
-func NewWindowedSchedule(i time.Duration, start *time.Time, stop *time.Time) *WindowedSchedule {
+func NewWindowedSchedule(i time.Duration, start *time.Time, stop *time.Time, count uint) *WindowedSchedule {
+	// if stop and count were both defined, ignore the `count`
```
IzabellaRaulin: Added: 
```
// NewWindowedSchedule returns an instance of WindowedSchedule with given interval, start and stop timestamp
// and count of expected runs. The value of `count` determines stop time, so specifying it together with `stop`
// is not allowed and the count will be set to defaults 0 in such cases.
```

docs/SNAPTEL.md
IzabellaRaulin: done

cmd/snaptel/task.go
IzabellaRaulin: I changed that to using `strconv.ParseUint` (this package is already imported) instead of calling function `stringValToInt` and converting its output from int to uint. Now the biggest count value might equal MAX_UINT.

docs/TASKS.md
```diff
@@ -48,46 +48,141 @@ The header contains a version, used to differentiate between versions of the tas
...
+   ```
+  
+    
+    
 #### Max-Failures
 
 By default, Snap will disable a task if there are 10 consecutive errors from any plugins within the workflow.  The configuration
 can be changed by specifying the number of failures value in the task header.  If the `max-failures` value is -1, Snap will
-not disable a task with consecutive failure.  Instead, Snap will sleep for 1 second for every 10 consecutive failures
+not disable a task with consecutive failure. Instead, Snap will sleep for 1 second for every 10 consecutive failures
```
IzabellaRaulin: Here just double space has been removed

mgmt/rest/rest_v1_test.go
```diff
@@ -858,9 +858,10 @@ func TestV1Task(t *testing.T) {
 			// GetTasks returns an unordered map,
 			// thus there is more than one possible response
 			So(
+				string(body),
+				ShouldBeIn,
 				responses,
-				ShouldContain,
-				string(body))
+			)
```
IzabellaRaulin: In this test file there are only changes related to keeping the pattern that an actual result should be at the first position:
```
So(actual-result, assertion, expected-result)
```

mgmt/rest/rest_v2_test.go
```diff
@@ -288,9 +288,10 @@ func TestV2Task(t *testing.T) {
 			body, err := ioutil.ReadAll(resp.Body)
 			So(err, ShouldBeNil)
 			So(
-				fmt.Sprintf(mock.ADD_TASK_RESPONSE, r.port),
+				string(body),
 				ShouldResemble,
-				string(body))
+				fmt.Sprintf(mock.ADD_TASK_RESPONSE, r.port),
+			)
```
IzabellaRaulin: In this test file there are only changes related to keeping the pattern that an actual result should be at the first position:
```
So(actual-result, assertion, expected-result)
```

mgmt/rest/v1/fixtures/mock_task_manager.go
```diff
@@ -249,7 +250,7 @@ const (
       }
     },
     "schedule": {
-      "type": "simple",
+      "type": "windowed",
```
IzabellaRaulin: After including "simple" into "windowed" schedule, the getting task response will contain "windowed" as schedule type

mgmt/rest/v1/rbody/task.go
```diff
@@ -217,12 +217,6 @@ func (s *ScheduledTaskEnabled) ResponseBodyType() string {
 
 func assertSchedule(s schedule.Schedule, t *AddScheduledTask) {
 	switch v := s.(type) {
-	case *schedule.SimpleSchedule:
-		t.Schedule = &core.Schedule{
-			Type:     "simple",
-			Interval: v.Interval.String(),
-		}
-		return
```
IzabellaRaulin: `schedule.SimpleSchedule` has no longer available, so it has to be removed from here.

mgmt/tribe/worker/worker.go
```diff
@@ -545,14 +545,49 @@ func (w worker) isPluginLoaded(n, t string, v int) bool {
 }
 
 func getSchedule(s *core.Schedule) schedule.Schedule {
```
IzabellaRaulin: Before:
In Tribe only Simple schedule was available

After:
- added to switch case others available schedule types
- processed here schedule validation to return only valid schedule
- added case `defaults` and log if schedule type is unrecognized

IzabellaRaulin: Still "nil" is returned if something was going wrong, so tribe-logs will point to invalid nil schedule as it works currently.

pkg/schedule/simple_schedule.go
```diff
@@ -1,65 +0,0 @@
-package schedule
```
IzabellaRaulin: Removed package with simple schedule which equals to windowed schedule without determined start and stop timestamp, so only windowed type has been kept. Notice that this only impact how the code looks like. For users, removing that will be **unvisible** - task with simple schedule will work as so far.

pkg/schedule/simple_schedule_test.go
```diff
@@ -1,113 +0,0 @@
...
-
-package schedule
-
-import (
-	"testing"
-	"time"
-
-	. "github.com/smartystreets/goconvey/convey"
-)
-
```
IzabellaRaulin: **To reviewers**:
All test cases from this file have been moved to small and medium tests for windowed schedule
You can find:
 - "test Wait()" nad "invalid interval" in [windowed_schedule_small_test.go](https://github.com/intelsdi-x/snap/pull/1548/files#diff-4c876ad90b686a38e10efef6096ba191R70)
-  tests corresponding to "Simple schedule with no misses" and "with a few misses  you can find [windowed_schedule_medium_test.go](https://github.com/intelsdi-x/snap/pull/1548/files#diff-1b8090dae9b2079edff17c74a4ba55acR36)

pkg/schedule/windowed_schedule.go
```diff
@@ -49,12 +105,24 @@ func (w *WindowedSchedule) Validate() error {
 	if w.Interval <= 0 {
 		return ErrInvalidInterval
 	}
+
+	// the schedule passed validation, set as active
+	w.state = Active
```
IzabellaRaulin: **To reviewers**:
Setting schedule state as active is very important in a scenario when the task is restarted, so validation happens once again to check if the schedule is still valid at this point in time. This is the best place to set schedule's state to active if validation is passed. The only active schedule might be fired again.

scheduler/scheduler.go
```diff
@@ -500,35 +500,9 @@ func (s *scheduler) startTask(id, source string) []serror.SnapError {
 		return errs
 	}
 
-	// Group dependencies by the node they live on
```
IzabellaRaulin: **To reviewers**:
This code from in lines 504-531 occurred in few places, so I decided to put that in [(t *task) SubscribePlugins()](https://github.com/intelsdi-x/snap/pull/1548/files#diff-77f163e381dd113e90ebd1bd76479b54R364)

scheduler/scheduler.go
```diff
@@ -813,13 +775,7 @@ func (s *scheduler) HandleGomitEvent(e gomit.Event) {
 		}).Debug("event received")
 		// We need to unsubscribe from deps when a task goes disabled
 		task, _ := s.getTask(v.TaskID)
-		depGroups := getWorkflowPlugins(task.workflow.processNodes, task.workflow.publishNodes, task.workflow.metrics)
-		for k := range depGroups {
-			mgr, err := task.RemoteManagers.Get(k)
-			if err == nil {
-				mgr.UnsubscribeDeps(task.ID())
-			}
-		}
```
IzabellaRaulin: **To reviewers**:
This code from in lines 816-822 occurred in few places, so I decided to put that in [(t *task) UnsubscribePlugins()](https://github.com/intelsdi-x/snap/pull/1548/files#diff-77f163e381dd113e90ebd1bd76479b54R364)

scheduler/task.go
```diff
@@ -254,7 +255,7 @@ func (t *task) Spin() {
 	// misses for the interval while stopped.
 	t.lastFireTime = time.Time{}
 
-	if t.state == core.TaskStopped {
+	if t.state == core.TaskStopped || t.state == core.TaskEnded {
```
IzabellaRaulin: **To reviewers**:
Stopped and Ended task might be spin again (resumable).

pkg/schedule/windowed_schedule.go
```diff
@@ -12,35 +31,72 @@ var (
...
+// and count of expected runs. The value of `count` determines stop time, so specifying it together with `stop`
+// is not allowed and the count will be set to defaults 0 in such cases.
+func NewWindowedSchedule(i time.Duration, start *time.Time, stop *time.Time, count uint) *WindowedSchedule {
+	// if stop and count were both defined, ignore the `count`
+	if count != 0 && stop != nil {
+		count = 0
+		// log about ignoring the `count`
+		logger.WithFields(log.Fields{
+			"_block": "NewWindowedSchedule",
+		}).Warning("Specifying both the window stop and the count is not allowed. The parameter `count` has been ignored.")
```
andrzej-k: allowed -> supported ?

pkg/schedule/windowed_schedule.go
```diff
@@ -12,35 +31,72 @@ var (
...
+		var newStop time.Time
+
+		// if start is not set or points in the past,
+		// use the current time to calculate stopOnTime
+		if w.StartTime != nil && time.Now().Before(*w.StartTime) {
+			newStop = w.StartTime.Add(time.Duration(w.Count) * w.Interval)
+		} else {
+			// set a new stop timestamp from this point in time
+			newStop = time.Now().Add(time.Duration(w.Count) * w.Interval)
+		}
```
andrzej-k: Are there small/medium tests for this logic?

cmd/snaptel/task.go
```diff
@@ -234,6 +248,13 @@ func (t *task) setScheduleFromCliOptions(ctx *cli.Context) error {
 	if !ctx.IsSet("interval") && interval == "" && t.Schedule.Interval == "" {
 		return fmt.Errorf("Usage error (missing interval value); when constructing a new task schedule an interval must be provided")
 	}
+
+	countValStr := ctx.String("count")
+	if ctx.IsSet("count") || countValStr != "" {
+		if count, err := stringValToUint(countValStr); err == nil {
+			t.Schedule.Count = count
+		}
```
IzabellaRaulin: Good point. I added returning the following error: `fmt.Errorf("Usage error (bad count format); %v", err)`, so it's similar to others error messages returning there.

pkg/schedule/windowed_schedule.go
IzabellaRaulin: yes, this is covered in [medium tests] (https://github.com/IzabellaRaulin/snap/blob/run_task_x_times/pkg/schedule/windowed_schedule_medium_test.go) - this function is called on each first "schedule.Wait", so all tests cases apply to that, but especially those ones [here](https://github.com/IzabellaRaulin/snap/blob/run_task_x_times/pkg/schedule/windowed_schedule_medium_test.go#L332)

scheduler/distributed_task_test.go
```diff
@@ -238,24 +245,22 @@ func TestDistributedSubscriptions(t *testing.T) {
...
 			// Start task. We expect it to fail while subscribing deps
-			terrs := sch.StartTask(t.ID())
+			terrs := s.StartTask(t.ID())
 			So(terrs, ShouldNotBeNil)
 			Convey("So dependencies should have been unsubscribed", func() {
 				// Ensure that unsubscribe call count is equal to subscribe call count
 				// i.e that every subscribe call was followed by an unsubscribe since
 				// we errored
 				So(remoteMockManager.UnsubscribeCallCount, ShouldEqual, remoteMockManager.SubscribeCallCount)
-				So(localMockManager.UnsubscribeCallCount, ShouldEqual, localMockManager.UnsubscribeCallCount)
```
IzabellaRaulin: **To reviewers**:
Changed
`So(localMockManager.UnsubscribeCallCount, ShouldEqual, localMockManager.UnsubscribeCallCount)`
(what does not have sense because it's always true)
to 
`So(localMockManager.UnsubscribeCallCount, ShouldEqual, localMockManager.SubscribeCallCount)`

scheduler/distributed_task_test.go
```diff
@@ -265,39 +270,109 @@ func TestDistributedSubscriptions(t *testing.T) {
...
-			remoteMockManager := &subscriptionManager{Fail: false}
-			schTask.RemoteManagers.Add(fmt.Sprintf("127.0.0.1:%v", port1), remoteMockManager)
-			terrs := sch.StartTask(t.ID())
-			So(terrs, ShouldBeNil)
-			Convey("So all depndencies should have been subscribed to", func() {
-				// Ensure that unsubscribe call count is equal to subscribe call count
-				// i.e that every subscribe call was followed by an unsubscribe since
-				// we errored
-				So(localMockManager.SubscribeCallCount, ShouldBeGreaterThan, 0)
-				So(remoteMockManager.SubscribeCallCount, ShouldBeGreaterThan, 0)
```
IzabellaRaulin: Lines 285-301 moved to convey "Task is expected to run until being stopped" in [line 284](https://github.com/intelsdi-x/snap/pull/1548/files#diff-cac84587e02930df0708d14acc453f20R284)

pkg/schedule/windowed_schedule.go
IzabellaRaulin: Changed to "The window stop timestamp and the count cannot be specified simultaneously. The parameter `count` has been ignored.". Is it ok?

PR #1556, opened 2017-03-16
Fixes #1555, autodiscover loaded plugins use configured tmp dir

### LINKED ISSUE #1555: Autoloaded plugins use default temp dir
https://github.com/intelsdi-x/snap/blob/master/control/control.go#L394

When a plugin is discovered using the auto-discovery path, it is loaded to the default temp directory. I think it should load to the path configured in the snapteld configuration like when a plugin is loaded through the rest api.

jtlisi:
Fixes #1555

Summary of changes:
- Plugins loaded through the autodiscover path with utilize the configured temp directory
- Added INFO log to declare temp directory before the start of the autodiscover load loop

Testing done:
- Small Test on Mac OS X

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -313,11 +313,7 @@ func (p *pluginControl) Start() error {
 		controlLogger.WithFields(log.Fields{
 			"_block": "start",
 		}).Info("auto discover path is enabled")
-		tempDirPath := p.GetTempDir()
-		controlLogger.WithFields(log.Fields{
-			"_block":      "start",
-			"tempdirpath": tempDirPath,
-		}).Info("temporary directory path set")
+
 		paths := filepath.SplitList(p.Config.AutoDiscoverPath)
 		p.SetAutodiscoverPaths(paths)
 		for _, pa := range paths {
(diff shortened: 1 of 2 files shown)


rashmigottipati:
LGTM.

rashmigottipati:
I think this can be merged after getting a second review. 
@jcooklin

jtlisi:
Awesome, sounds good!

IzabellaRaulin:
hello @jtlisi - thank you for contribution. Your pull request looks good to me, but I believe that some checking if provided `temp_dir_path` is correct should be made too. 
Also, setting `temp_dir_path` is possible in general, so I would like to move this log from [here](https://github.com/intelsdi-x/snap/pull/1556/files#diff-0c3cd3b2fc4730963e170173bba333e6R317) to another place to be able log about setting temp path not only for autodiscovery case, but for all.  

Based on your pull request, I introduced those points mentioned aboved and I will create a pull request to your forked repo. Please, check if you are ok with my proposed changed.

IzabellaRaulin:
This pull request LGTM, but there is also my code, so I unassigned myself from making a review. Could anyone look on that?

cc @intelsdi-x/snap-maintainers

PR #1557, opened 2017-03-17
Issue 1549: disable parallel xargs where not supported (i.e. on busybox)

### LINKED ISSUE #1549: can't build snap plugins on a busybox system
It appears you can't build the snap plugins on a busybox system, because `scripts/build_plugins.sh` invokes `xargs -P ...` and the busybox version of xargs doesn't support that flag. For portability it seems prudent to detect if we're on a busybox system and drop the -P if it's not supported. Yes, in that case the plugins wouldn't build in parallel and thus build time will be longer, but at least the build will succeed.

gonewest818:
Fixes #1549 

Summary of changes:
- in scripts/build_plugins.sh we now check if this is a busybox system
- if so, don't pass the "-P" flag to xargs, because busybox doesn't support that flag
- also change the logging output to show we're running serially

Testing done:
- performed a make in alpine linux

@intelsdi-x/snap-maintainers

### DIFF
--- a/scripts/build_plugins.sh
+++ b/scripts/build_plugins.sh
@@ -51,6 +51,16 @@ else
 fi
 p=${BUILD_JOBS:-"${p}"}
 
+# disable parallel xargs where not supported (ie on busybox)
+XARGS="xargs -P $p"
+XARGSDESC="in ${p} parallel processes"
+if [[ "${OS}" == "Linux" ]]; then
+  if [[ ! -x "$(command -v readlink)" || "$(basename $(readlink -f $(which xargs)))" == "busybox" ]]; then
+    XARGS="xargs"
+    XARGSDESC="serially"
+  fi
+fi
+
 if [[ "${GOARCH}" == "amd64" ]]; then
   build_path="${__proj_dir}/build/${GOOS}/x86_64"
 else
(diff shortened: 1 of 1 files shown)


scripts/build_plugins.sh
```diff
@@ -51,12 +51,21 @@ else
 fi
 p=${BUILD_JOBS:-"${p}"}
 
+# disable parallel xargs where not supported (ie on busybox)
+if [[ ! -x "$(command -v readlink)" || "$(readlink $(which xargs))" == "/bin/busybox" ]]; then
```
andrzej-k: Not sure, but shouldn't this be ```"$(readlink -f $(which xargs))"``` (added ```-f```)?

gonewest818: OS X doesn't support that -f flag, and I was trying to keep the check simple by not having additional layers of conditionals based on OS.

What I can do is check for busybox only in the Linux case, not the Darwin case. That's probably a safe assumption (have you ever seen a Darwin with busybox ?)

andrzej-k: I'm fine with supporting only Linux case. 
@jcooklin: Are you aware if busybox is supported on Darwin?

gonewest818:
I just pushed a change to my branch that addresses the comment about `readlink -f`, but as discussed, performing this check only when `[ "${OS}" == "Linux" ]` .

PR #1559, opened 2017-03-17
Add plugin entropy

### LINKED ISSUE #1439: Plugin Wanted: Entropy collector
When application is using /dev/random and there is no entropy available in the system it may cause application to block waiting on more randomness.
The most obvious example of applications using entropy are security related (like TLS).

How about creating an entropy collector plugin which will indicate how much entropy is available in the system?

From the design perspective it can just expose one metric (entropy_available), whose value is populated from /proc/sys/kernel/random/entropy_avail.

janczer:
Fixes #1439 

Summary of changes:
- Add to `docs/plugins.yml` snap-plugin-collector-entropy

@intelsdi-x/snap-maintainers


rashmigottipati:
LGTM

Issue #1562, opened 2017-03-20
Windows x86 Version


dishmael:
I have a customer that has about 200 nodes running Windows 7 Pro x86 (32-bit) and need to get a build of Snap that can be installed on those nodes.

jcooklin:
Closed by #1563

PR #1563, opened 2017-03-20
Added 32-bit build

dishmael:
Need a 32-bit build for customers running antiquated hardware.  I opened an issue to track this change here: https://github.com/intelsdi-x/snap/issues/1562

Fixes #

Summary of changes:
- 
- 
- 

Testing done:
- 

@intelsdi-x/snap-maintainers

### DIFF
--- a/scripts/build_all.sh
+++ b/scripts/build_all.sh
@@ -24,5 +24,9 @@ export GOARCH=amd64
 "${__dir}/build_snap.sh" &
 "${__dir}/build_plugins.sh" &
 
+export GOOS=windows32
+export GOARCH=386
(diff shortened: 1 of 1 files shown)


scripts/build_all.sh
```diff
@@ -24,5 +24,9 @@ export GOARCH=amd64
 "${__dir}/build_snap.sh" &
 "${__dir}/build_plugins.sh" &
 
+export GOOS=windows32
```
jcooklin: This needs to be `GOOS=windows`

dishmael: Doesn't that conflict with the earlier declaration of GOOS=windows?

PR #1568, opened 2017-03-22
Fixed flaky tests in scheduler

### LINKED ISSUE #1564: Flaky test in Snap scheduler
This test often fails but not every time.

```
Failures:

  * /home/travis/gopath/src/github.com/intelsdi-x/snap/scheduler/scheduler_medium_test.go 
  Line 273:
  Expected: 'Ended'
  Actual:   'Running'
  (Should be equal)
```

IzabellaRaulin:
Fixes #1564 and #1569 

### Summary of changes:
- added _TaskEnded_ event listener to avoid sleep statements
- fixed flaky tests in scheduler/scheduler_medium_test.go
- fixed flaky tests in scheduler/distribution_workflow_test.go

### Notice:
The best way to determined the right moment to check the state of task seems to be catching TaskEndedEvent instead of making time.Sleep() - and this is a target of https://github.com/intelsdi-x/snap/issues/1569

Testing done:
- medium, legacy  tests

@intelsdi-x/snap-maintainers

### DIFF
--- a/scheduler/distributed_task_test.go
+++ b/scheduler/distributed_task_test.go
@@ -37,6 +37,7 @@ import (
 	"github.com/intelsdi-x/snap/grpc/controlproxy"
 	"github.com/intelsdi-x/snap/pkg/schedule"
 	"github.com/intelsdi-x/snap/plugin/helper"
+	"github.com/intelsdi-x/snap/scheduler/fixtures"
 	"github.com/intelsdi-x/snap/scheduler/wmap"
 	. "github.com/smartystreets/goconvey/convey"
 )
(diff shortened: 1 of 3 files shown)


IzabellaRaulin:
@jcooklin, @andrzej-k thank You for reviewing that, but I changed approach and I would like to ask you to look on that once again.
I think that extending time sleep duration should not be based on the pecentage of the current interval, because there is no relation to an interval at all. @candysmurf reported me that there is still flaky tests when an interval is very small (10ms), so 10% is only 1ms.  So, it should be a constant value, e.g. 100ms seems to be totally enough (includes some reserve).

lmroz:
I'd consider making custom assertion for such kind of tests, which does not require waiting entire allowed time, but can be satisfied and finished earlier. So this time could be arbitrarily long without affecting testing time in typical scenario. But proposed solution is OK for me.

jcooklin:
I'd like to hold merging this for while this is investigated a bit further.

IzabellaRaulin:
@jcooklin - your commit "Adds event listener to avoid sleep statements" is included in this PR

candysmurf:
LGTM

PR #1572, opened 2017-03-27
Add verb to error message

janczer:
Summary of changes:

Before:

    $ snaptel plugin config get test:test:1
    Error requesting info: %!(EXTRA *rbody.Error=invalid plugin type name given test)

After:

    $ snaptel plugin config get test:test:1
    Error requesting info: invalid plugin type name given test


candysmurf:
LGTM

PR #1577, opened 2017-03-28
Fixed tests in distributed task

### LINKED ISSUE #1576: Flaky tests in scheduler/distributed_task_test.go
I received the following failures: 
image

 It looks like those tests reveal that for unsubscribing task dependencies happens in different moments for stopping task vs ended/disabled task.

a) Scenario for stop task:
stopTask() -> unsubscribePlugins -> emit TaskStoppedEvent
:point_right: https://github.com/intelsdi-x/snap/blob/master/scheduler/scheduler.go#L530
(in case TaskStoppedEvent occurs, only taskWatcher handle that: https://github.com/intelsdi-x/snap/blob/master/scheduler/scheduler.go#L749)

b) Scenario for disabled task:
occurring X consecutive failures -> change state to core.TaskDisabled -> emit TaskDisabledEvent -> unsubscribePlugins
:point_right: https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L489
:point_right: https://github.com/intelsdi-x/snap/blob/master/scheduler/scheduler.go#L768

c) Scenario for ended task (very similar to previous scenario in point b)
scheduler has Ended -> change state to core.TaskEnded -> emit TaskEndedEvent -> unsubscribePlugins

:point_right: https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L500
:point_right: https://github.com/intelsdi-x/snap/blob/master/scheduler/scheduler.go#L757

According to tests in distributed_task_test.go relating to Ended state, it might happen that deps are not unsubscribed yet even "TaskEnded" event has been emitted. That is the reason why those two tests are flaky:
:point_right: https://github.com/intelsdi-x/snap/blob/master/scheduler/distributed_task_test.go#L338
:point_right: https://github.com/intelsdi-x/snap/blob/master/scheduler/distributed_task_test.go#L383

### Possible Solution:
Do unsubscribePlugins() before emit TaskEnded/TaskDisabled event

IzabellaRaulin:
Fixes #1576

The following failures are fixed in `scheduler/distributed_task_test.go`:

image

The following failures are fixed in `pkg/schedule/windowed_schedule_medium_test.go`:
```
* /Users/egu/.gvm/pkgsets/go1.7/snap-1.7/src/github.com/intelsdi-x/snap/pkg/schedule/windowed_schedule_medium_test.go
  Line 471:
  Expected: '1'
  Actual:   '0'
  (Should be equal)

  * /Users/egu/.gvm/pkgsets/go1.7/snap-1.7/src/github.com/intelsdi-x/snap/pkg/schedule/windowed_schedule_medium_test.go
  Line 507:
  Expected: '10'
  Actual:   '7'
  (Should be equal)

 * /home/travis/gopath/src/github.com/intelsdi-x/snap/pkg/schedule/windowed_schedule_medium_test.go 
  Line 553:
  Expected: '1'
  Actual:   '0'
  (Should be equal)
```

### Summary of changes:
- added PluginsUnsubscribed event and listener to this event in distributed_workflow_test.go -> react on incoming PluginsUnsubscribed event fixes the flaky tests in `distributed_workflow_tests.go`

- added methods to proceed task disabling -> it's done couple times in different places, it's rather improving code readability, do not impact on how it behaves

- increase an interval to be sure that calculated stop_timestamp (equals to interval multiplied by count) does not pass before test starts -> this fixes tests in windowed_schedule_medium_test.go

### Testing done:
- manual tests
- medium, legacy and small tests

@intelsdi-x/snap-maintainers

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -777,6 +777,13 @@ func (s *scheduler) HandleGomitEvent(e gomit.Event) {
 		task, _ := s.getTask(v.TaskID)
 		task.UnsubscribePlugins()
 		s.taskWatcherColl.handleTaskDisabled(v.TaskID, v.Why)
+	case *scheduler_event.PluginsUnsubscribedEvent:
+		log.WithFields(log.Fields{
+			"_module":         "scheduler-events",
+			"_block":          "handle-events",
+			"event-namespace": e.Namespace(),
+			"task-id":         v.TaskID,
+		}).Debug("event received")
 	default:
 		log.WithFields(log.Fields{
 			"_module":         "scheduler-events",
--- a/scheduler/task.go
+++ b/scheduler/task.go
@@ -366,6 +366,11 @@ func (t *task) UnsubscribePlugins() []serror.SnapError {
 	depGroups := getWorkflowPlugins(t.workflow.processNodes, t.workflow.publishNodes, t.workflow.metrics)
 	var errs []serror.SnapError
 	for k := range depGroups {
+		event := scheduler_event.PluginsUnsubscribedEvent{
+			TaskID:  t.ID(),
+			Plugins: depGroups[k].subscribedPlugins,
+		}
+		defer t.eventEmitter.Emit(event)
 		mgr, err := t.RemoteManagers.Get(k)
 		if err != nil {
 			errs = append(errs, serror.New(err))
--- a/scheduler/task.go
+++ b/scheduler/task.go
@@ -275,8 +275,17 @@ func (t *task) stream() {
 			t.maxMetricsBuffer)
 		if err != nil {
 			consecutiveFailures++
-			e := checkTaskFailures(t, consecutiveFailures)
-			if e != nil {
+			// check task failures
+			if t.stopOnFailure >= 0 && consecutiveFailures >= t.stopOnFailure {
+				taskLogger.WithFields(log.Fields{
+					"_block":               "stream",
+					"task-id":              t.id,
+					"task-name":            t.name,
+					"consecutive failures": consecutiveFailures,
+					"error":                t.lastFailureMessage,
+				}).Error(ErrTaskDisabledOnFailures)
+				// disable the task
+				t.disable(t.lastFailureMessage)
 				return
 			}
 			// If we are unsuccessful at setting up the stream
@@ -321,8 +330,17 @@ func (t *task) stream() {
 					time.Sleep(resetTime)
 					done = true
 				}
-				e := checkTaskFailures(t, consecutiveFailures)
-				if e != nil {
+				// check task failures
+				if t.stopOnFailure >= 0 && consecutiveFailures >= t.stopOnFailure {
+					taskLogger.WithFields(log.Fields{
+						"_block":               "stream",
+						"task-id":              t.id,
+						"task-name":            t.name,
+						"consecutive failures": consecutiveFailures,
+						"error":                t.lastFailureMessage,
+					}).Error(ErrTaskDisabledOnFailures)
+					// disable the task
+					t.disable(t.lastFailureMessage)
 					return
 				}
 			}
@@ -330,28 +348,6 @@ func (t *task) stream() {
 	}
 }
 
-func checkTaskFailures(t *task, consecutiveFailures int) error {
-	if t.stopOnFailure >= 0 && consecutiveFailures >= t.stopOnFailure {
-		taskLogger.WithFields(log.Fields{
-			"_block":               "spin",
-			"task-id":              t.id,
-			"task-name":            t.name,
-			"consecutive failures": consecutiveFailures,
-			"error":                t.lastFailureMessage,
-		}).Error(ErrTaskDisabledOnFailures)
-		// You must lock on state change for tasks
-		t.Lock()
-		t.state = core.TaskDisabled
-		t.Unlock()
-		// Send task disabled event
-		event := new(scheduler_event.TaskDisabledEvent)
-		event.TaskID = t.id
-		event.Why = fmt.Sprintf("Task disabled with error: %s", t.lastFailureMessage)
-		defer t.eventEmitter.Emit(event)
-		return ErrTaskDisabledOnFailures
-	}
-	return nil
-}
 func (t *task) Stop() {
 	t.Lock()
 	defer t.Unlock()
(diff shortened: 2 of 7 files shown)


scheduler/scheduler.go
```diff
@@ -761,9 +761,6 @@ func (s *scheduler) HandleGomitEvent(e gomit.Event) {
 			"event-namespace": e.Namespace(),
 			"task-id":         v.TaskID,
 		}).Debug("event received")
-		// We need to unsubscribe from deps when a task has ended
-		task, _ := s.getTask(v.TaskID)
-		task.UnsubscribePlugins()
```
IzabellaRaulin: **To reviewers**:
Now this happens before emitting the event - see [scheduler/task.go](https://github.com/intelsdi-x/snap/pull/1577/files#diff-77f163e381dd113e90ebd1bd76479b54R566)

scheduler/scheduler.go
```diff
@@ -773,9 +770,6 @@ func (s *scheduler) HandleGomitEvent(e gomit.Event) {
 			"task-id":         v.TaskID,
 			"disabled-reason": v.Why,
 		}).Debug("event received")
-		// We need to unsubscribe from deps when a task goes disabled
-		task, _ := s.getTask(v.TaskID)
-		task.UnsubscribePlugins()
```
IzabellaRaulin: **To reviewers**:
Now this happens before emitting the event - see [scheduler/task.go](https://github.com/intelsdi-x/snap/pull/1577/files#diff-77f163e381dd113e90ebd1bd76479b54R552)

scheduler/task.go
```diff
@@ -321,37 +330,24 @@ func (t *task) stream() {
...
-		t.Unlock()
-		// Send task disabled event
-		event := new(scheduler_event.TaskDisabledEvent)
-		event.TaskID = t.id
-		event.Why = fmt.Sprintf("Task disabled with error: %s", t.lastFailureMessage)
-		defer t.eventEmitter.Emit(event)
-		return ErrTaskDisabledOnFailures
-	}
-	return nil
-}
```
IzabellaRaulin: FYI:
I see the reasoning of having `checkTaskFailures()` to avoid duplicating the same code, but the error log loses its value because this function might be called everywhere, so "_block" does not point to the exact place. So, I changed it to the following:
```
if t.stopOnFailure >= 0 && consecutiveFailures >= t.stopOnFailure {
		taskLogger.WithFields(log.Fields{
			"_block":               "stream",
			"task-id":              t.id,
			"task-name":            t.name,
			"consecutive failures": consecutiveFailures,
			"error":                t.lastFailureMessage,
		}).Error(ErrTaskDisabledOnFailures)
		// disable the task
		t.disable(t.lastFailureMessage)
		return
}
```

candysmurf: @iza, probably "_block" input can be passed in so that code may be reused? Of course, it's everywhere already. Probably it's no big deal for reuse 1 0r 2. If we can refactor to standardized Snap's logs in the future.

scheduler/task.go
```diff
@@ -537,6 +535,37 @@ func (t *task) fire() {
...
+	t.Unlock()
+
+	// Send task disabled event
+	event := new(scheduler_event.TaskDisabledEvent)
+	event.TaskID = t.id
+	event.Why = fmt.Sprintf("Task disabled with error: %s", failureMsg)
+	defer t.eventEmitter.Emit(event)
+
+	// We need to unsubscribe from deps when a task has disabled
+	t.UnsubscribePlugins()
```
candysmurf: @iza, sounds a good change. Only concern is that the disabled task can be enabled. Do we need to subscribe it before enabling it? Could you double check.

scheduler/task.go
```diff
@@ -537,6 +535,37 @@ func (t *task) fire() {
...
+func (t *task) end() {
+	t.Lock()
+	t.state = core.TaskEnded
+	t.Unlock()
+	event := new(scheduler_event.TaskEndedEvent)
+	event.TaskID = t.id
+	defer t.eventEmitter.Emit(event)
+
+	// We need to unsubscribe from deps when a task has ended
+	t.UnsubscribePlugins()
```
candysmurf: do we allow ended task be restarted? If the answer is yes, do we subscribe plugins in the restart?

candysmurf:
@IzabellaRaulin, tried to test your fix on my laptop. It didn't fix the issue. Here are errors for medium-test.

```
Failures:

  * /Users/egu/.gvm/pkgsets/go1.7/snap-1.7/src/github.com/intelsdi-x/snap/pkg/schedule/windowed_schedule_medium_test.go
  Line 471:
  Expected: '1'
  Actual:   '0'
  (Should be equal)

  * /Users/egu/.gvm/pkgsets/go1.7/snap-1.7/src/github.com/intelsdi-x/snap/pkg/schedule/windowed_schedule_medium_test.go
  Line 507:
  Expected: '10'
  Actual:   '7'
  (Should be equal)


47 total assertions

```

scheduler/scheduler.go
IzabellaRaulin: Ok, I will revert that

scheduler/scheduler.go
IzabellaRaulin: ok, reverted

scheduler/task.go
IzabellaRaulin: Answer to 
> Only concern is that the disabled task can be enabled. Do we need to subscribe it before enabling it?
The disabled task can be enabled and yes, we need to subscribe it before enabling it (exactly it is a part of enabling procedure). To be clear this PR does not impact on such behavior. 
> adding error handling
There was a discussion about what should happen first - emitting an event or unsubscribe deps. The decision is to keep it as it is what means emitting event as first, and then do unsubscribing in HandleGomitEvent. I will revert my changes to this aspect.

scheduler/task.go
IzabellaRaulin: As above

scheduler/task.go
IzabellaRaulin: @candysmurf - yes I fully agree. There is still a way to make some small optimization is this area -
but if we decide to do such thing it should be done across all files I suppose. So, is it ok for you as it is now?

candysmurf: totally, 👍

IzabellaRaulin:
Update done - please see also updated PR description. All identified flaky tests related with schedule fixed. 
It's ready to be re-reviewed.

cc: @intelsdi-x/snap-maintainers

PR #1588, opened 2017-04-04
Stopping a task with plugin initiated workflow should not panic

IzabellaRaulin:
Fixes https://github.com/intelsdi-x/snap/issues/1589

#### Before:
Stopping a task  with streaming schedule caused a panic:
image

#### After
Stopping the task behaves as expected:
  - there is no panic, no errors in snapteld logs
  - the task is in state Stopped after that

#### Summary of changes:
- set `done` to true after receiving that killChan has been Closed
- removed `g.Killed()` from control/plugin/client/grpc.go because:
   -  it should be done only for a streaming plugin based on its documentation:
image
  -  this staff here https://github.com/intelsdi-x/snap/blob/master/control/available_plugin.go#L255 seems to be enough

Testing done:
- manual tests

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/available_plugin.go
+++ b/control/available_plugin.go
@@ -236,7 +236,7 @@ func (a *availablePlugin) Stop(r string) error {
 	return a.client.Kill(r)
 }
 
-// Kill assumes aplugin is not able to here a Kill RPC call
+// Kill assumes a plugin is not able to hear a Kill RPC call
 func (a *availablePlugin) Kill(r string) error {
 	log.WithFields(log.Fields{
 		"_module":     "control-aplugin",
@@ -252,6 +252,7 @@ func (a *availablePlugin) Kill(r string) error {
 		}).Debug("deleting available plugin package")
 		os.RemoveAll(filepath.Dir(a.execPath))
 	}
+
 	// If it's a streaming plugin, we need to signal the scheduler that
 	// this plugin is being killed.
 	if c, ok := a.client.(client.PluginStreamCollectorClient); ok {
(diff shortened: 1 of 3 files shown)


candysmurf:
@IzabellaRaulin, do you mind open an issue with the steps to reproduce this issue?

IzabellaRaulin:
@candysmurf, I created https://github.com/intelsdi-x/snap/issues/1589 - there you can find step by step how to reproduce that issue

control/available_plugin.go
```diff
@@ -236,7 +236,7 @@ func (a *availablePlugin) Stop(r string) error {
 	return a.client.Kill(r)
 }
 
-// Kill assumes aplugin is not able to here a Kill RPC call
+// Kill assumes a plugin is not able to here a Kill RPC call
```
jcooklin: s/here/hear/

scheduler/task.go
```diff
@@ -295,8 +295,11 @@ func (t *task) stream() {
 			}
 			select {
 			case <-t.killChan:
+				t.Lock()
 				t.state = core.TaskStopped
-				break
+				t.Unlock()
+				done = true
```
rashmigottipati: After removing g.killed() in control/plugin/client/grpc.go, where are we closing the channel to be able to receive "closing of killChan" here?

control/available_plugin.go
IzabellaRaulin: good catch - I changed it to "hear"

scheduler/task.go
IzabellaRaulin: It works exactly in the same way as for regular spin() - so it happens here: https://github.com/intelsdi-x/snap/blob/master/scheduler/task.go#L360

candysmurf:
LGTM

IzabellaRaulin:
Thank You

PR #1592, opened 2017-04-06
(SDI-2303) Stop a task gracefully allowing it to execute the last scheduled iteration

rashmigottipati:
Summary of changes:
- Have the HandleGomitEvent unsubscribe plugins when event is emitted 
- Block until the task state is in firing (meaning that the workflow didn't complete yet, and we can't stop task immediately)
- Increment hitcount after the workflow completes 
- Add tests in scheduler_test.go and mock_medium_test.go

After the proposed changes:
test1

Testing done:
- make test-medium 
- make test-legacy

@intelsdi-x/snap-maintainers


scheduler/scheduler_test.go
```diff
@@ -45,13 +45,17 @@ type mockMetricManager struct {
...
 }
 
 func (m *mockMetricManager) StreamMetrics(string, map[string]map[string]string, time.Duration, int64) (chan []core.Metric, chan error, []error) {
 	return nil, nil, nil
 }
 
 func (m *mockMetricManager) CollectMetrics(string, map[string]map[string]string) ([]core.Metric, []error) {
+	if m.timeToWait != 0 {
+		time.Sleep(m.timeToWait)
+	}
```
IzabellaRaulin: This 'if' statement is not necessary - you can straightforward do `time.(m.timeToWait)

examples/tasks/mock-file.yaml
```diff
@@ -0,0 +1,14 @@
...
+    interval: "1s"
+  workflow: 
+    collect: 
+      metrics: 
+        /intel/mock/foo: {}
+      config: 
+        /intel/mock: 
+          name: "root"
+          password: "secret"          
+          time_to_wait: "5s"
```
candysmurf: why do we need "time_to_wait" property here?

scheduler/scheduler_test.go
```diff
@@ -45,13 +45,17 @@ type mockMetricManager struct {
 	failValidatingMetricsAfter int
 	failuredSoFar              int
 	autodiscoverPaths          []string
+	timeToWait                 time.Duration
```
candysmurf: why timeToWait?

scheduler/scheduler_test.go
```diff
@@ -199,9 +203,27 @@ func TestScheduler(t *testing.T) {
 		e := s.Start()
 		So(e, ShouldBeNil)
 		// create a simple schedule which equals to windowed schedule without start and stop time
-		_, te := s.CreateTask(schedule.NewWindowedSchedule(time.Second, nil, nil, 0), w, false)
+		t, te := s.CreateTask(schedule.NewWindowedSchedule(time.Second, nil, nil, 0), w, false)
 		So(te.Errors(), ShouldBeEmpty)
 
+		Convey("start task", func() {
+			c.timeToWait = 200 * time.Millisecond
```
candysmurf: You seem using Go Channel in the implementation. Your tests should use same pattern not timeTowait.

scheduler/task.go
```diff
@@ -353,6 +354,16 @@ func (t *task) Stop() {
 	defer t.Unlock()
 	if t.state == core.TaskFiring || t.state == core.TaskSpinning {
 		t.state = core.TaskStopping
+		for {
+			_, ok := <-t.isFiring
```
candysmurf: What is this waiting? It could wait forever. I don't see anywhere single t.isFiring is done.

IzabellaRaulin: Closing the `isFiring` channel happens in `schedulerWorkflow.Start()`, exactly [here](https://github.com/intelsdi-x/snap/pull/1592/files#diff-1d62bbee71bcbb0a2a124a3047dd2aa8R245). 

@rashmigottipati, I have couple concerns about that in existing form:
- maybe do that in `task.fire()` would be more accurate
- based on the `isFiring` channel is not for sending data (data is irrelevant in this case), consider changing its declaration to `chan struct{}`, not bool type

I still have doubts about if this `isFiring` is really necessary - my question is why locking task in [fire()](https://github.com/intelsdi-x/snap/pull/1592/files#diff-77f163e381dd113e90ebd1bd76479b54R538) and in [Stop()](https://github.com/intelsdi-x/snap/pull/1592/files#diff-77f163e381dd113e90ebd1bd76479b54R353) is not enough? @rashmigottipati, could you explain it to me more thoroughly?

IzabellaRaulin:
@rashmigottipati, please fix the issue with hanging tests. FYI, medium tests pass for commented out those lines [scheduler/task.go#L357-L362](https://github.com/rashmigottipati/snap/blob/rashmi/ib/1576-fix_distributed_task_tests/scheduler/task.go#L357-L362), what means it is strictly related to code changes proposed this PR.

rashmigottipati:
I'm closing this PR as I removed few lines that aren't necessary. I will create a new one with the proposed changes.

PR #1597, opened 2017-04-11
Add check plugin type as integer

### LINKED ISSUE #1590: Get configuration of non existing plugin type
If I try get configuration of non existing plugin type:

    $ snaptel plugin config get 1111:11:11

Plugin type will be numeric. First time program will panic, but second and more times it will show empty table.

janczer:
Fixes #1590 

Summary of changes:
- Add check plugin type


jcooklin:
LGTM

@janczer Thanks for your contribution.

PR #1599, opened 2017-04-11
Document Snap with valid swagger spec

### LINKED ISSUE #1573: Document Snap API V2 with OpenAPI Specification (fka Swagger)
Document Snap API V2 with OpenAPI Specification. Benefits:

**Works with various clients**:
* go-swagger generated snap-client-go
* swagger-ui
* curl
* Rest clients
* Snap CLI (snap-cli)

**Generate various clients for different programming languages**
* The specification may be used for generating various clients in different programming languages in the future.

candysmurf:
Fixes #1573 

Summary of changes:
- Added `swagger-go` annotations to Snap RestAPI v2
- Consolidate data models for plugins and tasks
- Renamed a few variables to make `go-swagger` compliance

Testing done:
- make test-small
- make test-medium
- make test-legacy

Note:
`go-swagger` generated client does not work with snap perfectly. The next PR will address the issue #1600 

@intelsdi-x/snap-maintainers


mgmt/rest/v2/config.go
```diff
@@ -33,17 +33,47 @@ type PolicyTable cpolicy.RuleTable
...
 type PolicyTableSlice []cpolicy.RuleTable
 
-// cdata.ConfigDataNode implements it's own UnmarshalJSON
+// PluginConfigItem represents the response of a plugin config items.
+//
+// swagger:response PluginConfigResponse
 type PluginConfigItem struct {
-	cdata.ConfigDataNode
+	// in: body
+	Config cdata.ConfigDataNode `json:"config"`
```
candysmurf: Exposed the config so that clients in different repositories/anywhere can reference it.

mgmt/rest/v2/mock/mock_task_manager.go
```diff
@@ -179,7 +179,7 @@ const (
       "deadline": "4ns",
       "creation_timestamp": -62135596800,
       "last_run_timestamp": -1,
-      "task_state": "Running",
+      "state": "Running",
```
candysmurf: changed `task_state` to `state` so that the JSON representation(currently it's `task_state`) is same as the variable name (`state`). The reason is:
* go-swagger prefers the variable name same as its JSON representation.
* `task_state` only appears in the Snap response but not any input. It's not a breaking change.
* `state` is inside `task` struct. The name `task_state` is redundant.

mgmt/rest/v2/plugin.go
```diff
@@ -40,49 +41,114 @@ import (
...
+	Signed           bool          `json:"signed"`
+	Status           string        `json:"status"`
+	LoadedTimestamp  int64         `json:"loaded_timestamp,omitempty"`
+	Href             string        `json:"href,omitempty"`
+	ConfigPolicy     []PolicyTable `json:"config_policy,omitempty"`
+	HitCount         int           `json:"hitcount,omitempty"`
+	LastHitTimestamp int64         `json:"last_hit_timestamp,omitempty"`
+	ID               uint32        `json:"id,omitempty"`
+	PprofPort        string        `json:"pprof_port,omitempty"`
+}
```
candysmurf: Merged `RunningPlugin` with 'Plugin` data type. The difference can be omitted in the JSON representation in terms of different plugin response scenarios. The reason is to simplify the plugin data model in swagger spec.

mgmt/rest/v2/task.go
```diff
@@ -49,8 +109,10 @@ type Task struct {
 	MissCount          int               `json:"miss_count,omitempty"`
 	FailedCount        int               `json:"failed_count,omitempty"`
 	LastFailureMessage string            `json:"last_failure_message,omitempty"`
-	State              string            `json:"task_state"`
-	Href               string            `json:"href"`
+	State              string            `json:"state,omitempty"`
+	Href               string            `json:"href,omitempty"`
+	Start              bool              `json:"start,omitempty"`
+	MaxFailures        int               `json:"max-failures,omitempty"`
```
candysmurf: Consolidated `Task` data model from Snap CLI and Rest API so that they reference the same `Task` data model in Swagger spec. The benefit is that no business logic in CLIs/clients and Snap server defines the client behaviors universally.

scheduler/wmap/wmap.go
```diff
@@ -282,10 +285,11 @@ func (c *CollectWorkflowMapNode) AddConfigItem(ns, key string, value interface{}
 }
 
 type ProcessWorkflowMapNode struct {
-	Name         string                   `json:"plugin_name"yaml:"plugin_name"`
-	Version      int                      `json:"plugin_version"yaml:"plugin_version"`
-	ProcessNodes []ProcessWorkflowMapNode `json:"process,omitempty"yaml:"process"`
-	PublishNodes []PublishWorkflowMapNode `json:"publish,omitempty"yaml:"publish"`
+	// required: true
+	PluginName    string                   `json:"plugin_name"yaml:"plugin_name"`
+	PluginVersion int                      `json:"plugin_version"yaml:"plugin_version"`
```
candysmurf: Changed `Name` to `PluginName` and 'Version` to `PluginVersion` for the consistency between variable names and JSON representation. The consequence is:
* `go-swagger` generated client can reference back easily and without patching shell scripts.
*  variable name changes are invisible to customers. Changes only affect tests inside code base.

candysmurf:
Issue #1600 will be addressed in the subsequence PRs.

mgmt/rest/v2/mock/mock_task_manager.go
jcooklin: We cannot change the response of the API after it's landed (V1 or V2).  Please revert this back to `task_state`.

jcooklin:
@candysmurf this looks good.  The only comment I have at this point is the one I made on `task_state`.  **I would like to get two other maintainers to give a LGTM before we land this** as there are several other changes backed up behind it that will also need reviews by folks with full context in understanding what else is coming (generated clients, new CLI based on generated clients, exposing swagger UI through snapteld, etc).   

CC:  @kjlyon

candysmurf:
@jcooklin, thanks for reviewing it. `task_state` is reverted back. If you like to take another look.

candysmurf:
@intelsdi-x/snap-maintainers, this PR has been there for awhile. It affects subsequence work related to this initiative. Your timely code review is greatly appreciated.

mgmt/rest/v2/plugin.go
kjlyon: Will this change the response, causing a breaking change?

scheduler/wmap/wmap.go
```diff
@@ -365,9 +369,10 @@ func (p *ProcessWorkflowMapNode) GetConfigNode() (*cdata.ConfigDataNode, error)
 }
 
 type PublishWorkflowMapNode struct {
-	Name    string `json:"plugin_name"yaml:"plugin_name"`
-	Version int    `json:"plugin_version"yaml:"plugin_version"`
-	// TODO publisher config
```
kjlyon: This TODO was taken out. Unless you did this work, it should be added back in.

mgmt/rest/v2/plugin.go
candysmurf: No, it'll not change responses. Please test yourself and let me know if you find anything different.

.travis.yml
```diff
@@ -1,10 +1,12 @@
-sudo: false
+sudo: true
 language: go
 go:
 - 1.7.3
 before_install:
 - bash scripts/gitcookie.sh
 - go get github.com/smartystreets/goconvey/convey
+- echo "deb https://dl.bintray.com/go-swagger/goswagger-debian ubuntu main" | sudo tee -a /etc/apt/sources.list
+- sudo apt-get update && sudo apt-get install -y swagger --force-yes
```
candysmurf: this script installs swagger to travis env.

swagger.json
```diff
@@ -0,0 +1,1344 @@
...
+    "contact": {
+      "name": "Snap Maintainer",
+      "url": "https://github.com/intelsdi-x/snap",
+      "email": "snap@intelsdi-x.slack.com"
+    },
+    "license": {
+      "name": "Apache 2.0",
+      "url": "http://www.apache.org/licenses/"
+    },
+    "version": "1.1.0"
```
IzabellaRaulin: I have a  question about the version - it looks like the version of Snap (however the latest one is 1.2.0, not 1.1.0). How should it be managed when a new release is coming? Or maybe it's rather about version of Snap Restful API? Could you explain me that? Thank You

swagger.json
```diff
@@ -0,0 +1,1344 @@
...
+        "type": "object",
+        "properties": {
+          "task_watch": {
+            "$ref": "#/definitions/StreamedTaskEvent"
+          }
+        }
+      }
+    },
+    "TasksResponse": {
+      "description": "TasksResp returns a list of created tasks.",
```
IzabellaRaulin: In desciption - `TasksResp` or `TasksResponse`?

core/schedule.go
```diff
@@ -27,9 +27,15 @@ import (
...
 
+// Schedule defines a scheduler.
+//
+// swagger:model Schedule
 type Schedule struct {
-	Type           string     `json:"type,omitempty"`
-	Interval       string     `json:"interval,omitempty"`
+	// required: true
+	// enum: simple, windowed, streaming, cron
+	Type string `json:"type"`
```
candysmurf: it's beneficial that users can select a type from an available GUI list instead of by leaving it empty although we could default it to `simple`. It's not clear when we have public exposing APIs.

candysmurf: of course, for the consistent reason or whatever, if team insists to leave it as not required. I will do that but adding a default value.

swagger.json
```diff
@@ -0,0 +1,1344 @@
...
+      "x-go-package": "github.com/intelsdi-x/snap/mgmt/rest/v2"
+    },
+    "ProcessWorkflowMapNode": {
+      "type": "object",
+      "required": [
+        "plugin_name"
+      ],
+      "properties": {
+        "config": {
+          "description": "TODO processor config",
```
IzabellaRaulin: This one should not have such description. Of course, this points to that [this line](https://github.com/intelsdi-x/snap/pull/1599/files#diff-e1f961b73cb3ace6fed8796b07b717fdR293) should be changed.

mgmt/rest/v2/config.go
```diff
@@ -33,17 +33,47 @@ type PolicyTable cpolicy.RuleTable
...
+	Config []string `json:"config"`
 }
 
 func (s *apiV2) getPluginConfigItem(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
 	var err error
 	styp := p.ByName("type")
 	if styp == "" {
 		cdn := s.configManager.GetPluginConfigDataNodeAll()
-		item := &PluginConfigItem{ConfigDataNode: cdn}
+		item := &PluginConfigItem{Config: cdn}
```
IzabellaRaulin: Small suggestion: changing `cdn` to for example `cfg` 
Why: after changing the name ConfigDataNode to Config, `cdn` does not make sense as it had before.

mgmt/rest/v2/config.go
```diff
@@ -65,7 +95,7 @@ func (s *apiV2) getPluginConfigItem(w http.ResponseWriter, r *http.Request, p ht
 	}
 
 	cdn := s.configManager.GetPluginConfigDataNode(typ, name, iver)
-	item := &PluginConfigItem{ConfigDataNode: cdn}
+	item := &PluginConfigItem{Config: cdn}
```
IzabellaRaulin: As above, small suggestion: changing `cdn` to for example `cfg` 
Why: after changing the name ConfigDataNode to Config, `cdn` does not make sense as it had before.

mgmt/rest/v2/api.go
```diff
@@ -56,25 +56,302 @@ func New(wg *sync.WaitGroup, killChan chan struct{}, protocol string) *apiV2 {
 
 func (s *apiV2) GetRoutes() []api.Route {
 	routes := []api.Route{
-		// plugin routes
+		// swagger:route GET /plugins plugins getPlugins
+		//
```
candysmurf: No, comments have to stay like this format in order for them to show up in UI pretty. Otherwise, it shows up but not as pretty as it's.

mgmt/rest/v2/task.go
```diff
@@ -33,14 +33,74 @@ import (
...
+	// required: true
+	Action string `json:"action"`
+}
+
+// Task represents Snap task definition.
 type Task struct {
-	ID                 string            `json:"id"`
-	Name               string            `json:"name"`
-	Deadline           string            `json:"deadline"`
+	Version            int               `json:"version,omitempty"`
```
IzabellaRaulin: Does Task has own version? Is it not a mistake?

mgmt/rest/v2/metric.go
```diff
@@ -33,27 +33,53 @@ import (
 	"github.com/julienschmidt/httprouter"
 )
 
+// MetricsResp is the representation of metric operation response.
+//
+// swagger:response MetricsResponse
+type MetricsResp struct {
```
candysmurf: The purpose for this is that APIs have appropriate response types.

mgmt/rest/v2/metric.go
```diff
@@ -33,27 +33,53 @@ import (
...
-	Version                 int              `json:"version,omitempty"`
-	Dynamic                 bool             `json:"dynamic"`
-	DynamicElements         []DynamicElement `json:"dynamic_elements,omitempty"`
-	Description             string           `json:"description,omitempty"`
-	Unit                    string           `json:"unit,omitempty"`
-	Policy                  PolicyTableSlice `json:"policy,omitempty"`
-	Href                    string           `json:"href"`
+	LastAdvertisedTimestamp int64 `json:"last_advertised_timestamp,omitempty"`
+	//required: true
+	Namespace string `json:"namespace"`
```
candysmurf: Do you agree the namespace of a metric should be required? Would you please let me know in what use case where a JSON/YAML representation won't specify the namespace for a metric?

scheduler/wmap/wmap.go
candysmurf: I can add it back but I really don't think it's still relevant. We can specify a config within a publisher.

swagger.json
candysmurf: good catch, It should be the version of our API.

mgmt/rest/v2/task.go
candysmurf: yes, currently task has its own version if you look the task manifest. Does the task version make sense? idk.

swagger.json
```diff
@@ -0,0 +1,1344 @@
...
+            "items": {
+              "$ref": "#/definitions/Metric"
+            },
+            "x-go-name": "Metrics"
+          }
+        }
+      }
+    },
+    "PluginConfigResponse": {
+      "description": "PluginConfigItem represents the response of a plugin config items.",
```
IzabellaRaulin: @candysmurf, please verify if the description is as you expected. I have a doubt about PluginConfigItem - maybe it should be PluginConfigResponse.

mgmt/rest/v2/task.go
IzabellaRaulin: You are right - task has a version used to differentiate between versions of **the task manifest parser** (right now, there is only one version: 1). For others reviewers - you can find more about it [here](https://github.com/intelsdi-x/snap/blob/master/docs/TASKS.md#version)

I have just a concern why `Task` struct did not have `Version` field before, but I need to be more acquainted with the flow of task version to say sth more. 

@candysmurf - what do you think about moving `Version` after `ID` and `Name`? Why: task ID and its name identified the task, so a good pattern keeps those two in the beginning. This is only a small suggestion, so up to you.

candysmurf: @IzabellaRaulin, changed the version as suggested, please let me know if you have other suggestions.

mgmt/rest/v2/mock/mock_config_manager.go
```diff
@@ -70,19 +70,25 @@ func (MockConfigManager) DeletePluginConfigDataNodeFieldAll(fields ...string) cd
 // rest_v2_test.go on the plugin config routes found in mgmt/rest/server.go
 const (
 	SET_PLUGIN_CONFIG_ITEM = `{
-  "user": "Jane"
+  "config": {
```
jcooklin: Can we avoid the breaking API change?  We should avoid changing response objects.

swagger.json
```diff
@@ -0,0 +1,1344 @@
...
+        },
+        "plugin_version": {
+          "type": "integer",
+          "format": "int64",
+          "x-go-name": "PluginVersion"
+        },
+        "process": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/ProcessWorkflowMapNode"
```
mkleina: There is circular reference, which causes swagger-ui to crash (RangeError: Maximum call stack size exceeded). Here is the definition of ProcessWorkflowMapNode:
```go
type ProcessWorkflowMapNode struct {
	// required: true
	PluginName    string                   `json:"plugin_name"yaml:"plugin_name"`
	PluginVersion int                      `json:"plugin_version"yaml:"plugin_version"`
	ProcessNodes  []ProcessWorkflowMapNode `json:"process,omitempty"yaml:"process"`
	PublishNodes  []PublishWorkflowMapNode `json:"publish,omitempty"yaml:"publish"`
	// Config the configuration of a processor.
	Config map[string]interface{} `json:"config,omitempty"yaml:"config"`
	Target string                 `json:"target"yaml:"target"`
}
```

scripts/swagger.sh
```diff
@@ -0,0 +1,27 @@
...
+#distributed under the License is distributed on an "AS IS" BASIS,
+#WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#See the License for the specific language governing permissions and
+#limitations under the License.
+
+__dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+swagger generate spec -o ${__dir}/../swagger.json
+
+perl -i -p -e 's/ProcessNodes/Process/g' ${__dir}/../swagger.json
```
mkleina: Is it really needed to rename `ProcessNodes` to `Process`? I think that we shouldn't modify auto-generated files.

swagger.json
mkleina: There is circular reference, which causes swagger-ui to crash (RangeError: Maximum call stack size exceeded). Here is the definition of ProcessWorkflowMapNode:
```go
type ProcessWorkflowMapNode struct {
	// required: true
	PluginName    string                   `json:"plugin_name"yaml:"plugin_name"`
	PluginVersion int                      `json:"plugin_version"yaml:"plugin_version"`
	ProcessNodes  []ProcessWorkflowMapNode `json:"process,omitempty"yaml:"process"`
	PublishNodes  []PublishWorkflowMapNode `json:"publish,omitempty"yaml:"publish"`
	// Config the configuration of a processor.
	Config map[string]interface{} `json:"config,omitempty"yaml:"config"`
	Target string                 `json:"target"yaml:"target"`
}
```
Otherwise, it can be also swagger issue:
https://github.com/swagger-api/swagger-ui/issues/1646

**NOTE**: Works OK with swagger-ui 2.2.9, error on swagger-ui 3.0.5

jcooklin: I think we should wait for the next release.  The issue you referenced looks promising in that they appear to be addressing it.

scripts/swagger.sh
candysmurf: @mkleina, otherwise generated client does not work with Snap server as the data model is different. Unless we rename our variables inside Snap code base. I don't think @jcooklin let me do that.

mgmt/rest/v2/mock/mock_config_manager.go
candysmurf: @jcooklin, the response change is from word `body` to `config`. if that's your concern, I can change it to `body`. In this context, `config` makes more sense. Please try it. With API V1, after set config, the response is:

**V1 API response after updating a config item:**
```
{
  "meta": {
    "code": 200,
    "message": "Plugin config item(s) set",
    "type": "config_plugin_item_created",
    "version": 1
  },
  "body": {
    "leader": "thunder",
    "password": "p@ssw0rd",
    "path": "whatever",
    "user": "whoever"
  }
}
``` 

**V2 API response after updating a config item:**
```
{
"config": {
"leader": "thunder",
"password": "p@ssw0rd",
"path": "whatever",
"user": "whoever"
}
}
```

candysmurf: @jcooklin, I tried to remove "config" totally. Currently, there is no way to do that in go-swagger. The only way is to add a wrap in snap-client-go to return the response we wanted. Please let me know if you like to do that?

mgmt/rest/v2/config.go
```diff
@@ -33,17 +33,51 @@ type PolicyTable cpolicy.RuleTable
...
 
-// cdata.ConfigDataNode implements it's own UnmarshalJSON
 type PluginConfigItem struct {
 	cdata.ConfigDataNode
 }
 
+// PluginConfigResponse represents the response of a plugin config items.
+//
+// swagger:response PluginConfigResponse
+type PluginConfigResponse struct {
```
mkleina: @candysmurf It will be used in future PR as I understand? Because it is not used anywhere now.

mkleina: @candysmurf `PluginConfigResponse` is not used anywhere, but it will change in future PR, do I understand correctly?

mgmt/rest/v2/error.go
```diff
@@ -40,6 +40,16 @@ var (
...
 )
 
+// ErrorResponse represents the Snap error response type.
+//
+// It includes an error message and a map of fields.
+//
+// swagger:response ErrorResponse
+type ErrorResponse struct {
+	// in:body
+	SnapError Error `json: "snap_error""`
```
mkleina: @candysmurf Mistype here probably, double `"`

mgmt/rest/v2/config.go
```diff
@@ -105,7 +139,7 @@ func (s *apiV2) deletePluginConfigItem(w http.ResponseWriter, r *http.Request, p
 		res = s.configManager.DeletePluginConfigDataNodeField(typ, name, iver, src...)
 	}
 
-	item := &PluginConfigItem{ConfigDataNode: res}
+	item := &PluginConfigItem{res}
```
mkleina: @candysmurf Will this (and all other responses) be changed to `PluginConfigResponse` in the future PR?

mgmt/rest/v2/config.go
jcooklin: This provides the (generated) client with a response object for plugin config calls.  You will notice that this struct is referenced in api.go (e.g. [here](https://github.com/candysmurf/snap/blob/swagger-1/mgmt/rest/v2/api.go#L141)).

mgmt/rest/v2/config.go
jcooklin: It's unlikely IMO.  Like mentioned above the response struct is to effectively support generating the client.

jcooklin:
@candysmurf once @mkleina finishes reviewing please squash the commits.

mgmt/rest/v2/config.go
mkleina: Ok, so all `Response` structs are just for declaring swagger specification and generating client, now I understand. I was confused cause I was looking on `/plugins` call swagger spec:

```
// Responses:
// 200: PluginsResponse
```

and `getPlugins()` handler method:

```go
Write(200, PluginsResponse{Plugins: filteredPlugins}, w)
```

and in this case the response struct passed to `Write()` is the same as in swagger spec, but now I see that even in this case, the "real" PluginsResponse for generating client is actually `PluginsResp`:

```go
// PluginsResp represents the response from plugins operations.
//
// swagger:response PluginsResponse
type PluginsResp struct {
	// List of plugins
	//
	// in: body
	Body struct {
		Plugins []Plugin `json:"plugins,omitempty"`
	}
}
```

mkleina: Ok, so all `Response` structs are just for declaring swagger specification and generating client, now I understand. I was confused cause I was looking on `/plugins` call swagger tags:

```
// Responses:
// 200: PluginsResponse
```

and `getPlugins()` handler method:

```go
Write(200, PluginsResponse{Plugins: filteredPlugins}, w)
```

and in this case the response struct passed to `Write()` is the same as in swagger tags, but now I see that even in this case, the "real" PluginsResponse for generating client is actually `PluginsResp`:

```go
// PluginsResp represents the response from plugins operations.
//
// swagger:response PluginsResponse
type PluginsResp struct {
	// List of plugins
	//
	// in: body
	Body struct {
		Plugins []Plugin `json:"plugins,omitempty"`
	}
}
```

candysmurf: @mkleina, glad that you got it!

Issue #1601, opened 2017-04-12
Plugins should be unsubscribed by the event handler after the task state is Stopped


rashmigottipati:
**Current behavior:**
When a stop task request is processed by the scheduler, it immediately starts unsubscribing all dependencies (plugins) of this task. This results in the plugins being used by the current firing of the workflow being killed.

**Desired behavior:**
Plugins should be unsubscribed after the task state is changed to Stopped state.  There is an intermediate state between firing and stopped (i.e. Stopping). When the task state is changed to stopping, we close the channel "killchan" and closing of this killchan is received in the spin go routine. Changing of task state from stopping to stopped takes place here. So, plugins should be unsubscribed: 
- By the event handler in scheduler
- After task state is changed to stopped (meaning that the task is truly stopped and the workflow finished executing)

PR #1604, opened 2017-04-12
Fixes #1598 and #1601 - ability to stop tasks gracefully

### LINKED ISSUE #1598: Stopping task before the workflow finishes executing
The Athena team has experienced issues where the control in Swan stops the task before it gets to push its last metrics to the Cassandra database and thus lose a data point. The request is to have the ability to let the currently scheduled run of a task complete when we issue stop task request, such that no data points are lost.

**Possible Solution:**
The request to stop task would have occurred after incrementing the hitCount (which indicates the number of times the task has fired) and before firing the task. In this case, the hitCount = 1 (as an example) but the task hasn't fired, which implies that the workflow didn't execute as well. From the user's perspective, as the hitCount = 1, this leads to the assumption that the task has fired and the workflow finished executing but there are no metrics collected. Currently, Snap already provides graceful shutdown, which means that if a task if firing and hasn't finished executing the workflow, it won't preempt the task and unsubscribe the plugins immediately. It waits till the workflow finishes executing and then stops the task. To fix the problem that Athena team has experienced, we need to increment hitCount after the workflow finishes executing and successfully returns. This way, as a user, if I see that the task has hitCount = 1, it's guaranteed that there will be one data entry. In the case where stop task request is issued before even the workflow is executed, the hitCount will be 0 and the user wouldn't be expecting any data as well.

rashmigottipati:
Fixes #1598 and #1601 

Summary of changes:
- Plugins are unsubscribed by the event handler after the task state is Stopped
- Increment hitCount after the workflow finishes executing
- TaskStoppedEvent is emitted after the task state is Stopped

Testing done:
- manual test using mock
- legacy tests

@intelsdi-x/snap-maintainers

### DIFF
--- a/scheduler/scheduler_medium_test.go
+++ b/scheduler/scheduler_medium_test.go
@@ -43,6 +43,7 @@ type mockMetricManager struct {
 	failValidatingMetricsAfter int
 	failuredSoFar              int
 	autodiscoverPaths          []string
+	timeToWait                 time.Duration
 }
 
 func (m *mockMetricManager) StreamMetrics(string, map[string]map[string]string, time.Duration, int64) (chan []core.Metric, chan error, []error) {
@@ -50,6 +51,7 @@ func (m *mockMetricManager) StreamMetrics(string, map[string]map[string]string, 
 }
 
 func (m *mockMetricManager) CollectMetrics(string, map[string]map[string]string) ([]core.Metric, []error) {
+	time.Sleep(m.timeToWait)
 	return nil, nil
 }
 
@@ -349,9 +351,12 @@ func TestCreateTask(t *testing.T) {
 
 func TestStopTask(t *testing.T) {
 	logrus.SetLevel(logrus.FatalLevel)
-	s := newScheduler()
-	s.Start()
+	c := new(mockMetricManager)
+	cfg := GetDefaultConfig()
+	s := New(cfg)
+	s.SetMetricManager(c)
 	w := newMockWorkflowMap()
+	s.Start()
 
 	Convey("Calling StopTask on a running task", t, func() {
 		sch := schedule.NewWindowedSchedule(interval, nil, nil, 0)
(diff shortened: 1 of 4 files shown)


plugin/collector/snap-plugin-collector-mock2/mock/mock.go
```diff
@@ -81,6 +81,15 @@ func (f *Mock) CollectMetrics(mts []plugin.MetricType) ([]plugin.MetricType, err
...
 
+		if c, ok := mts[i].Config().Table()["time_to_wait"]; ok {
+			dur, err := time.ParseDuration(c.(ctypes.ConfigValueStr).Value)
+			if err != nil {
+				log.Println(err.Error())
+			} else {
+				time.Sleep(dur)
+			}
+		}
+
```
IzabellaRaulin: @rashmigottipati, I know that introducing `time_to_wait` in snap-plugin-collector-mock2 was really helpful for testing gracefully shutdown behavior during working on it. However, I have a doubt about landing this change on master - is it really needed? Does it provide some value - I mean is there any large test which runs mock2 collector and testing this behavior? If yes, that is fine - I won't insist on removing that.
 -

rashmigottipati: @IzabellaRaulin thanks for testing it. I agree with you - I don't think it's necessary to land this change on master. I added another test in scheduler_test.go which has timeToWait in collect method and an assertion on the total elapsed time. In my opinion, it would suffice to land this test on master.

jcooklin: If you remove this don't forget to remove the example task that was addded too.

scheduler/scheduler_test.go
```diff
@@ -202,6 +204,22 @@ func TestScheduler(t *testing.T) {
...
+			So(t.State(), ShouldResemble, core.TaskSpinning)
+			time.Sleep(50 * time.Millisecond)
+			t.(*task).Stop()
+			end := time.Now().UnixNano() / int64(time.Millisecond)
+			elapsed := time.Duration(end - start)
+			So(elapsed*1e6, ShouldBeGreaterThanOrEqualTo, 500*time.Millisecond)
+			time.Sleep(50 * time.Millisecond)
+			So(t.State(), ShouldResemble, core.TaskStopped)
+			c.timeToWait = 0
+		})
```
IzabellaRaulin: @rashmigottipati , I have doubt about this test.  I will try to propose something another instead of doing that in this way. I will let you know when I have sth.

IzabellaRaulin:
Besides comments above, general code LGTM

rashmigottipati:
I deleted the mock test example and changed the medium test to listen on TaskStoppedEvent instead of using time.Sleep(). 
@jcooklin @IzabellaRaulin

scheduler/scheduler_medium_test.go
```diff
@@ -372,6 +377,35 @@ func TestStopTask(t *testing.T) {
...
+			// stop task when task state is firing
+			t.(*task).Stop()
+			// the last scheduled workflow execution should be allowed to finish
+			// so we expect that stopping the task is going to happen not early than 500ms (set by by timeToWait)
+			So(t.HitCount(), ShouldEqual, 1)
+			//the task should have fired once
+			select {
+			case <-lse.TaskStoppedEvents:
+				So(time.Since(startTime), ShouldBeGreaterThan, c.timeToWait)
+				//elapsed time should be greater than 500ms
```
IzabellaRaulin: (small suggestion) This comment in line 402 should be above the code in line 401

rashmigottipati: Fixed it.

IzabellaRaulin: Similar suggestions for the following lines 398 & 397, 391 & 390

rashmigottipati: Fixed them.

scheduler/scheduler_medium_test.go
```diff
@@ -54,6 +55,7 @@ func (m *mockMetricManager) CollectMetrics(string, map[string]map[string]string)
 }
 
 func (m *mockMetricManager) PublishMetrics([]core.Metric, map[string]ctypes.ConfigValue, string, string, int) []error {
+	time.Sleep(m.timeToWait)
```
IzabellaRaulin: Even that is only for testing purpose, I suggest moving this time.Sleep() to CollectMetrics rather than having that in the last step in the workflow which is PublishMetrics. The another reason is that collection normally takes longer than publishing, what seems to be more accurate to reality.

rashmigottipati: I intended to add this to CollectMetrics. Thanks for pointing it out. Fixed it.

scheduler/scheduler_medium_test.go
```diff
@@ -372,6 +377,35 @@ func TestStopTask(t *testing.T) {
...
+			t.(*task).Spin()
+			//allowing things to settle and waiting for task state to change to firing
+			time.Sleep(100 * time.Millisecond)
+			So(t.State(), ShouldResemble, core.TaskFiring)
+			// stop task when task state is firing
+			t.(*task).Stop()
+			// the last scheduled workflow execution should be allowed to finish
+			// so we expect that stopping the task is going to happen not early than 500ms (set by by timeToWait)
+			//the task should have fired once
+			So(t.HitCount(), ShouldEqual, 1)
```
IzabellaRaulin: Please move checking the hitCount into the `case <-lse.TaskStoppedEvents` after the checking that task is Stopped (so the HitCount for 100% sure does not change)

PR #1617, opened 2017-04-28
Aarontay/standalone plugin

aarontay:
Fixes #

Summary of changes:
- Added support for loading standalone plugins remotely
- Added checks to verify if plugin has a remote URI
- Updated unload / unsubscribe to handle remote plugins

Testing done:
- Tested with remote plugin 'snap-plugin-collector-rand' example in snap-plugin-lib-go and was able to load/create tasks/unload/terminate daemon when using remote plugins. Verified non-remote plugins were unaffected by new standalone plugin logic.

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -33,9 +33,9 @@ import (
 	"time"
 
 	log "github.com/Sirupsen/logrus"
-	"github.com/intelsdi-x/gomit"
 	"google.golang.org/grpc"
 
+	"github.com/intelsdi-x/gomit"
 	"github.com/intelsdi-x/snap/control/plugin"
 	"github.com/intelsdi-x/snap/control/plugin/client"
 	"github.com/intelsdi-x/snap/control/strategy"
@@ -587,6 +587,11 @@ func (p *pluginControl) verifySignature(rp *core.RequestedPlugin) (bool, serror.
 }
 
 func (p *pluginControl) returnPluginDetails(rp *core.RequestedPlugin) (*pluginDetails, serror.SnapError) {
+	if rp.Uri() != nil {
+		return &pluginDetails{
+			Uri: rp.Uri(),
+		}, nil
+	}
 	details := &pluginDetails{}
 	var serr serror.SnapError
 	//Check plugin signing
@@ -725,6 +730,13 @@ func (p *pluginControl) UnsubscribeDeps(id string) []serror.SnapError {
 }
 
 func (p *pluginControl) verifyPlugin(lp *loadedPlugin) error {
+	if lp.Details.Uri != nil {
+		// remote plugin
+		if core.IsUri(lp.Details.Uri.String()) {
+			return fmt.Errorf(fmt.Sprintf("Remote plugin failed to load: bad uri: (%x)", lp.Details.Uri))
+		}
+		return nil
+	}
 	b, err := ioutil.ReadFile(lp.Details.Path)
 	if err != nil {
 		return err
--- a/control/plugin_manager.go
+++ b/control/plugin_manager.go
@@ -23,8 +23,12 @@ package control
 
 import (
 	"crypto/sha256"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -170,6 +174,7 @@ type pluginDetails struct {
 	KeyPath     string
 	CACertPaths string
 	TLSEnabled  bool
+	Uri         *url.URL
 }
 
 type loadedPlugin struct {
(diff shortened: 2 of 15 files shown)


control/plugin_manager.go
```diff
@@ -321,52 +329,70 @@ func (p *pluginManager) LoadPlugin(details *pluginDetails, emitter gomit.Emitter
 	lPlugin.Details = details
 	lPlugin.State = DetectedState
 
+	//TODO (JC) deal with path
 	pmLogger.WithFields(log.Fields{
 		"_block": "load-plugin",
-		"path":   filepath.Base(lPlugin.Details.Exec[0]),
+		// "path":   filepath.Base(lPlugin.Details.Exec[0]),
```
rashmigottipati: This line should be uncommented.

mgmt/rest/v1/plugin.go
```diff
@@ -176,6 +177,27 @@ func (s *apiV1) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
...
 		lp.LoadedPlugins = append(lp.LoadedPlugins, catalogedPluginToLoaded(r.Host, pl))
 		rbody.Write(201, lp, w)
+	} else if strings.HasSuffix(mediaType, "json") {
+		body, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			rbody.Write(500, rbody.FromError(err), w)
+			return
+		}
+		var resp map[string]string
+		err = json.Unmarshal(body, &resp)
```
rashmigottipati: The error should be handled here. It should be logged as log.Fatal(err) or log.Fatal("invalid config")

rashmigottipati: The error should be handled here. It should be logged as log.Fatal(err) or log.Fatal("invalid config")

control/plugin_manager.go
```diff
@@ -156,14 +162,16 @@ func (l *loadedPlugins) findLatest(typeName, name string) (*loadedPlugin, error)
...
-	IsPackage bool
-	Manifest  *schema.ImageManifest
-	Path      string
-	Signed    bool
-	Signature []byte
+	CheckSum     [sha256.Size]byte
+	Exec         []string
+	ExecPath     string
+	IsPackage    bool
+	IsAutoLoaded bool
```
rashmigottipati: The field IsAutoLoaded isn't used anywhere. This should be removed.

control/control.go
```diff
@@ -572,6 +572,12 @@ func (p *pluginControl) verifySignature(rp *core.RequestedPlugin) (bool, serror.
 }
 
 func (p *pluginControl) returnPluginDetails(rp *core.RequestedPlugin) (*pluginDetails, serror.SnapError) {
+	if rp.Uri() != nil {
+		return &pluginDetails{
+			Uri:          rp.Uri(),
+			IsAutoLoaded: true,
```
rashmigottipati: Remove the line `IsAutoLoaded: true` after this field is removed from the struct in plugin_manager.go.

control/control.go
```diff
@@ -706,6 +712,10 @@ func (p *pluginControl) UnsubscribeDeps(id string) []serror.SnapError {
 }
 
 func (p *pluginControl) verifyPlugin(lp *loadedPlugin) error {
+	if lp.Details.Uri != nil {
+		// remote plugin
+		return nil
```
rashmigottipati: Is there any way to check for a valid uri? If it's an invalid uri, we can return err here instead of nil.

control/plugin_manager.go
```diff
@@ -321,52 +329,75 @@ func (p *pluginManager) LoadPlugin(details *pluginDetails, emitter gomit.Emitter
...
-		return nil, serror.New(ErrPluginAlreadyLoaded, map[string]interface{}{
-			"plugin-name":    resp.Meta.Name,
-			"plugin-version": resp.Meta.Version,
-			"plugin-type":    resp.Type.String(),
-		})
+		body, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			return nil, serror.New(err)
+		}
+		json.Unmarshal(body, &resp)
```
rashmigottipati: Error for Unmarshal should be checked here.

control/runner.go
```diff
@@ -368,12 +368,26 @@ func (r *runner) handleUnsubscription(pType, pName string, pVersion int, taskID
...
-			"_block":                  "handle-unsubscription",
-			"pool-count":              pool.Count(),
-			"pool-subscription-count": pool.SubscriptionCount(),
-		}).Debug(fmt.Sprintf("killing an available plugin in pool  %s:%s:%d", pType, pName, pVersion))
+		lp, err := r.pluginManager.get(fmt.Sprintf("%s"+core.Separator+"%s"+core.Separator+"%d", pType, pName, pVersion))
+		if lp != nil && lp.Details.Uri != nil {
+			runnerLog.WithFields(log.Fields{
+				"_block":     "handle-unsubscription",
+				"plugin-uri": lp.Details.Uri,
+			}).Debug(fmt.Sprintf("unsuscribe called on standalone plugin, moving on"))
```
rashmigottipati: typo: unsuscribe --> unsubscribe

control/plugin_manager.go
```diff
@@ -321,52 +329,75 @@ func (p *pluginManager) LoadPlugin(details *pluginDetails, emitter gomit.Emitter
...
+			})
+		}
+	} else {
 		pmLogger.WithFields(log.Fields{
 			"_block": "load-plugin",
-			"error":  err.Error(),
-		}).Error("load plugin error when starting plugin")
-		return nil, serror.New(err)
-	}
+			"path":   lPlugin.Details.Uri.String(),
```
rashmigottipati: Consider renaming "path" to ""plugin-uri"

control/runner.go
```diff
@@ -368,12 +368,26 @@ func (r *runner) handleUnsubscription(pType, pName string, pVersion int, taskID
 		return errors.New("pool not found")
 	}
 	if pool.SubscriptionCount() < pool.Count() {
-		runnerLog.WithFields(log.Fields{
-			"_block":                  "handle-unsubscription",
-			"pool-count":              pool.Count(),
-			"pool-subscription-count": pool.SubscriptionCount(),
-		}).Debug(fmt.Sprintf("killing an available plugin in pool  %s:%s:%d", pType, pName, pVersion))
+		lp, err := r.pluginManager.get(fmt.Sprintf("%s"+core.Separator+"%s"+core.Separator+"%d", pType, pName, pVersion))
+		if lp != nil && lp.Details.Uri != nil {
```
rashmigottipati: 👍

control/subscription_group.go
```diff
@@ -405,23 +409,53 @@ func (s *subscriptionGroup) subscribePlugins(id string,
...
 			}
-			err = s.pluginRunner.runPlugin(plg.Name(), plg.Details)
+		} else {
+			pool, err := s.pluginRunner.AvailablePlugins().getOrCreatePool(plg.Key())
 			if err != nil {
 				serrs = append(serrs, serror.New(err))
 				return serrs
 			}
+			pool.Subscribe(id)
+			if pool.Eligible() && plg.Details.Uri == nil {
```
rashmigottipati: Checking for `plg.Details.Uri == nil` isn't required here as the else case will be hit only if the Uri is nil.

control/subscription_group.go
```diff
@@ -405,23 +409,53 @@ func (s *subscriptionGroup) subscribePlugins(id string,
...
+					serrs = append(serrs, serror.New(err))
+					return serrs
+				}
+
+				body, err := ioutil.ReadAll(res.Body)
+				if err != nil {
+					serrs = append(serrs, serror.New(err))
+					return serrs
+				}
+				json.Unmarshal(body, &resp)
```
rashmigottipati: Check for error here and log it as Fatal.

control/subscription_group.go
```diff
@@ -405,23 +409,53 @@ func (s *subscriptionGroup) subscribePlugins(id string,
...
-		if err != nil {
-			serrs = append(serrs, serror.New(err))
-			return serrs
-		}
-		pool.Subscribe(id)
-		if pool.Eligible() {
-			err = s.verifyPlugin(plg)
+		if plg.Details.Uri != nil {
+			// this is a remote plugin
+			pool, err := s.pluginRunner.AvailablePlugins().getOrCreatePool(plg.Key())
```
rashmigottipati: Checking of this err is moved to bottom. Can you move it back to below this line?

control/subscription_group.go
```diff
@@ -405,23 +409,53 @@ func (s *subscriptionGroup) subscribePlugins(id string,
...
+			if pool.Count() < 1 {
```
rashmigottipati: Could you please explain what this implies?

control/subscription_group.go
```diff
@@ -405,23 +409,53 @@ func (s *subscriptionGroup) subscribePlugins(id string,
...
+					serrs = append(serrs, serror.New(err))
+					return serrs
+				}
+				json.Unmarshal(body, &resp)
+				ap, err := newAvailablePlugin(resp, s.eventManager, nil)
+				if err != nil {
+					serrs = append(serrs, serror.New(err))
+					return serrs
+				}
+				pool.Insert(ap)
```
rashmigottipati: In this if block, the pool count is less than 1, which I think implies that there are no available plugins in the pool which is why you're inserting an available plugin into the pool. 
Won't we also have subscribe the plugin by calling `pool.Subscribe(id)` ?

core/plugin.go
```diff
@@ -122,15 +126,31 @@ type SubscribedPlugin interface {
 }
 
 type RequestedPlugin struct {
-	path      string
-	checkSum  [sha256.Size]byte
-	signature []byte
+	path       string
+	checkSum   [sha256.Size]byte
+	signature  []byte
+	autoLoaded bool
```
rashmigottipati: Remove the field `autoLoaded`

core/plugin.go
```diff
@@ -122,15 +126,31 @@ type SubscribedPlugin interface {
...
+	return true
 }
 
 // NewRequestedPlugin returns a Requested Plugin which represents the plugin path and signature
 // It takes the full path of the plugin (path), temp path (fileName), and content of the file (b) and returns a requested plugin and error
 // The argument b (content of the file) can be nil
 func NewRequestedPlugin(path, fileName string, b []byte) (*RequestedPlugin, error) {
+	// Checks if string is URL
+	if isURL(path) {
+		if uri, err := url.ParseRequestURI(path); err == nil && uri != nil {
```
rashmigottipati: What happens when is there is a parse error? It won't return and it might cause a panic.

mgmt/rest/v1/plugin.go
aarontay: Added error handling to match source format: 
```
if err != nil { 
    rbody.Write(500, rbody.FromError(err), w)
}
```

control/plugin_manager.go
aarontay: removed

control/control.go
aarontay: removed

control/control.go
aarontay: Added check to verify for invalid URL

control/plugin_manager.go
aarontay: Added error check to match source formatting

control/runner.go
aarontay: fixed

control/plugin_manager.go
aarontay: fixed

control/control.go
```diff
@@ -705,7 +711,22 @@ func (p *pluginControl) UnsubscribeDeps(id string) []serror.SnapError {
 	return p.subscriptionGroups.Remove(id)
 }
 
+// Checks if string is URL
+func isURL(url string) bool {
```
mkleina: Consider moving this function body to single source file (for example `/grpc/common/common.go`) instead of redeclaring it multiple times

core/plugin.go
```diff
@@ -125,12 +129,27 @@ type RequestedPlugin struct {
 	path      string
 	checkSum  [sha256.Size]byte
 	signature []byte
+	uri       *url.URL
+}
+
+// Checks if string is URL
+func isURL(url string) bool {
```
mkleina: Consider moving this function body to single source file (for example `/grpc/common/common.go`) instead of redeclaring it multiple times

mgmt/rest/client/client.go
```diff
@@ -86,6 +88,14 @@ func parseURL(url string) error {
 	return nil
 }
 
+// Checks if string is URL
+func isURL(url string) bool {
```
mkleina: Consider moving this function body to single source file (for example `/grpc/common/common.go`) instead of redeclaring it multiple times

control/plugin_manager.go
```diff
@@ -440,7 +440,7 @@ func (p *pluginManager) LoadPlugin(details *pluginDetails, emitter gomit.Emitter
 	if err != nil {
 		pmLogger.WithFields(log.Fields{
 			"_block":         "load-plugin",
-			"plugin-type":    "collector",
```
IzabellaRaulin: good catch! 👍

PR #1619, opened 2017-04-28
Fix plugin load timeout in LoadPlugin and startPlugin 

### LINKED ISSUE #1618: Fix plugin load timeout in LoadPlugin and startPlugin
Currently, the global option PluginLoadTimeout is implemented such that we wait till we receive the plugin's response (which is the preamble) or we timeout. There could be cases where we are able to get to the point where we receive the plugin response but the plugin load is still unsuccessful. This requires us to change the implementation of timeout such that we wait until we receive a loaded plugin in plugin_manager.go. So, we need to block until receiving a loaded plugin or timeout expiry. 
Similarly, in startPlugin in runner.go, we need to wait until receiving a channel's response which either has an available plugin or error, and then timeout. 

Also, PluginLoadTimeout is hardcoded in control/runner.go. This makes it inconsistent with other code paths as the timeout value is being read from config in those. We need to make it more consistent and have all the code paths read the timeout value from the config as opposed to hardcoding the value.

rashmigottipati:
Fixes #1618

Summary of changes:
- Fix plugin load timeout in LoadPlugin and startPlugin by blocking until receiving a loaded plugin or timeout expiry. Currently, on master, we execute the plugin and wait for the plugin response or timeout waiting for the plugin's response. In this PR, we handle timeout in LoadPlugin in plugin_manager.go such that: we wait until receiving a channel's response which either has a loaded plugin or error, and then timeout. In control/runner.go, we wait until the channel sends an available plugin or an error and then timeout if we don't receive those. 

- Update runner.go to get the plugin load timeout value from config as opposed to a hardcoded value to make it consistent with other code paths.

Testing:
- Legacy test
- Manual testing

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/control.go
+++ b/control/control.go
@@ -116,6 +116,7 @@ type runsPlugins interface {
 	SetPluginManager(managesPlugins)
 	Monitor() *monitor
 	runPlugin(string, *pluginDetails) error
+	SetPluginLoadTimeout(int)
 }
 
 type managesPlugins interface {
@@ -174,6 +175,7 @@ func OptSetConfig(cfg *Config) PluginControlOpt {
 		c.Config = cfg
 		c.pluginManager.SetPluginConfig(cfg.Plugins)
 		c.pluginManager.SetPluginLoadTimeout(c.Config.PluginLoadTimeout)
+		c.pluginRunner.SetPluginLoadTimeout(c.Config.PluginLoadTimeout)
 	}
 }
 
--- a/control/plugin_manager_test.go
+++ b/control/plugin_manager_test.go
@@ -67,10 +67,16 @@ func TestLoadedPlugins(t *testing.T) {
 	})
 }
 
-func loadPlugin(p *pluginManager, fileName string) (*loadedPlugin, serror.SnapError) {
+func loadPlugin(p *pluginManager, fileName string, retries ...int) (*loadedPlugin, serror.SnapError) {
 	// This is a Travis optimized loading of plugins. From time to time, tests will error in Travis
 	// due to a timeout when waiting for a response from a plugin. We are going to attempt loading a plugin
 	// 3 times before letting the error through. Hopefully this cuts down on the number of Travis failures
+	var retryCount int
+	if len(retries) == 0 {
+		retryCount = 3
+	} else {
+		retryCount = retries[0]
+	}
 	var e serror.SnapError
 	var lp *loadedPlugin
 	file, err := os.Open(fileName)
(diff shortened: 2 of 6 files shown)


kjlyon:
LGTM

control/runner_test.go
```diff
@@ -473,6 +473,18 @@ func TestRunnerPluginRunning(t *testing.T) {
...
+
+					Convey("execution time of startPlugin should exceed timeout value", func() {
+						r := newRunner()
+						r.SetEmitter(new(MockEmitter))
+						r.SetPluginLoadTimeout(10)
+						exPlugin := new(MockExecutablePlugin)
+						startTime := time.Now()
+						ap, _ := r.startPlugin(exPlugin)
+						So(ap, ShouldBeNil)
+						So(time.Since(startTime), ShouldBeGreaterThan, r.pluginLoadTimeout)
```
candysmurf: the assertion seems meaningless as the 28.34 microsecond is less than 10s but the comparison returns true.
Should you test the scenario of timeout?

control/control.go
```diff
@@ -174,6 +175,7 @@ func OptSetConfig(cfg *Config) PluginControlOpt {
 		c.Config = cfg
 		c.pluginManager.SetPluginConfig(cfg.Plugins)
 		c.pluginManager.SetPluginLoadTimeout(c.Config.PluginLoadTimeout)
+		c.pluginRunner.SetPluginLoadTimeout(c.Config.PluginLoadTimeout)
```
rashmigottipati: To be able to access PluginLoadTimeout and set it in runner.go, I would need to add the method SetPluginLoadTimeout to the pluginRunner, similar to pluginManager. Currently, on master, the timeout value is hardcoded in runner.go - https://github.com/intelsdi-x/snap/blob/master/control/runner.go#L183
This PR fixes that by reading it from config, so that it's consistent across LoadPlugin in plugin_manager.go and startPlugin in runner.go.

control/plugin_manager_test.go
```diff
@@ -102,28 +108,25 @@ func loadPlugin(p *pluginManager, fileName string) (*loadedPlugin, serror.SnapEr
...
 		lp, e = p.LoadPlugin(details, nil)
 		if e == nil {
-			break
-		}
-		if e != nil && i == 2 {
-			return nil, e
-
+			break //success
 		}
+		// if e != nil && i == 2 {
```
jcooklin: Let's remove this commented code.

PR #1630, opened 2017-05-10
Update LICENSE

skonefal:
Included dates and copyright owner.

Summary of changes:
- Included date
- Included copyright owner
- Newline in the end

@intelsdi-x/snap-maintainers

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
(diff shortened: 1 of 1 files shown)


andrzej-k:
Hey @skonefal - Thanks for this PR, but I don't think it's necessary. The section you have updated is saying what should be the format of the license header that should be included in the source files. 
And we are following this, for example, take a look here: https://github.com/intelsdi-x/snap/blob/master/snapteld.go#L5

Looking for your feedback...

Issue #1631, opened 2017-05-10
StreamCollector - active after task stop or remove


obno:
There seems to be no way to notify a StreamCollector plugin (once started) when (all?) tasks that were referring to its metrics have been stopped or removed. This means that the collection process will be still active event though no more tasks require its metrics.       

To demonstrate this, take the **test-rand-streamer** from the snap-plugin-lib-go examples and add a debug message inside the `func streamIt` collection for loop. Then load it to snapteld and create a task for it:

````
$snaptel plugin load plugins/test-rand-streamer
Plugin loaded
Name: test-rand-streamer
Version: 1
Type: collector
Signed: false
Loaded Time: Wed, 10 May 2017 19:12:25 EEST    

$snaptel plugin list
NAME                     VERSION         TYPE            SIGNED          STATUS          LOADED TIME
test-rand-streamer       1               collector       false           loaded          Wed, 10 May 2017 19:12:25 EEST
file                     2               publisher       false           loaded          Wed, 10 May 2017 19:12:38 EEST

$curl -vX POST localhost:8181/v1/tasks -d @tasks/test-rand-streamer.json --header "Content-Type: application/json"
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8181 (#0)
> POST /v1/tasks HTTP/1.1
> Host: localhost:8181
> User-Agent: curl/7.47.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 426
>
* upload completely sent off: 426 out of 426 bytes

< HTTP/1.1 201 Created
< Deprecated: true
< Date: Wed, 10 May 2017 16:13:02 GMT
< Content-Length: 1045
< Content-Type: text/plain; charset=utf-8
<
{
  "meta": {
    "code": 201,
    "message": "Scheduled task created (f99b9bed-aa24-44a8-ac87-2b11d9f95e4c)",
    "type": "scheduled_task_created",
    "version": 1
  },
  "body": {
    "id": "f99b9bed-aa24-44a8-ac87-2b11d9f95e4c",
    "name": "Task-f99b9bed-aa24-44a8-ac87-2b11d9f95e4c",
    "deadline": "5s",
    "workflow": {
      "collect": {
        "metrics": {
          "/random/float": {
            "version": 0
          },
          "/random/integer": {
            "version": 0
          },
          "/random/string": {
            "version": 0
          }
        },
        "publish": [
          {
[...]

Issue #1587, opened 2017-04-04
Plugin Wanted: Graphite Streaming Collector


jtlisi:
With the recent creation of streaming plugins, one plugin I had in mind was a graphite streaming plugin.

The plugin will accept metrics in the graphite metric format over the plaintext or pickle protocol. 
> *Plaintext*: `<metric path> <metric value> <metric timestamp>`

> *Pickle*: `[(path, (timestamp, value)), ...]`

This collector would utilize a model similar to the SNMP collector by updating the metric catalog using a set-file. It would translate metrics into the snap data model as follows.

metric: `http_requests.count.status.200.index 105 1491304254`

rule:
```yaml
rules:
  - expression: 'http_request.count.status.*.*'
     namespace:
       - http_request
       - count
       - 1                       # Translates the first wildcard element
         name: 'status_code'
         description: 'http status code'
       - 2                       # Translates the second wildcard element
         name: 'endpoint'
         description: 'url endpoint'
       - status
    tags: 
      example_tag_wildcard: 1
      example_tag_string: 'tag'
    description: 'description of this metric'
    unit: 'unit of this metric'
    config:
      etc...
```

I think a collector like this would be very useful and allow snap to integrate more tightly with the current graphite streaming ecosystem. The use case I have in mind is configuring [springboot](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html) to stream metrics directly to snap. Let me know what you guys think and if you have any suggestions or better ways to do this.

Thanks!

jcooklin:
Hi Jacob, I'm working on a standalone relay that will stream metrics. It's
not yet ready to be made public but I'll add you as a collaborator so you
can provide feedback.

jtlisi:
Cool, I'd love to test it!

PR #1632, opened 2017-05-11
Adds support for streaming collectors

jcooklin:
img
The docker-compose file and task used in this demo can be found [here](https://gist.github.com/jcooklin/5d81a20b852dec2c2e9250030bf7ba68).  

Related:
* PRs
  * https://github.com/intelsdi-x/snap-relay/pull/3
  * https://github.com/intelsdi-x/snap-plugin-lib-go/pull/85
* Issues
  * https://github.com/intelsdi-x/snap/issues/1631 - StreamCollector - active after task stop or remove
  * https://github.com/intelsdi-x/snap/issues/1587 - Plugin Wanted: Graphite Streaming Collector

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/plugin/plugin.go
+++ b/control/plugin/plugin.go
@@ -91,7 +91,7 @@ var (
 		"collector",
 		"processor",
 		"publisher",
-		"streamCollector",
+		"streaming-collector",
 	}
 
 	routingStrategyTypes = [...]string{
--- a/control/plugin_manager.go
+++ b/control/plugin_manager.go
@@ -491,7 +491,7 @@ func (p *pluginManager) LoadPlugin(details *pluginDetails, emitter gomit.Emitter
 		lPlugin.LoadedTime = time.Now()
 		lPlugin.State = LoadedState
 
-		if resp.Type == plugin.CollectorPluginType {
+		if resp.Type == plugin.CollectorPluginType || resp.Type == plugin.StreamCollectorPluginType {
 			cfgNode := p.pluginConfig.getPluginConfigDataNode(core.PluginType(resp.Type), resp.Meta.Name, resp.Meta.Version)
 
 			if lPlugin.ConfigPolicy != nil {
@@ -543,9 +543,11 @@ func (p *pluginManager) LoadPlugin(details *pluginDetails, emitter gomit.Emitter
 			metricTypes, err := colClient.GetMetricTypes(cfg)
 			if err != nil {
 				pmLogger.WithFields(log.Fields{
-					"_block":      "load-plugin",
-					"plugin-type": "collector",
-					"error":       err.Error(),
+					"_block":         "load-plugin",
+					"plugin-type":    resp.Type.String(),
+					"error":          err.Error(),
+					"plugin-name":    ap.Name(),
+					"plugin-version": ap.Version(),
 				}).Error("error in getting metric types")
 				resultChan <- result{nil, serror.New(err)}
 				return
@@ -714,7 +716,7 @@ func (p *pluginManager) UnloadPlugin(pl core.Plugin) (*loadedPlugin, serror.Snap
 	p.loadedPlugins.remove(plugin.Key())
 
 	// Remove any metrics from the catalog if this was a collector
-	if plugin.TypeName() == "collector" {
+	if plugin.TypeName() == core.CollectorPluginType.String() || plugin.TypeName() == core.StreamingCollectorPluginType.String() {
 		p.metricCatalog.RmUnloadedPluginMetrics(plugin)
 	}
 
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -35,6 +35,7 @@ import (
 
 	"github.com/intelsdi-x/gomit"
 
+	"github.com/intelsdi-x/snap/control/plugin"
 	"github.com/intelsdi-x/snap/core"
 	"github.com/intelsdi-x/snap/core/cdata"
 	"github.com/intelsdi-x/snap/core/ctypes"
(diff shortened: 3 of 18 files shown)


katarzyna-z:
During tests of this PR please remember to take into account scenario described in #1650, for now snapteld crashes when task is created with invalid task manifest.

jcooklin:
@katarzyna-z, @ericho  e89cc4e fixes #1650.

Given the following task will now error on create where before it would have panic'd the framework.
```
---
  version: 1
  schedule:
    type: "simple"
    interval: "5s"
  workflow:
    collect:
      metrics:
        /relay/collectd: {}
      publish:
        -
            plugin_name: "influxdb"
            config:
                host: influxdb
                port: 8086
                database: "snap"
                user: "admin"
                password: "admin"
```

Note, the metric `/collectd/relay` is retrieved from a streaming plugin.  

Output:
```
bash-4.3# snaptel task create -t /tmp/collectd.yml
Error creating task:Plugin is incompatible with the tasks schedule type.
```

CC: @rashmigottipati

scheduler/scheduler.go
```diff
@@ -337,17 +342,73 @@ func (s *scheduler) createTask(sch schedule.Schedule, wfMap *wmap.WorkflowMap, s
...
+						)
+					}
+				}
+				return nil
+			})
+			// assert only a single streaming plugin
+			subscribedPluginAsserts = append(subscribedPluginAsserts, func(plugins []core.SubscribedPlugin) serror.SnapError {
+				if len(plugins) > 1 {
+					return serror.New(
+						ErrPluginIncompatibleWithScheduleType,
```
rashmigottipati: Consider changing the error type here to ErrMultipleStreamingPlugins.

control/plugin_manager.go
```diff
@@ -543,9 +543,12 @@ func (p *pluginManager) LoadPlugin(details *pluginDetails, emitter gomit.Emitter
...
 				pmLogger.WithFields(log.Fields{
-					"_block":      "load-plugin",
-					"plugin-type": "collector",
-					"error":       err.Error(),
+					"_block":         "load-plugin",
+					"plugin-type":    resp.Type.String(),
+					"error":          err.Error(),
+					"plugin-name":    ap.Name(),
+					"plugin-version": ap.Version(),
+					"plugin-id":      ap.ID(),
```
IzabellaRaulin: At this moment plugin-id is not generated yet, so api.ID() returns zero. Based on that this field might be removed from the log because does not present any valuable info.

scheduler/scheduler.go
```diff
@@ -337,17 +342,73 @@ func (s *scheduler) createTask(sch schedule.Schedule, wfMap *wmap.WorkflowMap, s
...
+			// assert no streaming plugins
+			subscribedPluginAsserts = append(subscribedPluginAsserts, func(plugins []core.SubscribedPlugin) serror.SnapError {
+				for _, plg := range plugins {
+					if plg.TypeName() == plugin.StreamCollectorPluginType.String() {
+						return serror.New(
+							ErrPluginIncompatibleWithScheduleType,
+							map[string]interface{}{
+								"schedule_type": fmt.Sprintf("%T", sch),
+								"plugin_name":   plg.Name(),
+								"plugin_type":   plg.TypeName(),
```
IzabellaRaulin: I did not receive those "schedule_type", "plugin_name", etc. in error message. 
What I received for test-scenario of using "simple" schedule for metrics coming from stream collector plugin is: 
```
$ snaptel task create -t simple.json
Error creating task:Plugin is incompatible with the tasks schedule type.
```
```
$ curl -vX POST localhost:8181/v1/tasks -d @simple.json --header "Content-Type: application/json"
{
  "meta": {
    "code": 500,
    "message": "Plugin is incompatible with the tasks schedule type.",
    "type": "error",
    "version": 1
  },
  "body": {
    "message": "Plugin is incompatible with the tasks schedule type.",
    "fields": {}
  }
* Connection #0 to host localhost left intact
}
```

jcooklin: That is an issue but I feel it's outside the scope of this PR.   If we add `log.Debugf("!!! we are losing these fields %+v", e.Fields())` [here](https://github.com/intelsdi-x/snap/blob/master/core/task.go#L310) you will see something like the following in snapteld's log which is what we would like to see.  

img

In the end I think [CreateTaskFromContent](https://github.com/intelsdi-x/snap/blob/master/core/task.go#L247) needs to return a collection of `serrors` and not just an `error`.

IzabellaRaulin: I fully agree, this is the issue which should be addressed outside the scope of this PR. I  created GH issue to address that in this way - https://github.com/intelsdi-x/snap/issues/1654

control/plugin/plugin.go
```diff
@@ -91,7 +91,7 @@ var (
 		"collector",
 		"processor",
 		"publisher",
-		"streamCollector",
+		"streaming collector",
```
IzabellaRaulin: I suggest changing type name to "streaming-collector" instead of having space there. The reason is to avoid having the skip space for example in command `snaptel plugin unload <plugin_type> <plugin_name> <plugin_version>`

kjlyon:
Assuming the functionality is there and works with the other PRs from snap-relay and plugin-lib-go, this code LGTM.

IzabellaRaulin:
@kjlyon, @jcooklin - what about changing `streaming collector` to `streaming-collector`, so user can straighforward call 
`snaptel plugin unload streaming-collector snap-relay 1`
instead of
```snaptel plugin unload "streaming collector" snap-relay 1```

Let me know, if you think is worth changing or not.

File: control/plugin_manager.go [hunk 5]
```diff
@@ -714,7 +716,7 @@ func (p *pluginManager) UnloadPlugin(pl core.Plugin) (*loadedPlugin, serror.Snap
 	p.loadedPlugins.remove(plugin.Key())
 
 	// Remove any metrics from the catalog if this was a collector
-	if plugin.TypeName() == "collector" {
+	if plugin.TypeName() == "collector" || plugin.TypeName() == core.StreamingCollectorPluginType.String() {
```
IzabellaRaulin: I would like to keep the same pattern for "collector" as well and change it to
```
if plugin.TypeName() == core.CollectorPluginType.String() || plugin.TypeName() == core.StreamingCollectorPluginType.String() {
```

but it's a minor thing - @jcooklin, I will prepare PR with that to your repo - accept it or reject, up to you

control/plugin/plugin.go
IzabellaRaulin: @jcooklin, I prepared PR with that change: -> https://github.com/jcooklin/snap/pull/2 - use it or not, up to you

IzabellaRaulin:
@Joel, I found out that errors are not reflected in Fails count for the task with streaming-collector. But I think it will be better to address that as a separate issue. I will open an issue with details how to reproduce it.

PR #1634, opened 2017-05-15
Correcting sub-bullet spacing

mbbroberg:
Current README does not follow CommonMark standard that [GitHub recently began to use](https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown), which is why it renders incorrectly. Correcting that and sending the team an FYI on the change.

Before: 
screen shot 2017-05-15 at 10 27 11 am

After: 
screen shot 2017-05-15 at 10 27 19 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,19 +34,19 @@ limitations under the License.
 
 1. [Overview](#overview)
 2. [Getting Started](#getting-started)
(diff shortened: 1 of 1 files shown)


rashmigottipati:
LGTM

Issue #1637, opened 2017-05-17
Deprecating client API v1


aarontay:
Client API v1 is being deprecated. Update to use [API v2](https://github.com/intelsdi-x/snap/tree/master/mgmt/rest/v2). See additional documentation on the client here (pending update to v2): https://github.com/intelsdi-x/snap/blob/master/docs/REST_API.md

andrzej-k:
Hey @aarontay - For how long do you plan to keep API v1 available to the users? Are there plans (dates) for removing API v1 from the code base?

woodsaj:
https://github.com/intelsdi-x/snap-client-go just 404's for me.  Do you need to make that repo public?

aarontay:
@woodsaj the snap-client-go repo being made pubic is part of what the deprecations is pending on. @candysmurf is there an estimate date set for the client?

PR #1638, opened 2017-05-17
Added deprecation tags to v1 functions; added deprecation comments

aarontay:
Fixes #

Summary of changes:
- Added support to log_handle to check if client has been deprecated
- Added comments to v1 source to indicate deprecated API with link to Github issue and v2 repo
- Created issue [1637](https://github.com/intelsdi-x/snap/issues/1637) to address deprecated API additional information

Testing done:
- Verified that snapteld printed the deprecated API info to the debug log 

@intelsdi-x/snap-maintainers

### DIFF
--- a/mgmt/rest/log_handler.go
+++ b/mgmt/rest/log_handler.go
@@ -52,4 +52,12 @@ func (l *Logger) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.Ha
 		"status-code": res.Status(),
 		"status":      http.StatusText(res.Status()),
 	}).Debug("API response")
+
+	if deprecationInfo := rw.Header().Get("Deprecated"); len(deprecationInfo) != 0 {
+		restLogger.WithFields(log.Fields{
+			"method": r.Method,
+			"url":    r.URL.Path,
+		}).Warning(deprecationInfo)
+	}
+
 }
--- a/mgmt/rest/v1/rbody/body.go
+++ b/mgmt/rest/v1/rbody/body.go
@@ -32,7 +32,7 @@ import (
 	"github.com/urfave/negroni"
 )
 
-const deprecationInfo = "The client API V1 has been depricated. Find more information here: https://github.com/intelsdi-x/snap/issues/1637"
+const deprecationInfo = "The client API V1 is being deprecated; switch to API v2. Find more information here: https://github.com/intelsdi-x/snap/issues/1637"
 
 type Body interface {
 	// These function names are rather verbose to avoid field vs function namespace collisions
--- a/mgmt/rest/v1/rbody/body.go
+++ b/mgmt/rest/v1/rbody/body.go
@@ -32,6 +32,8 @@ import (
 	"github.com/urfave/negroni"
 )
 
+const deprecationInfo = "The client API V1 has been depricated. Find more information here: https://github.com/intelsdi-x/snap/issues/1637"
+
 type Body interface {
 	// These function names are rather verbose to avoid field vs function namespace collisions
 	// with varied object types that use them.
@@ -40,7 +42,8 @@ type Body interface {
 }
 
 func Write(code int, b Body, w http.ResponseWriter) {
-	w.Header().Set("Deprecated", "true")
+	w.Header().Set("Deprecated", deprecationInfo)
+
 	resp := &APIResponse{
 		Meta: &APIResponseMeta{
 			Code:    code,
--- a/mgmt/rest/v1/tribe.go
+++ b/mgmt/rest/v1/tribe.go
@@ -42,12 +42,16 @@ var (
 	ErrMemberNotFound        = errors.New("Member not found")
 )
 
+// Deprecated: Update to apiV2(https://github.com/intelsdi-x/snap-client-go).
+// Find more information here: https://github.com/intelsdi-x/snap/issues/1637
 func (s *apiV1) getAgreements(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 	res := &rbody.TribeListAgreement{}
 	res.Agreements = s.tribeManager.GetAgreements()
 	rbody.Write(200, res, w)
 }
 
+// Deprecated: Update to apiV2(https://github.com/intelsdi-x/snap-client-go).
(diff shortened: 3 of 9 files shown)


candysmurf:
@aarontay, please rebase your PR and resolve conflicts.

mgmt/rest/log_handler.go
```diff
@@ -50,6 +52,17 @@ func (l *Logger) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.Ha
 		"method":      r.Method,
 		"url":         r.URL.Path,
 		"status-code": res.Status(),
-		"status":      http.StatusText(res.Status()),
+		"status":      deprecated(http.StatusText(res.Status()), rw.Header().Get("Deprecated"), rw.Header().Get("DeprWhat"), rw.Header().Get("DeprInfo")),
```
IzabellaRaulin: @aarontay,  what do you think about:
 - logging about apiV1 depreciation as a warning, not debug
- do it a separate log  (do not combine api response + deprecation info in one line)

**Why**: All of this to increase readability + keep the same pattern which has been used in RPC deprecation.

#### How it looks currently:
   For example: `$snaptel plugin load <plugin_path>`
    
   image

#### Suggestion:
   Separate log for deprecation info on warning level - done in a similar way as for deprecated RPC protocol, see below:
image

mgmt/rest/log_handler.go
```diff
@@ -50,6 +52,17 @@ func (l *Logger) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.Ha
...
+// 'deprecate' is string boolean value to deprecate, mapped to 'Deprecated' key
+// 'deprWhat' is what is being deprecated, mapped to 'DeprWhat' key
+// 'deprInfo' is link to more information on what is being deprecated, mapped to 'DeprInfo' key
+func deprecated(status string, deprecate string, deprWhat string, deprInfo string) string {
+	dep, err := strconv.ParseBool(deprecate)
+	if err != nil || !dep {
+		return status
+	}
+	return status + ". '" + deprWhat + "' has been depricated. Find more information here: " + deprInfo
+}
```
IzabellaRaulin: I have a doubt if this is really need to be done in this way. What do you think about making it more straightforward, e.g.: 
```
       if deprecationInfo := rw.Header().Get("Deprecated"); len(deprecationInfo) != 0 {
                restLogger.WithFields(log.Fields{
                        "method": r.Method,
                        "url":    r.URL.Path,
                }).Warning(deprecationInfo)
        }
```

where in [mgmt/rest/v1/rbody/body.go](https://github.com/intelsdi-x/snap/pull/1638/files#diff-d53afab335763da0de3e3036f60107faR43) const might determined the string value which is set in Write():
```
const deprecationInfo = "API V1 has been depricated. Find more information here: https://github.com/intelsdi-x/snap/issues/1637"

func Write(code int, b Body, w http.ResponseWriter) {
        w.Header().Set("Deprecated", deprecationInfo)
[...]
}
```

mgmt/rest/v1/tribe.go
```diff
@@ -42,12 +42,16 @@ var (
 	ErrMemberNotFound        = errors.New("Member not found")
 )
 
+// Deprecated: Update to apiV2(https://github.com/intelsdi-x/snap-client-go).
```
candysmurf: @aarontay, I don't think tribe has v2. Do we want to deprecate tribe?

aarontay: I added deprecation comments to any functions that incorporate apiV1... This is part of the rest-tribe module located under the rest/v1 src directory; wouldn't we want to deprecate this specific module based on its dependency to Rest API v1?

aarontay: Good catch; I added a notice to the REST_API.md doc. Thanks!

IzabellaRaulin: There is no Tribe routes in API V2, but before removing API V1, Tribe should be supported in API V2 too.

aarontay: @IzabellaRaulin done!

mgmt/rest/v1/rbody/body.go
```diff
@@ -32,6 +32,8 @@ import (
 	"github.com/urfave/negroni"
 )
 
+const deprecationInfo = "The client API V1 has been depricated. Find more information here: https://github.com/intelsdi-x/snap/issues/1637"
```
andrzej-k: To be more useful to the user the message should something like this: "The client API v1 has been deprecated, and will be removed <provide date or release here>. Please switch to API v2."

Is API v2 ready to be exposed to end users?

aarontay: Okay, I updated the message, but I did not provide a date... The deprecation is pending on snap-client-go to be made public and API v2 tribe support, so I am just leaving the link to the issue that provides user details and the message can be updated when a specific date is actually known.

mgmt/rest/v1/tribe.go
aarontay: @candysmurf I previously added the notification to REST_API.md and mentioned in the issue it is pending the deprecation announcement; any other requests?

mgmt/rest/v1/rbody/body.go
```diff
@@ -32,6 +32,8 @@ import (
 	"github.com/urfave/negroni"
 )
 
+const deprecationInfo = "The client API V1 is being deprecated; switch to API v2. Find more information here: https://github.com/intelsdi-x/snap/issues/1637"
```
andrzej-k: I feel a bit confused. Issue #1637 suggests that https://github.com/intelsdi-x/snap-client-go is the new API v2. Regardless of the fact that it is still private and not accessible to end users, it doesn't seem to contain API v2 but the client built on top of this API. API v2 seems to be defined here: https://github.com/intelsdi-x/snap/tree/master/mgmt/rest/v2. Is that correct?

IzabellaRaulin:
This pull request is quite old and pending on snap-client-go to be made public and API v2 tribe support. Based on that, it seems to be better close this PR and come back to the discussion once again.

Issue #1421, opened 2016-12-06
Example config files errors


IRCody:
```
% snapteld -t 0 -l 1 --config examples/configs/snap-config-sample.yaml                                                                                                                                                
FATA[0000] ConfigFile Error: Port can not be specified both as a port value and as part of address ```

When trying to use any of the example config file I get the this error.

One consideration with changing them is that they are used in tests so it might be a little more involved to fix than it seems at first glance.

**Comment by nanliu** [comment]
Yeah, they cram every option possible, maybe it makes sense to comment out conflicting options so they are at least working configs?

**Comment by kindermoumoute** [comment]
The thing is those config file are also used in tests, so if we comment out something in those files, test fail. Removing those test is not a solution either, some possibly way on the top of my head are:
- Include the mock file `examples/configs/snap-config-sample.yaml` in a Go file `simple_config_sample_test.go`.
- Remove inconsistencies in config examples, but create new config files (if needed) to keep every use case, and update code to make it test right config files.

**Comment by IzabellaRaulin** [comment]
I cannot reproduce exactly the issue described in #1421 to get the error message "Port can not be specified both as a port value and as part of address". However, those files are not ready to use by users as it is - users need to make some adjustment/customization inside, for example, comment out unneeded config params or set to them own values.

**Fortunately**, a user who uses one of the exemplary config files for the first time will receive such error:
```
$ snapteld --config examples/configs/snap-config-sample.json
FATA[0000] stat /var/log/snap: no such file or directory
```
```
$ snapteld --config examples/configs/snap-config-sample.yaml
FATA[0000] stat /var/log/snap: no such file or directory
```
what suggests that values of config params need to be customized and adjusted to user use-case. **What is more, YAML file contains "NOTICE" to inform about that in a clear way.**

Providing reference config file is also an option which might be taken into consideration, but I suppose some difficulties might be found in defining what "the reference" really is. As far I know,  the most common case is running snapteld without global config file and basing just on defaults. Even some specific collector plugins which need global config to be able to load usually contain dedicated config file in their plugin repo, e.g. [snap-plugin-collector-mysql](https://github.com/intelsdi-x/snap-plugin-collector-mysql/tree/master/examples). Based on that, in my opinion, there are no real needs of providing reference config file at the moment and this bug might be treat as resolved by PR https://github.com/intelsdi-x/snap/pull/1644

cc: @intelsdi-x/snap-maintainers

katarzyna-z:
[...]

PR #1644, opened 2017-05-18
Updated snap-config-sample files

IzabellaRaulin:
Related to #1421 

### Summary of changes:
-   added "Notice" to snap-config-sample.yaml
 - uniformity provided config value (see log_path is now the same in both files)
- renaming `snapclt-config-sample.json` to `snaptel-config-sample.json`

### Testing done:
-  small, medium, legacy

@intelsdi-x/snap-maintainers


katarzyna-z:
It looks good but I've written a one idea under the issue #1421 ;)

IzabellaRaulin:
@katarzyna-z - I think your idea should be taken into consideration. I will prepare such config file with default values and we will see how it will look and if it provides benefits for users, then it definitely should land as an exemplary config file.

katarzyna-z:
@IzabellaRaulin ok, LGTM

PR #1656, opened 2017-06-07
Update ConfigValue type in grpc assert

rashmigottipati:
Summary of changes:

- Update configvalue to be a non-pointer type so that the assert doesn't break which enables the config being read and set correctly. 

@intelsdi-x/snap-maintainers


kjlyon:
LGTM

PR #1666, opened 2017-06-13
Updating parsed plugin list from plugins.yml

mbbroberg:
Adds RDT plugin to the catalog.


mbbroberg:
It's for the website, so YOLOing and verifying 😊

mbbroberg:
RDT now shows and the site still works 👍

Issue #1662, opened 2017-06-13
failed_count more then the hit_count ?


fengw084:
Running the task, and occasionally there will be several failed_count more then the hit_count, this is BUG or normal？

kjlyon:
Hi @fengw084 , thank you for bringing this up! I would like to dig deeper into this, do you mind sharing your task manifest with me to see if I can reproduce the error?

mbbroberg:
I also wonder if you're asking for clear definitions. I believe it goes that: 
* `hit_count` is the number of times a workflow ran successfully 
* `failed_count` is the number of times a workflow failed to run

Making it so that `hit_count` + `failed_count` = total attempts to run workflow. 

FWIW, I don't believe that this relationship is clear.

kjlyon:
Hi @fengw084, it's been a while since we have heard from you. I hope @mbbroberg and I were able to resolve your question. Please open another issue or contact us on [our Slack channel](https://intelsdi-x.herokuapp.com/) if you have more questions, we would be happy to help!

PR #1668, opened 2017-06-15
Clarifying hit_count is successful runs only

mbbroberg:
Minor clarification to help out cases like #1662.

### DIFF
--- a/docs/REST_API.md
+++ b/docs/REST_API.md
@@ -341,7 +341,7 @@ Snap task APIs provide the functionality to create, start, stop, remove, enable,
 | deadline                         | task timeout time                       |
 | creation_timestamp               | task creation time                      |
 | last_run_timestamp               | last running time of a task             |
-| hit_count                        | number of times a task ran              |
+| hit_count                        | number of times a task succeeded        |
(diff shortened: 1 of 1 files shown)


docs/REST_API.md
```diff
@@ -341,7 +341,7 @@ Snap task APIs provide the functionality to create, start, stop, remove, enable,
 | deadline                         | task timeout time                       |
 | creation_timestamp               | task creation time                      |
 | last_run_timestamp               | last running time of a task             |
-| hit_count                        | number of times a task ran              |
+| hit_count                        | number of times a task succeeded        |
```
rashmigottipati: Just a suggestion: Would it be better to say "number of times a workflow finished successfully"?

rashmigottipati: Either way, LGTM

IzabellaRaulin: Simpler is better, so it LGTM as it is. @rashmigottipati - are you ok with keeping this as it is? If yes, please merge it. 

@mbbroberg, thank You for your contribution.

mbbroberg:
I like it simpler too. Merging!

PR #1671, opened 2017-06-19
fixed #1669: Remove env var SNAP_REST_PASSWORD

### LINKED ISSUE #1669: Remove SNAP_REST_PASSWORD env var
When setting the env var `SNAP_REST_PASSWORD` in any terminal, then the basic authentication of snaptel does not work anymore. The error is

```
$ snaptel plugin list
could not parse snap as bool value for flag password, p: strconv.ParseBool: parsing "snap": invalid syntax

```

OR

```
$ snaptel -p plugin list
could not parse snap as bool value for flag password, p: strconv.ParseBool: parsing "snap": invalid syntax
```

candysmurf:
Fixes #1669 

Summary of changes:
- removed the env var SNAP_REST_PASSWORD

Testing done:
- all

@intelsdi-x/snap-maintainers

### DIFF
--- a/cmd/snaptel/flags.go
+++ b/cmd/snaptel/flags.go
@@ -50,9 +50,8 @@ var (
 		Usage: "Shows running plugins",
 	}
 	flPassword = cli.BoolFlag{
-		Name:   "password, p",
-		Usage:  "Require password for REST API authentication",
-		EnvVar: "SNAP_REST_PASSWORD",
+		Name:  "password, p",
+		Usage: "Require password for REST API authentication, e.g. snaptel -p plugin list",
 	}
 	flConfig = cli.StringFlag{
 		Name:   "config, c",


cmd/snaptel/flags.go
```diff
@@ -49,10 +49,11 @@ var (
 		Name:  "running",
 		Usage: "Shows running plugins",
 	}
-	flPassword = cli.BoolFlag{
```
IzabellaRaulin: In my point of view, the issue is #1669 it's not valid - trying to pass a string when bool is expected should return an error. Please, notice how the usage is formed: "Require password for REST API authentication". Set this to true cause that user will be asked to provide the password - [here](https://github.com/intelsdi-x/snap/pull/1671/files#diff-437ff6f8535e7c3766454ed3a7568823L129). I do not think that this mechanism should be changed at the moment and in this way. @candysmurf, could we come back and trigger discussion about another idea(s) and bringing benefits? If you agree, feel free to open a new issue to make a space for such discussion (but as an RFC, not as identified bug) with providing use cases. Thank You

candysmurf: @IzabellaRaulin, the reason changing to the string flag instead of bool has two benefits:
* Supports environmental variable. It does not work totally for the current implementation which uses `bool`
* Original implementation was not user-friendly. When you specify a flag `-p`, intuition is to follow a password. It does not work by using `bool`. It's quite confusing. Please try it out yourself.

The change is good! Please feel free to share your suggestions.

candysmurf: @IzabellaRaulin, using `bool` is a design mistake in my opinion while env var is also supported. The implementation does not work for env var. It's a bug.

candysmurf:
#1672

PR #1678, opened 2017-07-06
Updated exemplary streaming collector plugin

### LINKED ISSUE #1675: Error creating a streaming task: Plugin is incompatible with the tasks schedule type.
**Steps to reproduce:**
1. Start snap using `snapteld -t 0 -l 1`
2. Load streaming rand plugin (snap-plugin-stream-collector-rand1) from `build/darwn/x86_64/plugins` 
3. Create any streaming task, using `snaptel task create -t rand.yaml` that results in:
`Using task manifest to create task
Error creating task:Plugin is incompatible with the tasks schedule type.`

**Expected behavior:** The streaming task loads without any errors, similar to the result of running the rand plugin from snap-plugin-lib-go. Reason for this error in the snap framework is because the snap-plugin-stream-collector-rand1 binary isn't updated and it returns a "collector" plugin type when it's actually supposed to return a "streaming-collector" plugin type which makes it inconsistent with the schedule type - "streaming".

IzabellaRaulin:
Fixes #1675 

Summary of changes:
- updated glide.lock to the latest snap-plugin-lib-go
- aligned exemplary streaming collector plugin with this one available on snap-plugin-lib-go repo
- removed binary files (remotey and streamer)
- added `Notice` to snap/plugin/collector/snap-plugin-streaming-collector-rand1/README.md to direct users to original repo of the exemplary plugin

Testing done:
- streaming-collector plugin loads successfully

@intelsdi-x/snap-maintainers


kjlyon:
Why is this example called streaming-collector-rand1 ? why not streaming-collector-rand?

IzabellaRaulin:
These exemplary plugins are there for a testing purpose. I do not change the pattern that version is a part of plugin's name what is helpful in tests validating switching to the latest version on the fly:
image

PR #1681, opened 2017-07-10
Communicate taskID to streaming plugins

rashmigottipati:
Fixes #

Summary of changes:
- Add taskID to context and metadata which is passed through the grpc stream
- Update glide.yaml to latest grpc 
- Also set the MaxMetricsBuffer and MaxCollectDuration and add them to context for the plugins to receive the values already set

Testing done:
- Manual

@intelsdi-x/snap-maintainers


rashmigottipati:
https://github.com/intelsdi-x/snap-plugin-lib-go/pull/90 needs to land first, for the travis build to pass in this PR. The conflict is in the glide.lock which is using plugin-lib-go which refers to an old version of grpc. The open PR on snap-plugin-lib-go points to the latest grpc version.

pkg/rpcutil/rpc.go
```diff
@@ -31,13 +32,13 @@ import (
 const grpcDialDefaultTimeout = 2 * time.Second
 
 // GetClientConnection returns a grcp.ClientConn that is unsecured
-func GetClientConnection(addr string, port int) (*grpc.ClientConn, error) {
-	return GetClientConnectionWithCreds(addr, port, nil)
+func GetClientConnection(ctx context.Context, addr string, port int) (*grpc.ClientConn, error) {
```
kjlyon: Don't forget to update the test file to reflect this change. Legacy tests are currently failing because of this change. The failure is in control_grpc_server_test.go at line 135. It is failing because not enough arguments were provided.

kjlyon:
lgtm

control/plugin/client/grpc.go
```diff
@@ -132,7 +135,13 @@ func NewCollectorGrpcClient(address string, timeout time.Duration, security GRPC
 
 // NewStreamCollectorGrpcClient returns a stream collector gRPC client
 func NewStreamCollectorGrpcClient(address string, timeout time.Duration, security GRPCSecurity) (PluginStreamCollectorClient, error) {
-	p, err := newPluginGrpcClient(address, timeout, security, plugin.StreamCollectorPluginType)
+	header := metadata.New(map[string]string{
+		"MaxCollectDuration": "5s",
```
marcintao: Why are the streaming control parameters given here? I couldn't easily find any place where these items were received (I've checked in snap and snap-plugin-lib-go).

rashmigottipati: Thanks for the comment. The idea was to pass just the task-id. I've added these other parameters for testing purposes but we currently don't receive them anywhere. So I'll go ahead and remove these.

rashmigottipati: Updated the PR.

glide.lock
```diff
@@ -25,14 +25,15 @@ imports:
...
   version: fa3f63826f7c23912c15263591e65d54d080b458
   subpackages:
   - codec
 - name: github.com/hashicorp/memberlist
   version: a93fbd426dd831f5a66db3adc6a5ffa6f44cc60a
 - name: github.com/intelsdi-x/gomit
-  version: db68f6fda248706a71980abc58e969fcd63f5ea6
+  version: 286c3ad6599724faed681cd18a9ff595b00d9f3f
 - name: github.com/intelsdi-x/snap-plugin-lib-go
   version: a8b9252d1c8305deb3c24495f1e9ce26607fcbd7
```
IzabellaRaulin: Could you check the version of snap-plugin-lib-go? I suppose it should point to commit 3e667 at least (taking the latest is preferable).

control/plugin/client/grpc.go
```diff
@@ -123,7 +125,8 @@ func SecurityTLSOff() GRPCSecurity {
 
 // NewCollectorGrpcClient returns a collector gRPC Client.
 func NewCollectorGrpcClient(address string, timeout time.Duration, security GRPCSecurity) (PluginCollectorClient, error) {
-	p, err := newPluginGrpcClient(address, timeout, security, plugin.CollectorPluginType)
+	ctx := context.Background()
+	p, err := newPluginGrpcClient(ctx, address, timeout, security, plugin.CollectorPluginType)
```
IzabellaRaulin: @rashmigottipati, does it have sense to move creating context (as in line https://github.com/intelsdi-x/snap/pull/1681/files#diff-706e573cd7a189d87a3edea5ee3ed0f4R128) from `NewCollectorGrpcClient`  to `newPluginGrpcClient()` or even lower to `newGrpcClient()`? Is this possible to avoid passing ctx cross all layers?

glide.lock
rashmigottipati: I fixed it to point to the commit: 3e667

Issue #1659, opened 2017-06-12
Task is still running even used publisher plugin has been unloaded


IzabellaRaulin:
At present, there is a possibility to unload the plugin even if it is used by running task.  I used [exemplary task manifest](https://github.com/intelsdi-x/snap-plugin-collector-cpu/blob/master/examples/tasks/cpu-file.json) (collecting cpu metrics and publishing them to a file) to verify is that behavior is what is really expected.

*Scenario A: unloading collector plugin:*
1) Load cpu and file plugins. Create and start the task 
2) Unload collector plugin: `snaptel plugin unload collector cpu 6`
Observation:
Receiving errors i.a. " Metric not found: /intel/procfs/cpu/*/guest_jiff:". After 10 Fails, task status changes from `Running` to `Disabled`

*Scenario B: with unloading publisher plugin:*
1) Load cpu and file plugins. Create and start the task 
2) Unload publisher plugin: `snaptel plugin unload publisher file 2`
Observation:
No errors, no misses. The task has status _Running_. Metrics are still publishing to file.
3) Stop the task and try to re-start it
    ```
    snaptel task start 94969d7a-cd0c-4f6d-8281-109e3977b0b6
    Error starting task:
    error 0: Plugin not found: type(publisher) name(file) version(-1)
    ```

Basically, **the behavior should be similar in both cases** - this is what I expect. 
The another option which might be taken into consideration is **not allowing for unloading the plugin which is used in a running task**.

PR #1683, opened 2017-07-12
Fix publisher / processor unloading

### LINKED ISSUE #1676: SubscriptionGroup does not provide the version of subscribed publisher and processor plugins
### Introduction
The goal of subscriptionGroup is to provide
- **requestedPlugins** -  contains only processors and publishers with version as declared in task manifest 
- **requestedMetrics** - requested metrics in exactly the same form as declared in task manifest 
- **plugins**  - contains subscribed plugins (all types: collectors, processors, publishers) with exact   versions of the plugin which are currently used by running task
See definition [here](https://github.com/intelsdi-x/snap/blob/master/control/subscription_group.go#L66)

**What happened vs what is expected**:
There is no defined version for publisher/processor in subscribedPlugins
 - requestedPlugins

type | name | version | notes
-----|--------|------|-----
processor | passthru | -1 | as expected
publisher | mock-file | -1 | as expected

 - subscribedPlugins

 type | name | version | notes
-----|--------|------|-----
collector | mock | 2 | as expected
processor | passthru | -1 | !!!!Info about already use version should be here (e.g. 1)
publisher | mock-file | -1 | !!!!Info about already use version should be here (e.g. 3)

### Impact
How to unsubscribe plugin when the version is not provided explicitly - it might be related with https://github.com/intelsdi-x/snap/issues/1659 (as one element in the whole chain )

mkleina:
Fixes #1676 #1659 

Summary of changes:
- findLatestPool() now returns latest pool with at least 1 plugin assigned, empty pools are skipped
- Plugin version for processors and publishers is resolved during subscription processing, if not provided in task manifest

Testing done:
- Manual testing
- Unit test

Task fails now after unloading all versions of used publisher / processor, unloading single version causes switching to another (latest one)

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/available_plugin.go
+++ b/control/available_plugin.go
@@ -625,7 +625,7 @@ func (ap *availablePlugins) findLatestPool(pType, name string) (strategy.Pool, s
 	var latest strategy.Pool
 	for key, pool := range ap.table {
 		tnv := strings.Split(key, core.Separator)
-		if tnv[0] == pType && tnv[1] == name {
+		if tnv[0] == pType && tnv[1] == name && pool.Count() > 0 {
 			latest = pool
 			break
 		}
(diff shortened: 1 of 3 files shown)


mkleina:
@marcintao Thanks! I squashed changes to single commit.

PR #1692, opened 2017-07-18
Fix so plugin will get unloaded and tasks disabled after plugin killed

### LINKED ISSUE #1680: Streaming plugin remains running after plugin killed
**Behavior:**
After loading a streaming plugin into Snap, if you kill the plugin with ctrl+C, in Snap the plugin still shows as loaded. Additionally tasks that depend on that plugin also remain running without error. Snap logs show that it is trying to reconnect with the plugin, but it isn't able to reconnect.

**Expected Behavior:**
If a loaded streaming plugin is killed then Snap should unload the plugin and the task should error and eventually stop. 

It appears that we will need to set a limit to number of times the framework will try to reconnect to the plugin before it stops trying and eventually unloads the plugin.

kjlyon:
Fixes #1680

Summary of changes:
- Plugin gets unloaded and relevant tasks get disabled after a stand-alone plugin is killed

To Test:
- Start Snap
- Start plugin in stand-alone mode
- Load plugin to Snap
- Load task
- Manually kill plugin (ctrl+c)

Expected Behavior:  
- Plugin is unloaded, including metrics removed from list
- Relevant tasks are disabled 

Testing done:
- Manual
- Unit

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/available_plugin.go
+++ b/control/available_plugin.go
@@ -302,13 +302,6 @@ func (a *availablePlugin) Kill(r string) error {
 // CheckHealth checks the health of a plugin and updates
 // a.failedHealthChecks
 func (a *availablePlugin) CheckHealth() {
-	if a.IsRemote() {
-		runnerLog.WithFields(log.Fields{
-			"_module": "control-aplugin",
-			"_block":  "check-health",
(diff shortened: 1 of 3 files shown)


PatrykMatyjasek:
LGTM

PR #1694, opened 2017-07-19
fix for streaming plugins

kjlyon:
Summary of changes:
- In getMetricsAndCollectors will get from both collectors and streaming-collectors
- Fix in StreamMetrics()

Testing done:
- unit

@intelsdi-x/snap-maintainers


rashmigottipati:
LGTM

PR #1703, opened 2017-08-02
SDI-2730: Add SSL/TLS capability to snap-cli

candysmurf:
Summary of changes:
- Added TLS plugin support for REST API v2

Testing done:
- small, medium

@intelsdi-x/snap-maintainers

### DIFF
--- a/mgmt/rest/v2/plugin.go
+++ b/mgmt/rest/v2/plugin.go
@@ -41,6 +41,15 @@ import (
 	"github.com/julienschmidt/httprouter"
 )
 
+const (
+	// TLSCertPrefix defines a prefix for file fragment carrying path to TLS certificate
+	TLSCertPrefix = "crt."
+	// TLSKeyPrefix defines a prefix for file fragment carrying path to TLS private key
+	TLSKeyPrefix = "key."
+	// TLSCACertsPrefix defines a prefix for file fragment carrying paths to TLS CA certificates
+	TLSCACertsPrefix = "cacerts."
+)
+
 // PluginResponse represents the response from plugin operations.
 //
 // swagger:response PluginResponse
@@ -122,6 +131,24 @@ type PluginPostParams struct {
 	//
 	// swagger:file
 	PluginData *bytes.Buffer `json:"plugin_data"`
+	// Plugin GRPC TLS server key
+	//
+	// in: formData
+	//
+	// swagger:file
+	PluginKey *bytes.Buffer `json:"plugin_key"`
+	// Plugin GRPC TLS server certification
+	//
+	// in: formData
+	//
+	// swagger:file
+	PluginCert *bytes.Buffer `json:"plugin_cert"`
+	// CA root certification
+	//
+	// in: formData
+	//
+	// swagger:file
+	CACerts *bytes.Buffer `json:"ca_certs"`
 }
 
 // Name plugin name string
@@ -148,11 +175,12 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
 		return
 	}
 
+	var certPath, keyPath, caCertPaths string
 	if strings.HasPrefix(mediaType, "multipart/") {
 		var signature []byte
 		var checkSum [sha256.Size]byte
 		mr := multipart.NewReader(r.Body, params["boundary"])
-		var i int
+
 		for {
 			var b []byte
 			p, err := mr.NextPart()
@@ -183,40 +211,44 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
 				}
 			}
 
-			// A little sanity checking for files being passed into the API server.
-			// First file passed in should be the plugin. If the first file is a signature
-			// file, an error is returned. The signature file should be the second
-			// file passed to the API server. If the second file does not have the ".asc"
-			// extension, an error is returned.
-			// If we loop around more than twice before receiving io.EOF, then
-			// an error is returned.
-
-			switch {
-			case i == 0:
-				if filepath.Ext(p.FileName()) == ".asc" {
-					e := errors.New("Error: first file passed to load plugin api can not be signature file")
-					Write(400, FromError(e), w)
+			switch p.FormName() {
+			case "ca_certs":
+				fn, err := createTempFile(b, TLSCACertsPrefix)
+				if err != nil {
+					Write(500, FromError(err), w)
 					return
 				}
+				caCertPaths = fn
+			case "plugin_key":
+				fn, err := createTempFile(b, TLSKeyPrefix)
+				if err != nil {
+					Write(500, FromError(err), w)
+					return
+				}
+				keyPath = fn
+			case "plugin_cert":
+				fn, err := createTempFile(b, TLSCertPrefix)
+				if err != nil {
+					Write(500, FromError(err), w)
+					return
+				}
+				certPath = fn
+			// plugin_data is from REST API and snap-plugins is from rest_v2_test.go.
+			case "plugin_data", "snap-plugins":
 				if rp, err = core.NewRequestedPlugin(p.FileName(), s.metricManager.GetTempDir(), b); err != nil {
 					Write(500, FromError(err), w)
 					return
 				}
 				checkSum = sha256.Sum256(b)
-			case i == 1:
+			default:
 				if filepath.Ext(p.FileName()) == ".asc" {
 					signature = b
 				} else {
-					e := errors.New("Error: second file passed was not a signature file")
+					e := errors.New("Error: An unknown file found " + p.FileName())
 					Write(400, FromError(e), w)
 					return
 				}
-			case i == 2:
-				e := errors.New("Error: More than two files passed to the load plugin api")
-				Write(400, FromError(e), w)
-				return
 			}
-			i++
 		}
 
 		// Sanity check, verify the checkSum on the file sent is the same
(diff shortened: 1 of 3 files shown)


mgmt/rest/v2/plugin.go
```diff
@@ -244,13 +280,40 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
 			default:
 				ec = 500
 			}
+			cleanUpTempFiles(rp)
```
IzabellaRaulin: consider using `defer cleanUpTempFiles` instead of lines 283 and 287

mgmt/rest/v2/plugin.go
```diff
@@ -244,13 +280,40 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
...
+	if err != nil {
+		return "", err
+	}
+	return file.Name(), nil
+}
+
+func cleanUpTempFiles(rp *core.RequestedPlugin) {
+	defer os.Remove(rp.CACertPaths())
+	defer os.Remove(rp.CertPath())
+	defer os.Remove(rp.KeyPath())
```
IzabellaRaulin: In my opinion using "defer" here does not have sense - @candysmurf, maybe it is something which is no longer valid and you missed removing that. Could you look on that again?

candysmurf: @IzabellaRaulin, once defer is added for clearnupTempFiles, then defer here needs to be removed.

mgmt/rest/v2/plugin.go
candysmurf: @iza, leaving line 283 as it is? as the op is included in an error.

mgmt/rest/v2/plugin.go
```diff
@@ -41,6 +41,15 @@ import (
 	"github.com/julienschmidt/httprouter"
 )
 
+const (
+	// TLSCertPrefix defines a prefix for file fragment carrying path to TLS certificate
+	TLSCertPrefix = "crt."
+	// TLSKeyPrefix defines a prefix for file fragment carrying path to TLS private key
+	TLSKeyPrefix = "key."
+	// TLSCACertsPrefix defines a prefix for file fragment carrying paths to TLS CA certificates
+	TLSCACertsPrefix = "cacerts."
```
IzabellaRaulin: @candysmurf, could we keep one pattern and move declaration of those constants to https://github.com/intelsdi-x/snap/blob/master/mgmt/rest/v2/api.go (as it's done for V1, see https://github.com/intelsdi-x/snap/blob/master/mgmt/rest/v1/api.go)

mgmt/rest/v2/plugin.go
```diff
@@ -148,11 +175,12 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
 		return
 	}
 
+	var certPath, keyPath, caCertPaths string
 	if strings.HasPrefix(mediaType, "multipart/") {
```
IzabellaRaulin: Declaration of vars should be moved inside this `if` statement. There is no use of those vars outside.

mgmt/rest/v2/plugin.go
```diff
@@ -183,40 +211,44 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
...
-			// an error is returned.
-
-			switch {
-			case i == 0:
-				if filepath.Ext(p.FileName()) == ".asc" {
-					e := errors.New("Error: first file passed to load plugin api can not be signature file")
-					Write(400, FromError(e), w)
+			switch p.FormName() {
+			case "ca_certs":
+				fn, err := createTempFile(b, TLSCACertsPrefix)
```
IzabellaRaulin: @candysmurf, I looked more thoroughly on this PR, and I do not think so that creating "temp" file is a good approach. What needs to be done here is only checking if provided path exists - so why you decided to add 
 [createTempFile](https://github.com/intelsdi-x/snap/pull/1703/files#diff-d96666069f58edcf865ae5d038e1ce95R299) function instead of straightforward do sth like that:  `if _, err := os.Stat(certPath); os.IsNotExist(err)` ?

My preference is avoid creating temp file if they are not needed. Please consider following what is done in V1 - https://github.com/intelsdi-x/snap/blob/master/mgmt/rest/v1/plugin.go#L144

mgmt/rest/v2/plugin.go
```diff
@@ -227,6 +259,10 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
 			return
 		}
 		rp.SetSignature(signature)
+		rp.SetCACertPaths(caCertPaths)
+		rp.SetCertPath(certPath)
+		rp.SetKeyPath(keyPath)
+		rp.SetTLSEnabled(isTLSEnabled(certPath, keyPath))
```
IzabellaRaulin: I understand the goal, however, it looks like inception.  But what is more, there is no handling of error when one of pair is not provided. 
I suggest to change it as it is done in API v1 - see https://github.com/intelsdi-x/snap/blob/master/mgmt/rest/v1/plugin.go#L182
```
if certPath != "" && keyPath != "" {
	rp.SetTLSEnabled(true)
} else if certPath != "" || keyPath != "" {
	e := errors.New("Error: TLS setup incomplete - missing one of pair: certificate, key files")
	rbody.Write(500, rbody.FromError(e), w)
	return
}
```

mgmt/rest/v2/plugin.go
```diff
@@ -244,13 +280,40 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
...
+	file, err := ioutil.TempFile(os.TempDir(), fn)
+	if err != nil {
+		return "", err
+	}
+	_, err = file.Write(b)
+	if err != nil {
+		return "", err
+	}
+	return file.Name(), nil
+}
```
IzabellaRaulin: Please, check if creating temp files is necessary. As far I know, checking if the path points to existing file is enough
Example (taken from API v1):
```
if _, err := os.Stat(keyPath); os.IsNotExist(err) {
	e := errors.New("Error: given key file is not available")
	rbody.Write(500, rbody.FromError(e), w)
	return
}
```

mgmt/rest/v2/plugin.go
```diff
@@ -244,13 +280,40 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
...
+
+func cleanUpTempFiles(rp *core.RequestedPlugin) {
```
IzabellaRaulin: It might be removed if you decide to remove `createTempFile`

mgmt/rest/v2/plugin.go
candysmurf: @iza, thanks for your review. The situation here is different than v1. V1's client always resides inside Snap on the same machine. V2 passes the content of certs and key over to Snap as `snap-cli` may install on a different server. By leveraging what is available in Snap (requires paths of certs and the key), that's how creating temp files are necessary. Otherwise, I need to change Snap's code to use the content of certs and the key directly. 

Agreed, I don't prefer to use temp files. Unless we assume that those certs and the key files are always accessible even when they're on a different server. 

Please let me know how you like to proceed.

mgmt/rest/v2/plugin.go
candysmurf: @IzabellaRaulin: added the error handling. Please check.

mgmt/rest/v2/plugin.go
candysmurf: @Iza, I would be happy to change the implementation if we can assume those certs and the key files are always accessible. Note that I implemented `snaptel` in V2 using "https" for this. We can remove https as well. Please let me know you guys take on this.

mgmt/rest/v2/plugin.go
```diff
@@ -183,40 +211,45 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
...
+			case "ca_certs":
+				fn, err := createTempFile(b, TLSCACertsPrefix)
+				if err != nil {
+					Write(500, FromError(err), w)
 					return
 				}
-				if rp, err = core.NewRequestedPlugin(p.FileName(), s.metricManager.GetTempDir(), b); err != nil {
+				caCertPaths = fn
+			case "plugin_key":
+				fn, err := createTempFile(b, TLSKeyPrefix)
```
marcintao: Storing contents of private key in arbitrary directory won't be safe.

mgmt/rest/v2/plugin.go
```diff
@@ -244,13 +291,47 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
 			default:
 				ec = 500
 			}
+			defer cleanUpTempFiles(rp)
 			Write(ec, rb, w)
 			return
 		}
+		defer cleanUpTempFiles(rp)
```
marcintao: Secure plugins won't work with this line in place.

mgmt/rest/v2/plugin.go
marcintao: This method is unnecessary. None of the files listed here should lend in temporary directory.

candysmurf: @marcintao, are you suggesting that we only provide the paths. If they installed snap-cli on a different machine/remote, it's their issue to resolve it? Please confirm.

candysmurf:
@marcintao, I found that in Snap[cmd package](https://github.com/intelsdi-x/snap/blob/master/cmd/snaptel/plugin.go#L211), the key and certs are copied to the temp directory. Would you please explain why you have to do that? 

What I found is that cert files get deleted if I don't do that. Please elaborate.

mgmt/rest/v2/plugin.go
```diff
@@ -226,7 +244,21 @@ func (s *apiV2) loadPlugin(w http.ResponseWriter, r *http.Request, _ httprouter.
 			Write(400, FromError(e), w)
 			return
 		}
+
+		if hasTLS(certPath, keyPath) {
```
IzabellaRaulin: I have only one small suggestion to add comments to this fragment of code:
- that `hasTLS` checks if one of TLS params (cert or key) has been provided, if not skip the part related to TLS
- that `isTLSEnabled` checks if both of required params have been provided to setup TLS connection; if not return an appropriate error

I made a PR to your repo to not bother you with this small change (https://github.com/candysmurf/snap/pull/4) - use it or add some comments by your own.

PR #1704, opened 2017-08-02
Closes #1137 - more detailed error message

### LINKED ISSUE #1137: Snapctl task create needs more detailed error message from malformed input files
When user creates a task in snapctl: "snapctl task create -t task.json" and the input format is malformed the error message that the user is given for some errors is: 

Error parsing JSON file input - json: cannot unmarshal string into Go value of type map[string]interface {}

Even a line number would be useful in this situation. Now the user needs to figure out the correct line where the error lies. In namespace error the message is:

Error creating task:Invalid namespace: intel/proc

When the user is missing the first "/". This line gives more information for correcting the issue compared to the "Error parsing JSON file input - json: cannot unmarshal string into Go value of type map[string]interface {}".

I'm asking if the line number from malformed input file would be added to the error message.

Alrim42:
Closes #1137 - Snapctl task create needs more detailed error message from malformed input files
-

Summary of changes:
- 
- added more detailed error message containing line where the error occurred

How to verify it:
-
1. Create malformed JSON file - task manifest (or workflow manifest)
 Example (missing quotation mark):
```
{
  "version": 1,
  "schedule": {
    "type": "simple",
    "interval": "1s"
  },
  "max-failures": 1,
  "workflow": {
    "collect": {
      "metrics": {
        /intel/psutil/load/load15": {}
      },
      "publish": [
        {
          "plugin_name": "influxdb",
          "config": {
            "host": "${INFLUXDB_HOST}",
            "port": 8086,
            "database": "test",
            "user": "admin",
            "password": "admin",
            "scheme": "http",
            "skip-verify": false,
            "isMultiFields": true
          }
        }
      ]
    }
  }
}
```

2. Create task from task manifest (or workflow manifest)
```
$ snaptel task create -t test.json
Fail: Cannot parse JSON due to a syntax error at line 11: invalid character '/' looking for beginning of object key string
Error parsing JSON file input - invalid character '/' looking for beginning of object key string
```

Error now contains line in the JSON file where the error occurred.

No changes are needed when handling YAML files, because it is already done:
```
$ snaptel task create -t test.yml
Error parsing YAML file input - error converting YAML to JSON: yaml: line 10: did not find expected key
```

Testing done:
- 
- `make` works
- small test passed
- medium test passed
- large test passed
- legacy test passed
- manual test satisfying

@intelsdi-x/snap-maintainers

### DIFF
--- a/cmd/snaptel/task.go
+++ b/cmd/snaptel/task.go
@@ -364,6 +364,7 @@ func createTaskUsingTaskManifest(ctx *cli.Context) error {
 	case ".json":
 		e = json.Unmarshal(file, &t)
 		if e != nil {
+			showLineWithError(file, e)
 			return fmt.Errorf("Error parsing JSON file input - %v\n", e)
 		}
 	default:
@@ -419,15 +420,14 @@ func createTaskUsingWFManifest(ctx *cli.Context) error {
 	var wf *wmap.WorkflowMap
 	switch ext {
 	case ".yaml", ".yml":
-		// e = yaml.Unmarshal(file, &t)
 		wf, e = wmap.FromYaml(file)
 		if e != nil {
 			return fmt.Errorf("Error parsing YAML file input - %v\n", e)
 		}
 	case ".json":
 		wf, e = wmap.FromJson(file)
-		// e = json.Unmarshal(file, &t)
 		if e != nil {
+			showLineWithError(file, e)
 			return fmt.Errorf("Error parsing JSON file input - %v\n", e)
 		}
 	}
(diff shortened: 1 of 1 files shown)


cmd/snaptel/task.go
```diff
@@ -364,6 +364,22 @@ func createTaskUsingTaskManifest(ctx *cli.Context) error {
 	case ".json":
 		e = json.Unmarshal(file, &t)
 		if e != nil {
+			if jsonError, ok := e.(*json.SyntaxError); ok {
```
mkleina: Instead of duplicating whole block of code with long error messages, consider moving it to separate function, for example handleJSONSyntaxError(file, error) and use just like:
```go
e = json.Unmarshal(file, &t)
if e != nil {
   handleJSONSyntaxError(file, e)
}
```

Issue #1697, opened 2017-07-21
Reconnect to a stand-alone plugin if blip in connection


kjlyon:
**Current Behavior:** Currently if there is any disruption in the connection between Snap and a stand alone plugin then the task is disabled and the plugin is unloaded.

**Expected Behavior:** If there is a blip in the connection between Snap framework and a stand alone plugin, Snap should try for 3 seconds to reconnect to the plugin. If the connection cannot be reestablished then the task should be disabled and plugin unloaded. 

This presents some issues if the plugin was restarted on a new dynamically allocated port. If that happens then Snap won't know what port to connect to.  Though if the plugin is restarted on the same port or if the plugin never stopped then the grpc connection should be able to reconnect.

PR #1719, opened 2017-08-22
Documentation for stand-alone mode & plugin diagnostics

### LINKED ISSUE #1687: Missing documentation for stand-alone feature
Missing documentation for stand-alone feature (at least I don't see it). Users should know how use stand-alone plugins.

katarzyna-z:
Fixes #1687 

Summary of changes:
- Documentation for standalone plugin
- Documentation for plugin diagnostics

Please check if I covered all information related to standalone plugin and plugin diagnostics from user perspective.
I'm open for your suggestion related to this docs :turtle: 

@intelsdi-x/snap-maintainers

### DIFF
--- a/docs/PLUGIN_DIAGNOSTICS.md
+++ b/docs/PLUGIN_DIAGNOSTICS.md
@@ -1,0 +1,67 @@
+# Plugin Diagnostics
+
+Plugin diagnostics provides a simple way to verify that a plugin is capable of running without requiring the plugin to be loaded and a task started.
+This feature works for collector plugins written using one of our snap-plugin-libs ([snap-plugin-lib-go](https://github.com/intelsdi-x/snap-plugin-lib-go), [snap-plugin-lib-py](https://github.com/intelsdi-x/snap-plugin-lib-py), [snap-plugin-lib-cpp](https://github.com/intelsdi-x/snap-plugin-lib-cpp)).
+
+Plugin diagnostics delivers following information:
+- runtime details (plugin name, version, etc.)
+- configuration details
+- list of metrics exposed by the plugin
+- metrics with their values that can be collected right now
+- and times required to retrieve this information
+
+## Running plugin diagnostics
+It is possible to run plugin diagnostics without any arguments (default values are used), e.g.:
+```
+$ ./snap-plugin-collector-psutil
+```
+or with plugin configuration in the form of a JSON:
+```
+$ ./snap-plugin-collector-psutil --config {\"mouting_points\":\"/dev/sda\"}
+```
+
+Example output:
+
+```
+$ ./snap-plugin-collector-psutil
+
+Runtime Details:
+    PluginName: psutil, Version: 14
+    RPC Type: gRPC, RPC Version: 1
+    Operating system: linux
+    Architecture: amd64
+    Go version: go1.7
+printRuntimeDetails took 11.878µs
+
+Config Policy:
+NAMESPACE 		 KEY 		 TYPE 		 REQUIRED 	 DEFAULT 	 MINIMUM 	 MAXIMUM
+intel.psutil.disk 	 mount_points 	 string 	 false 		
+printConfigPolicy took 67.654µs
+
+Metric catalog will be updated to include: 
+    Namespace: /intel/psutil/load/load1 
+    Namespace: /intel/psutil/load/load5 
+    Namespace: /intel/psutil/load/load15 
+    Namespace: /intel/psutil/cpu/*/softirq 
+    Namespace: /intel/psutil/cpu/cpu-total/softirq 
+    [...]
+printMetricTypes took 504.483µs 
+
+Metrics that can be collected right now are: 
+    Namespace: /intel/psutil/load/load1        Type: float64     Value: 1.48 
+    Namespace: /intel/psutil/load/load5        Type: float64     Value: 1.68 
+    Namespace: /intel/psutil/load/load15       Type: float64     Value: 1.65 
+    Namespace: /intel/psutil/cpu/cpu0/softirq  Type: float64     Value: 20.36 
+    Namespace: /intel/psutil/cpu/cpu1/softirq  Type: float64     Value: 13.62 
+    Namespace: /intel/psutil/cpu/cpu2/softirq  Type: float64     Value: 9.96 
+    Namespace: /intel/psutil/cpu/cpu3/softirq  Type: float64     Value: 3.6 
+    Namespace: /intel/psutil/cpu/cpu4/softirq  Type: float64     Value: 1.42 
+    Namespace: /intel/psutil/cpu/cpu5/softirq  Type: float64     Value: 0.69 
+    Namespace: /intel/psutil/cpu/cpu6/softirq  Type: float64     Value: 0.54 
+    Namespace: /intel/psutil/cpu/cpu7/softirq  Type: float64     Value: 0.31 
+    Namespace: /intel/psutil/cpu/cpu-total/softirq  Type: float64     Value: 50.52
+    [...]
+printCollectMetrics took 7.470091ms 
+
+showDiagnostics took 8.076025ms
+```
--- a/docs/STAND-ALONE_MODE.md
+++ b/docs/STAND-ALONE_MODE.md
@@ -1,0 +1,39 @@
+# Stand-alone mode
+
+Stand-alone mode enables plugin launching on different machine than Snap daemon (`snapteld`).
+This feature works for plugins written using one of our snap-plugin-libs ([snap-plugin-lib-go](https://github.com/intelsdi-x/snap-plugin-lib-go),
+[snap-plugin-lib-py](https://github.com/intelsdi-x/snap-plugin-lib-py), [snap-plugin-lib-cpp](https://github.com/intelsdi-x/snap-plugin-lib-cpp)).
+
+## Running a plugin in stand-alone mode
+To run a plugin in stand-alone mode, you must start it with the `--stand-alone` flag, e.g.:
+```
+$ ./snap-plugin-collector-psutil --stand-alone
+```
+
+A plugin running in stand-alone mode creates a HTTP server for communication with the Snap framework.
+By default the plugin listens on port `8182`.
+
+To specify a different listening port, use the `--stand-alone-port` flag, e.g.:
+```
+$ ./snap-plugin-collector-psutil --stand-alone --stand-alone-port 8183
+```
+##  Loading a plugin
+When loading a plugin in stand-alone mode into `snapteld` you must provide a URL to indicate which
+machine the plugin is running (IP address/hostname with port number), e.g.:
+
+```
+$ snaptel plugin load http://127.0.0.1:8182
+```
+
+or
+
+```
+$ snaptel plugin load http://localhost:8182
+```
+
+The rest of operations remains exactly the same as is it for plugins running in regular mode.
+
+## Notice
+
+If there is any disruption in the connection between Snap and a stand-alone plugin then the task is disabled and the plugin is unloaded,
+see [github issue](https://github.com/intelsdi-x/snap/issues/1697).


docs/STANDALONE_PLUGIN.md
```diff
@@ -0,0 +1,30 @@
+# Standalone Plugin
```
kdembler: I think we could go with `Standalone Mode` instead of `Standalone Plugin`, as every plugin can be ran in standalone, and `Standalone Plugin` kinda indicates that it's some special kind of plugin.

docs/STANDALONE_PLUGIN.md
```diff
@@ -0,0 +1,30 @@
...
+
+Standalone plugin is a plugin which can be launched on different machine than Snap daemon (`snapteld`).
+
+To run plugin in standalone mode it is needed to launched plugin with `--stand-alone` flag, e.g.:
```
kdembler: `to launch` instead `to launched`

kdembler:
LGTM

docs/PLUGIN_DIAGNOSTICS.md
```diff
@@ -0,0 +1,114 @@
+# Plugin Diagnostics
+
+Plugin diagnostics provides a simple way to verify that a plugin is capable of running without requiring the plugin to be loaded and a task started.
```
IzabellaRaulin: @katarzyna-z, could we simplify this sentence to be more clear - I will think about some proposal of that. Also, plugin-diagnostics is only collector plugin feature - that should be mentioned somewhere.

docs/PLUGIN_DIAGNOSTICS.md
```diff
@@ -0,0 +1,114 @@
...
+
+It is possible to run plugin diagnostics without any arguments, e.g.:
```
IzabellaRaulin: Add here notice that if the plugin has some configuration needed, the defaults will be taken.

docs/PLUGIN_DIAGNOSTICS.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+or with plugin configuration in the form of a JSON:
+
+```
+$ ./snap-plugin-collector-psutil --config {\"mouting_points\":\"/dev/sda\"}
+```
+
+Plugin diagnostics delivers following information:
+- Runtime details, e.g.:
```
IzabellaRaulin: @katarzyna-z, what do you think about listing all information one by one, and below provide just exemplary output from plugin diagnostics?

Plugin diagnostics delivers following information:
- Runtime details (plugin name, version, etc.)
- Configuration details 
- List of metrics expose by the plugin
- Metrics with their values that can be collected right now

docs/PLUGIN_DIAGNOSTICS.md
```diff
@@ -0,0 +1,114 @@
...
+    [...]
+    printMetricTypes took 504.483µs 
+    [...]
+    printCollectMetrics took 7.470091ms 
+    [...]
+    showDiagnostics took 8.076025ms
+    ```
+## Note
+
+For now plugin diagnostics works only for collector plugins.
```
IzabellaRaulin: Ok, now I find it.

docs/STANDALONE_MODE.md
```diff
@@ -0,0 +1,30 @@
...
+$  snaptel plugin load http://127.0.0.1:8182
+```
+
+or
+
+```
+$  snaptel plugin load http://localhost:8182
+```
+
+The rest of operations remains exactly the same as is it for regular plugins.
```
IzabellaRaulin: Using `regular plugins` might be misleading. Maybe the following is better: `The rest of operations remains exactly the same as is it for plugins running in regular mode.`

docs/STANDALONE_MODE.md
```diff
@@ -0,0 +1,30 @@
+# Standalone mode
```
kjlyon: The flag currently is `--stand-alone` with the hyphen. In this document `standalone` is used. These should match. For usability I think the flag should change to one word, but for ease, it makes more sense to just change the documentation to match the flag.

docs/STANDALONE_MODE.md
```diff
@@ -0,0 +1,30 @@
...
+
+Standalone mode enables plugin launching on different machine than Snap daemon (`snapteld`).
```
kjlyon: Rather than `launching`, I think `start` would be a better word. Like this: 
```Stand-alone mode enables plugins to start and run on a different machine than Snap daemon (`snapteld`).```

File: docs/STANDALONE_MODE.md [hunk 10]
```diff
@@ -0,0 +1,30 @@
...
+
+To run plugin in standalone mode it is needed to launch plugin with `--stand-alone` flag, e.g.:
```
kjlyon: Reword: ```To run a plugin in stand-alone mode, you must start it with the `--stand-alone` flag, e.g.: ```

docs/STANDALONE_MODE.md
```diff
@@ -0,0 +1,30 @@
...
+```
+$ ./snap-plugin-collector-psutil --stand-alone
+```
+Plugin in standalone mode creates HTTP server for communication with Snap framework, by default plugin listens on `8182` port.
```
kjlyon: Reword: ```A plugin running in stand-alone mode will create a HTTP server for communication with the Snap framework. By default the plugin listens on port `8182`.```

docs/STANDALONE_MODE.md
```diff
@@ -0,0 +1,30 @@
...
+
+It is possible to configure different port for communication with Snap framework using `--stand-alone-port` flag, e.g.:
```
kjlyon: Reword: ```To specify a different listening port, use the `--stand-alone-port` flag, e.g.:```

docs/STANDALONE_MODE.md
```diff
@@ -0,0 +1,30 @@
...
+
+```
+$ ./snap-plugin-collector-psutil --stand-alone --stand-alone-port 8183
+```
+
+To load plugin in standalone mode it is required to provide URL to indicate a machine
```
kjlyon: I think this section should have its own header so that it stands out more.

kjlyon: Possible reword: ```
When loading a stand-alone plugin into snapteld you must provide a URL to idicate which machine the plugin is running (IP address/hostname with port number), e.g.:```

kjlyon:
I'm not sure if this is the right location for this, but in the documentation somewhere we should mention that if the connection between snapteld and the plugin running in stand-alone mode is disrupted it will try to reconnect. If it is unable to reconnect the plugin will be unloaded, metric catalog updated, and any dependent tasks will stop.

IzabellaRaulin:
According to the last comment:
> I'm not sure if this is the right location for this, but in the documentation somewhere we should mention that if the connection between snapteld and the plugin running in stand-alone mode is disrupted it will try to reconnect. If it is unable to reconnect the plugin will be unloaded, metric catalog updated, and any dependent tasks will stop.

There might be added notice about known issue with brief description and link to the GH issue (1697)

docs/PLUGIN_DIAGNOSTICS.md
katarzyna-z: @kjlyon Great ideas! It is important that plugin must be written using one of our libs and diagnostics works also for streaming collectors.

docs/STANDALONE_MODE.md
katarzyna-z: @kjlyon Do you have any proposition of the header?

katarzyna-z: @kjlyon thanks!

katarzyna-z:
@kjlyon 

> I'm also thinking we need to add documentation about how to write a plugin that has diagnostics and stand-alone mode enabled. 

There is nothing special to implement in plugin to enable diagnostics and standalone mode. Diagnostics and standalone mode work for any plugin which uses one of our libs and in my opinion this information will be enough.

docs/STANDALONE_MODE.md
katarzyna-z: I don't think that it is good idea to change now `--stand-alone` flag to `--standalone` because some of our plugins have been update to new versions of plugin libs to provide diagnostics information. I think that in documentation I can change "stanalone mode" to "stand-alone mode".  @kjlyon what do you think? 

In the future if someone decide to change flag he/she should also take into account that it is needed to update all plugins which use old versions of snap-plugin-libs.

kjlyon: I agree.

docs/STANDALONE_MODE.md
kjlyon: Something like `Loading a plugin in snapteld` or simply, `Loading a plugin`

kjlyon:
@katarzyna-z 
```There is nothing special to implement in plugin to enable diagnostics and standalone mode. Diagnostics and standalone mode work for any plugin which uses one of our libs and in my opinion this information will be enough.
``` 
I agree, this information is enough, but that we should add a sentence in each of these documents saying that plugins using our libs will have these features.

katarzyna-z:
@kjlyon @IzabellaRaulin @kdembler I've updated this PR. Please take a look at this once again. Let me know if I forgot about something ;)

docs/PLUGIN_DIAGNOSTICS.md
```diff
@@ -0,0 +1,67 @@
...
+- and times required to retrieve this information
+
+## Running plugin diagnostics
+It is possible to run plugin diagnostics without any arguments (default values are used), e.g.:
+```
+$ ./snap-plugin-collector-psutil
+```
+or with plugin configuration in the form of a JSON:
+```
+$ ./snap-plugin-collector-psutil --config {\"mouting_points\":\"/dev/sda\"}
```
IzabellaRaulin: @katarzyna-z, could you check the way how to provide the config? Ssee my example below, but please verify that. The another thing is that the psutil collector does not have `mouting_points` config param, but ` mount_points`, so it's better to use config's name which really exists

```
$ ./snap-plugin-collector-psutil --config '{"mount_points":"/dev/sda"}'
```

docs/STAND-ALONE_MODE.md
```diff
@@ -0,0 +1,39 @@
...
+A plugin running in stand-alone mode creates a HTTP server for communication with the Snap framework.
+By default the plugin listens on port `8182`.
+
+To specify a different listening port, use the `--stand-alone-port` flag, e.g.:
+```
+$ ./snap-plugin-collector-psutil --stand-alone --stand-alone-port 8183
+```
+##  Loading a plugin
+When loading a plugin in stand-alone mode into `snapteld` you must provide a URL to indicate which
+machine the plugin is running (IP address/hostname with port number), e.g.:
```
IzabellaRaulin: Consider rephrasing 
`When loading a plugin in stand-alone mode into `snapteld` you must provide a URL to indicate which machine the plugin is running (IP address/hostname with port number), e.g.:`
to

`To load a plugin in stand-alone mode, provide a URL to indicate to the machine on which the plugin is running (IP address/hostname with port number), e.g.:`

docs/STAND-ALONE_MODE.md
```diff
@@ -0,0 +1,39 @@
...
+```
+$ snaptel plugin load http://localhost:8182
+```
+
+The rest of operations remains exactly the same as is it for plugins running in regular mode.
+
+## Notice
+
+If there is any disruption in the connection between Snap and a stand-alone plugin then the task is disabled and the plugin is unloaded,
+see [github issue](https://github.com/intelsdi-x/snap/issues/1697).
```
IzabellaRaulin: What do you think about renaming the section to "Known issues" instead of "Notice" and rephrasing its content:

`
If some disruption occurs in the connection between Snap and a stand-alone plugin, the running task will be stopped with disabled status and the plugin will be unloaded. Providing the mechanism of reconnecting stand-alone plugins upon network disruption is in our scope, addressed by issue #1697.`

IzabellaRaulin:
@katarzyna-z, I created a PR to your branch with addressed my comments - https://github.com/katarzyna-z/snap/pull/1. Besides these, LGTM

PR #1733, opened 2017-09-14
Add PID file support

### LINKED ISSUE #878: snapd daemon mode support
currently when executing snapd --config, the snapd process stays running in the foreground. snapd does not appear to support a mode where it runs as a daemon. This requires some work around when trying to turn snapd into a service, because the start-stop-daemon command on ubuntu, and daemon snapd on redhat will just hang forever.

```
+ start-stop-daemon --start --quiet --pidfile /var/run/snapd.pid --startas /usr/bin/snapd -- --config /etc/snap/snapd.conf.yaml
```

If we follow convention for *d process snapd should default as a daemon and only run in the foreground under debug mode such as -vv.

cce:
Fixes #878 by allowing for systemd, init scripts, etc, to be able to better manage snapteld and send it kill or restart signals.

Summary of changes:
- Added `--pid-file` flag and `pid_file` config file option to snapteld
- Added PIDFile to snapteld Config struct.

Testing done:
- Added PIDFile to command-line flags testing in snapteld_test.go.

@intelsdi-x/snap-maintainers


snapteld.go
```diff
@@ -323,6 +334,25 @@ func action(ctx *cli.Context) error {
...
+	if cfg.PIDFile != "" {
+		log.Info("Creating PID file", cfg.PIDFile)
+		f, err := os.OpenFile(cfg.PIDFile, os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			log.Error("Unable to create pidfile", err)
+		} else {
+			fmt.Fprintf(f, "%d\n", os.Getpid())
+			f.Close()
+			defer func() {
+				err := os.Remove(cfg.PIDFile)
```
donzh: Sorry, I have a stupid question to ask.
Why delete the PIDFile?

WangJialei-A:
@katarzyna-z  I think it is better to handle the PID file in the service script, but not in snapteld.

snapteld.go
cce: Most servers are expected to delete their PID file when the server shuts down — for example, [Apache removes its PID file when shutting down](https://httpd.apache.org/docs/trunk/stopping.html#gracefulstop).

However, since snapteld calls `os.Exit()` when it shuts down, I recently realized that this deferred function will not be executed, since `os.Exit` shuts down immediately without calling deferred functions.

donzh: @cce 
Thanks for your explanation!
Yes, I think the the actual behavior is not your meaning.
Could you please update this PR according your design when you are free?

PR #1736, opened 2017-09-21
Add detection of plugin child process log level

cce:
This attempts to guess if a plugin is reporting an Error, Warning, or Info log message so that when snapteld reads the captured output, it can log the message at the same level.

@intelsdi-x/snap-maintainers

### DIFF
--- a/control/plugin/execution.go
+++ b/control/plugin/execution.go
@@ -131,10 +131,10 @@ func (e *ExecutablePlugin) Run(timeout time.Duration) (Response, error) {
 					respReceived = true
 					close(doneChan)
 				} else {
-					execLogger.WithFields(log.Fields{
+					logCapturedLine(execLogger.WithFields(log.Fields{
 						"plugin": e.name,
 						"io":     "stdout",
-					}).Debug(stdOutScanner.Text())
+					}), stdOutScanner.Text())
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


control/plugin/execution.go
```diff
@@ -228,3 +228,19 @@ func (e *ExecutablePlugin) captureStderr() {
 		}
 	}()
 }
+
+// logCapturedLine logs a captured stderr/stdout line from a plugin, after choosing the
+// appropriate log level to use based on the line's contents.
+func logCapturedLine(entry *log.Entry, text string) {
+	lc := strings.ToLower(text)
+	switch {
+	case strings.Contains(lc, "error"):
```
WangJialei-A: @cce  I think it is not a good idea to detect log level in this way. 
For example, the message 'warn: cannot collect metric /xx/xx/xx/error_rate from xxx platform' will be an error instead of warning in your case.

cce: You're right, I was looking for a quick solution to ensure that a plugin's error logs are not logged at "DEBUG" level. What do you think is the best approach? Could we assume logrus format and look for `level=error`?

dancyding: @cce There's a similar issue already opened.
https://github.com/intelsdi-x/snap/issues/1186

dancyding: @cce I think parsing "level=error" like strings approach is acceptable. Though it will not be perfect (may be confused if this phrase occurred in log message), but the probability is rather low and the cost is not too high.

PR #1762, opened 2018-07-01
updated to abstract the hard path of bash since freebsd does not incl…

chair300:
…ude bash in the default path of /bin

Fixes #

Summary of changes:
- 
- 
- 

Testing done:
- 

@intelsdi-x/snap-maintainers


WangJialei-A:
LGTM

